PR #9, opened 2023-03-23
syslog-ng-splunk-connector image

MrAnno:
This PR introduces stable SC4S images for the following platforms:
- linux/amd64
- linux/arm64

The image is based on the Axoflow syslog-ng image (Alpine).

TODO:
- [x] Auto-draft release
- [x] Fix db_close
- [x] Fix persist collisions
- [x] Fix v3 typing compat
- [x] Fix zscaler parsing (`%z` with no-daylight-saving local time)
- [x] Smoke test

Waiting for splunk/sc4s# 2028 to be released.

Resolves axoflow/syslog-ng-testing#8


MrAnno:
splunk/sc4s# 2028 is needed to make the smoke test green.

.github/workflows/sc4s-stable.yml
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            if (context.ref.startsWith('refs/tags/sc4s-')) {
+              const tag = "v" + context.ref.replace("refs/tags/sc4s-", "");
+              console.log('Version from tag: ' + tag);
+              core.setOutput('ref', tag);
+              return;
+            }
+
+            const owner = '${{ inputs.owner || 'splunk' }}';
```
alltilla, line 68: [optional]

Can we factor out the fallback default values to `env:`, and only reference them here? Or maybe we could work solely from environment variables every time, and we could have a prepare step, which fills them from the input, if we are running on `workflow_distpatch`, else the hard coded defaults.

It is easy to forget to update the default values, if they are scattered in the workflow file.

alltilla (approved):
Awesome!

.github/workflows/sc4s-stable.yml
MrAnno: Good idea.

MrAnno: Fixed.

MrAnno:
Retest with timezone fix:
https://github.com/MrAnno/splunk-connect-for-syslog/actions/runs/4546283907/jobs/8014721593

MrAnno:
@alltilla I think we can merge this even if the smoke test is red.
Once splunk/sc4s# 2028 is released, it will start working.

Proof:
https://github.com/MrAnno/syslog-ng-docker/actions/runs/4546202385

MrAnno:
Some example CI executions:
https://github.com/MrAnno/syslog-ng-docker/actions

.github/workflows/ssc-draft-release.yml
```diff
@@ -0,0 +1,75 @@
...
+    env:
+      GH_TOKEN: ${{ github.token }}
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v3
+      - name: Report test issue
+        if: contains(needs.test.result, 'failure')
+        run: |
+          gh issue create --title "ssc-${{ needs.detect-version.outputs.new_version }}" \
+            --body "ssc-${{ needs.detect-version.outputs.new_version }} test failed"
```
alltilla, line 71: [nitpick]

This steps seems strange in the `draft-release` job. To me it feels like this should be a separate job, and we shouldn't create a draft release, if the test failed.

.github/workflows/ssc-test.yml
```diff
@@ -0,0 +1,142 @@
...
+          docker run -d --network host \
+            -e SC4S_DEST_SPLUNK_HEC_DEFAULT_URL="https://127.0.0.1:8088" \
+            -e SC4S_DEST_SPLUNK_HEC_DEFAULT_TOKEN="70b6ae71-76b3-4c38-9597-0c5b37ad9630" \
+            -e SC4S_DEST_SPLUNK_HEC_DEFAULT_TLS_VERIFY="no" \
+            -e SC4S_DEBUG_CONTAINER="yes" \
+            ssc:test
+
+          pip3 install poetry
+          poetry install
+          mkdir -p test-results || true
```
alltilla, line 140: I think the `|| true` is unnecessary here:
```
$ man mkdir | grep -A1 '\-p'
       -p, --parents
              no error if existing, make parent directories as needed
```

alltilla (commented):
Awesome! I only had one observation, which I would like to discuss, but otherwise LGTM.

.github/workflows/ssc-draft-release.yml
MrAnno: I wouldn't add a separate job just for this for now.
The draft will be created unconditionally, so the name at least does not lie, it just has a side effect.

alltilla (dismissed):

.github/workflows/ssc-test.yml
MrAnno: Fixed.

alltilla (approved):

PR #63, opened 2024-02-10
RFC: Generalize helm chart

bazsi:


charts/axosyslog/templates/service.yaml
```diff
@@ -1,19 +1,46 @@
...
 kind: Service
 metadata:
-  name: "{{ template "axosyslog.fullname" . }}"
+  name: {{ template "axosyslog.fullname" . }}-syslog
 spec:
   selector:
-    app: "{{ template "axosyslog.fullname" . }}"
-    release: {{ .Release.Name | quote }}
+    app: {{ template "axosyslog.fullname" . }}-syslog
+  type: NodePort
```
orymate, line 9: this could be configurable. ClusterIP (the default) and LoadBalancer make sense too

charts/axosyslog/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: axosyslog
+description: AxoSyslog for Kubernetes
+type: application
+version: 0.8.1
+appVersion: "4.6.0"
```
adrianrobotka, line 6: It's fine as it is. I'd like to just point out that the release pipeline could also fill out this value (to avoid manual errors).

adrianrobotka (approved):

charts/axosyslog/Chart.yaml
bazsi: I've now added a local shell script to bump the syslog-ng version number, which should eventually become a release pipeline.

charts/axosyslog/templates/service.yaml
bazsi: can you point me to a sample how this is usually configured in other charts?

bazsi:
I am merging this as is, it's already a lot more useful than before.

PR #70, opened 2024-03-22
Autoremove old nightly images

MrAnno:
Keeping only the last 30 days (snapshot + snapshot-dbg images).


alltilla (approved):
Really cool, thanks!

PR #73, opened 2024-04-11
syslog-ng: install package signing public key after generation

MrAnno:


alltilla (approved):
Thanks! I tried the same, but was missing the passwd part, now I know how to do it :)

PR #82, opened 2024-04-27
syslog-ng: add webhook() and webhook-json() sources

MrAnno:


syslog-ng/python-modules/webhook/source.py
```diff
@@ -0,0 +1,139 @@
...
+    def init(self, options: dict[str, Any]) -> bool:
+        self.logger = logging.getLogger("http")
+        self.set_transport_name("http")
+        if not self.init_options(options):
+            return False
+
+        self.ssl_ctx = None
+        if self.tls_key_file:
+            self.ssl_ctx = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
+            self.ssl_ctx.load_cert_chain(self.tls_cert_file, self.tls_key_file)
```
alltilla, line 78: As far as I can understand, without CAs loaded we do not verify the client's cert. Do I understand right? Don't we want to do so?

For reference:
 * https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_default_certs
 * https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_verify_locations

syslog-ng/python-modules/webhook/source.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def get_current_user(self):
+        if not self.source.auth_token:
+            return True
+
+        token = self.request.headers.get("Authorization", "").split(" ")
+        if len(token) != 2:
+            return False
+
+        token = token[1]
```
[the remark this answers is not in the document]
alltilla: Nevermind, I have forgot to add `Bearer`

syslog-ng/python-modules/webhook/source.py
alltilla: We don't want this right now.

alltilla (dismissed):

syslog-ng/python-modules/webhook/source.py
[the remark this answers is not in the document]
MrAnno: Note: works with "Basic" too.
For now, the auth method doesn't matter, we accept anything and use it as a "shared secret" method.

syslog-ng/python-modules/webhook/source.py
MrAnno: I'm adding this, just in case anyone needs it.

alltilla (approved):

PR #89, opened 2024-05-15
syslog-ng/webhook: add SOURCEIP and URL path+query param enrichment

MrAnno:
Depends on syslog-ng/syslog-ng 4964


MrAnno:
Oh jeez, I've pushed the wrong repo again...

PR #97, opened 2024-05-17
filterx: various small fixes

alltilla:

### DIFF
--- a/lib/filterx/object-primitive.c
+++ b/lib/filterx/object-primitive.c
@@ -317,7 +317,7 @@ static gboolean
 _repr(FilterXObject *s, GString *repr)
 {
   LogMessageValueType t;
-  return filterx_object_marshal(s, repr, &t);
+  return filterx_object_marshal_append(s, repr, &t);
 }
 
 FILTERX_DEFINE_TYPE(integer, FILTERX_TYPE_NAME(object),
--- a/lib/filterx/object-primitive.c
+++ b/lib/filterx/object-primitive.c
@@ -106,9 +106,10 @@ _double_map_to_json(FilterXObject *s, struct json_object **object, FilterXObject
 gboolean
 double_repr(double val, GString *repr)
 {
-  gchar buf[G_ASCII_DTOSTR_BUF_SIZE];
-  g_ascii_dtostr(buf, sizeof(buf), val);
-  g_string_append(repr, buf);
+  gsize init_len = repr->len;
+  g_string_set_size(repr, init_len + G_ASCII_DTOSTR_BUF_SIZE);
+  g_ascii_dtostr(repr->str + init_len, G_ASCII_DTOSTR_BUF_SIZE, val);
+  g_string_set_size(repr, init_len + strlen(repr->str + init_len));
   return TRUE;
 }
 
--- a/lib/filterx/tests/test_expr_comparison.c
+++ b/lib/filterx/tests/test_expr_comparison.c
@@ -533,10 +533,17 @@ Test(expr_comparison, test_string_to_datetime_string_based_comparison)
                      FCMPX_NE | FCMPX_STRING_BASED, TRUE);
 }
 
-// FCMPX_TYPE_AWARE tests
-// in case of LHS type is one of: [string, bytes, json, protobuf, message_value], uses FCMPX_STRING_BASED comparsion
-// in case of any of LHS or RHS is filterx null, compares filterx types (null less than any)
-// uses FCMPX_NUM_BASED in any other cases
+/*
+ * Type aware tests.
+ *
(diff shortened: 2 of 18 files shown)


lib/filterx/object-primitive.c
```diff
@@ -106,9 +106,10 @@ _double_map_to_json(FilterXObject *s, struct json_object **object, FilterXObject
 gboolean
 double_repr(double val, GString *repr)
 {
-  gchar buf[G_ASCII_DTOSTR_BUF_SIZE];
-  g_ascii_dtostr(buf, sizeof(buf), val);
-  g_string_append(repr, buf);
+  gsize init_len = repr->len;
+  g_string_set_size(repr, init_len + G_ASCII_DTOSTR_BUF_SIZE);
+  g_ascii_dtostr(repr->str + init_len, G_ASCII_DTOSTR_BUF_SIZE, val);
+  g_string_set_size(repr, strchr(repr->str + init_len, '\0') - repr->str);
```
MrAnno, line 112: 0 is always added by g_ascii_dtostr

lib/filterx/tests/test_expr_comparison.c
```diff
@@ -552,14 +559,13 @@ Test(expr_comparison, test_null_cases_type_aware_comparison)
...
-  _assert_comparison(filterx_string_new("foobar", 6), filterx_null_new(), FCMPX_GT | FCMPX_TYPE_AWARE, TRUE);
+  _assert_comparison(filterx_string_new("foobar", 6), filterx_null_new(), FCMPX_GT | FCMPX_TYPE_AWARE, FALSE);
   _assert_comparison(filterx_string_new("foobar", 6), filterx_null_new(), FCMPX_NE | FCMPX_TYPE_AWARE, TRUE);
 }
 
 Test(expr_comparison, test_string_cases_type_aware_comparison)
 {
-  // TODO: fix float precision error in g_ascii_dtostr 3.14 -> "3.1400000000000000000000001"
-  // _assert_comparison(filterx_string_new("3.14", 4), filterx_double_new(3.14), FCMPX_EQ | FCMPX_TYPE_AWARE, TRUE);
+  _assert_comparison(filterx_string_new("3.14", 4), filterx_double_new(3.14), FCMPX_EQ | FCMPX_TYPE_AWARE, TRUE);
```
MrAnno, line 568: How was this resolved?

lib/filterx/tests/test_expr_comparison.c
```diff
@@ -570,29 +576,29 @@ Test(expr_comparison, test_string_cases_type_aware_comparison)
   // check if compared as string
   _assert_comparison(filterx_string_new("a443", 4), filterx_integer_new(443), FCMPX_EQ | FCMPX_TYPE_AWARE, FALSE);
   _assert_comparison(filterx_string_new("a443", 4), filterx_integer_new(443), FCMPX_LT | FCMPX_TYPE_AWARE, FALSE);
-  _assert_comparison(filterx_string_new("a443", 4), filterx_integer_new(443), FCMPX_GT | FCMPX_TYPE_AWARE, TRUE);
+  _assert_comparison(filterx_string_new("a443", 4), filterx_integer_new(443), FCMPX_GT | FCMPX_TYPE_AWARE, FALSE);
   _assert_comparison(filterx_string_new("a443", 4), filterx_integer_new(443), FCMPX_NE  | FCMPX_TYPE_AWARE, TRUE);
 
   // bytes - boolean
-  _assert_comparison(filterx_bytes_new("true", 4), filterx_boolean_new(true), FCMPX_EQ | FCMPX_TYPE_AWARE, TRUE);
+  _assert_comparison(filterx_bytes_new("true", 4), filterx_boolean_new(true), FCMPX_EQ | FCMPX_TYPE_AWARE, FALSE);
```
MrAnno, line 583: This would have been great, but I guess it's not a big deal as it works with strings.

lib/filterx/tests/test_expr_comparison.c
alltilla: Now we compare the lhs and rhs type, and they are different. This means that we no longer compare them as strings, where this 000001 matters, but we compare them as doubles, where the DBL_EPSILON check works as intended.

lib/filterx/tests/test_expr_comparison.c
alltilla: We can write custom parsing logic for these, and convert them to 0 or 1 numbers then follow through with the number comparison. It might make sense.

lib/filterx/tests/test_expr_comparison.c
MrAnno: Thanks for the explanation.

PR #103, opened 2024-05-17
autotools: fix pkg* paths

MrAnno:


MrAnno:
Sorry, I didn't notice about this. I'm merging it ASAP as it installs to the wrong directories.

PR #107, opened 2024-05-18
package-indexer: S3 and Cloudflare

alltilla:
Not my proudest commit history, sorry.

### DIFF
--- a/packaging/package-indexer/remote_storage_synchronizer/s3_bucket_synchronizer.py
+++ b/packaging/package-indexer/remote_storage_synchronizer/s3_bucket_synchronizer.py
@@ -22,12 +22,12 @@
 
 from hashlib import md5
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Set
+from typing import Any, Dict, List
 
 from boto3 import Session
 from botocore.exceptions import ClientError, EndpointConnectionError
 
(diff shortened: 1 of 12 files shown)


alltilla:
let me do one refactor round before merging

alltilla:
> let me do one refactor round before merging

done

packaging/package-indexer/remote_storage_synchronizer/s3_bucket_synchronizer.py
```diff
@@ -0,0 +1,296 @@
...
+            "Downloading file.",
+            remote_path=relative_file_path,
+            local_path=str(download_path),
+        )
+
+        download_path.parent.mkdir(parents=True, exist_ok=True)
+        with download_path.open("wb") as downloaded_object:
+            self.__client.download_fileobj(self.__bucket, relative_file_path, downloaded_object)
+
+        md5sum = md5(download_path.read_bytes()).digest()
```
MrAnno, line 129: Since we are the ones calculating the remote hashes, can't we use something more modern and trusted, like sha256?
I know it's not important in this use case because we have endless layers of integrity checks (and PGP signature as well), but it would just look better if we didn't see md5 in this unit.

alltilla: Sure, done!

MrAnno (approved):

PR #113, opened 2024-05-21
App specific transformations

bazsi:


alltilla:
Rebased to main.

modules/appmodel/app-transform-generator.h
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (c) 2024 Balazs Scheidler <balazs.scheidler@axoflow.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
```
alltilla, line 5: @bazsi Is this intentionally LGPL?

modules/appmodel/app-parser-generator.c
```diff
@@ -175,7 +194,7 @@ _generate_framing(AppParserGenerator *self, GlobalConfig *cfg)
         g_string_append(self->block, "        channel {\n");
 
       g_string_append(self->block,
-                      "            filter { tags('.app.doesnotexist'); };\n"
+                      "            filterx { false; };\n"
```
alltilla, line 197: `_generate_action()` can still generate "traditional" `rewrite`, too.

I think we should either support both everywhere or support only filterx.
This is true for `_generate_empty_frame()`, too.

What do you think?

alltilla: Talked about this IRL. Functionally these 2 are the same, but `filterx` is quicker if we ever used a `filterx` block, as the simple `filter` block would force a `LogMessage` sync.

alltilla:
@bazsi can you please check the last 3 commits? I am okay with the rest of the PR.

alltilla (approved):

PR #114, opened 2024-05-21
filterx: otel dict

alltilla:


bazsi (approved):
Approved based on my review in syslog-ng/syslog-ng

PR #116, opened 2024-05-21
Filterx parse csv

bshifter:
CSVParser functionality ported to filterx:

filterx-func-parse-csv function added with the support of:
- custom column names
- delimiters
- dialects
- flag: greedy
- flag: strip_whitespace

It uses the new 'named function arguments' to handle options, so all the option arguments are optional. The only positional argument supported by the function is the target string to be parsed.

known issues:
filterx function framework does not support inline json literals, this way the custom columns argument need to be built in a separate command (see below). however, the interpretation of column's values can not be made on preprocess, so the columns' expression evaluation happens at run time (on every _eval). This will be fixed in the future, since this workaround's performance cost is relative high. 

example usage:

`filterx {
                        cols = json_array(["foo", "bar", "baz"]);
                        res = parse_csv($MSG,
                                columns=cols,
                                delimiters=",;:",
                                dialect="escape-backslash-with-sequences",
                                greedy=true,
                                strip_whitespace=false);
}`


modules/csvparser/filterx-func-parse-csv.c
```diff
@@ -0,0 +1,346 @@
...
+    {
+      if (!obj)
+        {
+          error_str = FILTERX_FUNC_PARSE_CSV_ARG_NAME_GREEDY " argument evaluation error";
+          goto error;
+        }
+      if (filterx_object_truthy(obj))
+        opt_flags |= CSV_SCANNER_GREEDY;
+      else
+        opt_flags &= ~CSV_SCANNER_GREEDY;
```
alltilla, line 266: I would use `filterx_function_args_get_named_literal_boolean()` here, to make sure it is a boolean object. Same for `strip_whitespaces`.

modules/csvparser/filterx-func-parse-csv.c
```diff
@@ -0,0 +1,346 @@
...
+    return TRUE;
+  FilterXObject *cols_obj = filterx_expr_eval(self->columns);
+  if (!cols_obj)
+    return FALSE;
+
+  if (filterx_object_is_type(cols_obj, &FILTERX_TYPE_NAME(null)))
+    {
+      result = TRUE;
+      goto exit;
+    }
```
alltilla, line 64: I don't think we should support these now that `columns` is an optional argument. I can't imagine someone setting `columns=null`.

PR #124, opened 2024-05-24
Open qdisk dirlock in rw mode

bazsi:
As reported by @akunszt, this should fix creating disk buffer files on NFS.


MrAnno (approved):

MrAnno:
A news entry will be needed.

PR #127, opened 2024-05-24
s3: add server side encryption support

akunszt:
This PR enables to use the server side encryption on the s3() destination. It is a common pattern that data stored on the S3 bucket is encrypted at the rest.

S3 supports multiple server side encryption possibilities. This is only for the `aws:kms` method. In theory `aws:kms:dsse` should also work but as it uses the same arguments but I didn't test it so I disabled that intentionally.

Check the [s3api](https://docs.aws.amazon.com/cli/latest/reference/s3api/create-multipart-upload.html) documentation for the details.

This means that the server-side-encryption() option accepts only `aws:kms` or an empty string, which is the default.

The kms-key() should be:
* an ID of a key
* an alias of a key, but in that case you have to add the `alias/` prefix
* an ARN of a key

For example:
```
destination d_s3 {
        s3(
                bucket( "log-archive-bucket" )
                object-key( "logs/syslog" )
                server-side-encryption( "aws:kms" )
                kms-key( "alias/log-archive" )
        );
};
```

To be able to use the `aws:kms` encryption the AWS Role or User has to have the following permissions on the given key:

* kms:Decrypt
* kms:Encrypt
* kms:GenerateDataKey

Check [this](https://repost.aws/knowledge-center/s3-large-file-encryption-kms-key) page on why the kms:Decrypt is mandatory.

Setting server-side-encryption() to `aws:kms` without setting kms-key() will terminate the syslog-ng process.

Setting kms-key() without setting server-side-encryption() emits a warning and ignores the kms-key(). The server side encryption will be disabled in this case.

It also worth mentioning that kms-key() cannot be changed after the create_multipart_upload executed. I am not sure what will happen if someone modifies the key and then just reloads the syslog-ng configuration. A normal restart would work as syslog-ng finishes the multipart upload during the shutdown.

### DIFF
--- a/modules/python-modules/syslogng/modules/s3/s3_object.py
+++ b/modules/python-modules/syslogng/modules/s3/s3_object.py
@@ -165,8 +165,6 @@ class S3ObjectPersist:
                 "compress",
                 "compresslevel",
                 "chunk-size",
-                "server-side-encryption",
-                "kms-key",
                 "storage-class",
                 "finished",
                 "upload-id",
--- a/modules/python-modules/syslogng/modules/s3/s3_object.py
+++ b/modules/python-modules/syslogng/modules/s3/s3_object.py
@@ -156,6 +156,16 @@ class S3ObjectPersist:
                     raise PersistLoadError from e
 
         if path:
+            # fields added in later releases can be missing from the persist JSON
+            # set a sane default value which can be overriden later if the user set
+            # the appropriate configuration directives
+            for upgrade_field in {
+                "canned-acl",
+                "kms-key",
+                "server-side-encryption",
+            }:
+                cache[upgrade_field] = cache.get(upgrade_field, "")
+
             for field in {
                 "persist-name",
                 "bucket",
@@ -170,12 +180,16 @@ class S3ObjectPersist:
                 "upload-id",
                 "uploaded-parts",
                 "pending-parts",
+                "canned-acl",
+                "kms-key",
+                "server-side-encryption",
             }:
                 try:
                     cache[field]
                 except KeyError as e:
                     raise PersistLoadError from e
 
+        # fields
         self.__persist_name: str = cache.get("persist-name", persist_name)
         self.__bucket: str = cache.get("bucket", bucket)
         self.__target_key: str = cache.get("target-key", target_key)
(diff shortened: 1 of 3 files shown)


modules/python-modules/syslogng/modules/s3/s3_object.py
```diff
@@ -161,6 +165,8 @@ def __init__(
                 "compress",
                 "compresslevel",
                 "chunk-size",
+                "server-side-encryption",
+                "kms-key",
```
bazsi, line 169: I think this will cause existing JSON files to load.

e.g. if the JSON file was saved by a previous version of the S3 destination and we add this code, then the old file will not load, complaining about these missing fields.

These should be made optional, I think.

bazsi:
I have found just one problem that I pointed out in a review comment above. Other than that this looks good to me.

modules/python-modules/syslogng/modules/s3/s3_object.py
akunszt: Thanks. If I understand this correctly I only had to remove those entries from the `field` variable. Please correct me if I am wrong.

bazsi: You are right, and I can see you removed that check, with that this should be ready to be merged.

alltilla: If we are loading the JSON and do not have the `"kms-key"` field in it, `self.__kms_key` will be set to `None` by `self.__kms_key: str = cache.get("kms-key", kms_key)`, where we expect it to be a `str`. 

I think we should have an "upgrade" step, where we give defaults (`""`) to the JSON.

Something like:
```py
        if path:
            for upgrade_field in {
                "canned-acl",
                "server-side-encryption",
                "kms-key",
            }:
                cache[upgrade_field] = cache.get(upgrade_field, "")

            for field in {
                "persist-name",
                "bucket",
                "target-key",
                "timestamp",
                "index",
                "compress",
                "compresslevel",
                "chunk-size",
                "storage-class",
                "finished",
                "upload-id",
                "uploaded-parts",
                "pending-parts",
                "canned-acl",
                "server-side-encryption",
                "kms-key",
            }:
                try:
                    cache[field]
                except KeyError as e:
                    raise PersistLoadError from e
```

Did not test it though.

alltilla: This `__init__` function became really cryptic because of the 2 ctor paths, sorry for that.

alltilla (approved):
Thank you! :)

PR #130, opened 2024-05-28
lib/ivykis: update ivykis to include the epoll_pwait2() fix

bazsi:
ivykis 0.43 started supporting epoll_pwait2() which is not permitted in docker environments in older distributions (RHEL7 came up, as well as Debian buster)

This branch updates our ivykis submodule to a branch in the ivykis upstream, which will hopefully be released as 0.43.1 soon enough. 

I'd wait a few days for ivykis 0.43.1 to be released, once that happens we could use the official commit. @buytenh


buytenh:
I merged the ivykis fix referenced in this PR to the ivykis master branch, so the commit ID should remain stable.

I'm going to try to see if I will have some time this weekend to reproduce and maybe fix https://github.com/buytenh/ivykis/issues/32 .  If I manage that, I'll release 0.43.1 with a fix for that issue included.

bazsi:
I've updated this branch to contain both patches. The official tag is not there, but both of the patches are on master already, so this should be good enough for merging without changing the submodule ref.

alltilla:
https://github.com/buytenh/ivykis/tree/v0.43.1 was released a couple of hours ago. I think it would be best to update to that tag.

alltilla:
Thanks for the heads up, I updated to `v0.43.1-trunk`.

alltilla (approved):

PR #132, opened 2024-05-31
Filterx format csv

bshifter:
add csv formatter function to filterx, helps re-format csv data parsed by parse_csv previously
- also accepts json_array and json_object as input which are the possible output types for parse_csv filter func
- supports optional named argument for delimiter character
- supports optional columns argument with json_object input type for column filtering/ordering
- ignores column names with json_array input
-  uses json_object's key ordering when column names are not set (which is seemingly correct order)
-  auto double-quotes column values when they contain delimiter character
- auto escapes double quotes when double-quoting a column value

examples:

```
filterx {
...
  # array to unnamed cols
  x = parse_csv($MSG); # returns json_array as '["foo","bar","baz"]'
  $MSG = format_csv(x, delimiter=","); # custom delimiter
...
}
```
```
filterx {
...
  # dict to unnamed cols
  cols = json_array(["future_use1","receive_time","serial_number","type"]);
  x = parse_csv($MSG, columns=cols); # returns json_object as '{"future_use1":"foo"...}'
  $MSG = format_csv(x); # uses json_oject's key ordering
...
}
```
```
filterx {
...
  #dict to named cols
  cols = json_array(["future_use1","receive_time","serial_number","type"]);
  x = parse_csv($MSG, columns=cols);
  cols2 = json_array(["type", "receive_time"]);
  $MSG = format_csv(x, delimiter=" ", columns=cols2); # filters and orders columns by cols2 
...
}
```

### DIFF
--- a/modules/csvparser/filterx-func-format-csv.c
+++ b/modules/csvparser/filterx-func-format-csv.c
@@ -1,0 +1,256 @@
+/*
+ * Copyright (c) 2024 shifter
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 as published
+ * by the Free Software Foundation, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "filterx-func-format-csv.h"
+#include "filterx/expr-literal.h"
+#include "filterx/object-string.h"
+#include "filterx/object-null.h"
+#include "filterx/object-dict-interface.h"
+#include "filterx/object-list-interface.h"
+#include "filterx/filterx-eval.h"
+
+#include "scratch-buffers.h"
+#include "utf8utils.h"
+
+#define FILTERX_FUNC_FORMAT_CSV_USAGE "Usage: format_csv($input(list or dict), delimiter=\",\", columns=$columns(list))"
+
+typedef struct FilterXFunctionFormatCSV_
+{
+  FilterXFunction super;
+  FilterXExpr *input;
+  gchar delimiter;
+  FilterXExpr *columns;
+} FilterXFunctionFormatCSV;
+
+static gboolean
+_append_to_buffer(FilterXObject *key, FilterXObject *value, gpointer user_data)
+{
+  if (!value)
+    return FALSE;
+
+  FilterXFunctionFormatCSV *self = ((gpointer *) user_data)[0];
+  GString *buffer = ((gpointer *) user_data)[1];
+
+  if (filterx_object_is_type(value, &FILTERX_TYPE_NAME(dict)) ||
+      filterx_object_is_type(value, &FILTERX_TYPE_NAME(list)))
+    {
+      msg_debug("FilterX: format_csv(): skipping object, type not supported",
+                evt_tag_str("type", value->type->name));
+      return TRUE;
+    }
+
+  if (buffer->len)
+    g_string_append(buffer, &self->delimiter);
+
+  gsize len_before_value = buffer->len;
+  if (!filterx_object_repr_append(value, buffer))
+    return FALSE;
+
+  /* TODO: make the characters here configurable. */
+  if (memchr(buffer->str + len_before_value, self->delimiter, buffer->len - len_before_value) != NULL)
+    {
+      ScratchBuffersMarker marker;
+      GString *value_buffer = scratch_buffers_alloc_and_mark(&marker);
+
+      g_string_assign(value_buffer, buffer->str + len_before_value);
+      g_string_truncate(buffer, len_before_value);
+      g_string_append_c(buffer, '"');
+      append_unsafe_utf8_as_escaped_binary(buffer, value_buffer->str, value_buffer->len, "\"");
+      g_string_append_c(buffer, '"');
+
+      scratch_buffers_reclaim_marked(marker);
+    }
+
+  return TRUE;
+}
+
+static gboolean
+_handle_list_input(FilterXFunctionFormatCSV *self, FilterXObject *csv_data, GString *formatted)
+{
+  guint64 size;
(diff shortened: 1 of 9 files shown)


modules/csvparser/filterx-func-format-csv.c
```diff
@@ -0,0 +1,255 @@
...
+
+FilterXFunction *
+filterx_function_format_csv_new(const gchar *function_name, FilterXFunctionArgs *args, GError **error)
+{
+  FilterXFunctionFormatCSV *self = g_new0(FilterXFunctionFormatCSV, 1);
+  filterx_function_init_instance(&self->super, function_name);
+
+  self->super.super.eval = _eval;
+  self->super.super.free_fn = _free;
+  self->delimiter = ',';
```
bazsi, line 237: the default delimiter for the parse_csv() function is a space, so I think we should match that here.

modules/csvparser/filterx-func-format-csv.c
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+  gboolean success = FALSE;
+  GString *formatted = scratch_buffers_alloc();
+
+  if (filterx_object_is_type(csv_data, &FILTERX_TYPE_NAME(list)))
+    {
+      guint64 size;
+      if (!filterx_object_len(csv_data, &size))
+        return FALSE;
```
bazsi, line 105: we should return NULL here, probably.

modules/csvparser/filterx-func-format-csv.c
```diff
@@ -0,0 +1,255 @@
...
+
+static FilterXObject *
+_eval(FilterXExpr *s)
+{
+  FilterXFunctionFormatCSV *self = (FilterXFunctionFormatCSV *) s;
+
+  FilterXObject *csv_data = filterx_expr_eval_typed(self->input);
+  if (!csv_data)
+    {
+      filterx_eval_push_error("Failed to evaluate input. " FILTERX_FUNC_FORMAT_CSV_USAGE, s, NULL);
```
bazsi, line 94: I think we don't need to push the error here as filterx_expr_eval() would set that already.

modules/csvparser/filterx-func-format-csv.c
```diff
@@ -0,0 +1,255 @@
...
+
+      gpointer user_data[] = { self, formatted };
+      success = TRUE;
+      for (guint64 i = 0; i < size && success; i++)
+        {
+          FilterXObject *elt = filterx_list_get_subscript(csv_data, i);
+          success = _append_to_buffer(NULL, elt, user_data);
+          filterx_object_unref(elt);
+        }
+    }
```
bazsi, line 115: [optional] I'd extract the specializations to separate functions to make this a bit easier to follow.

modules/csvparser/filterx-func-format-csv.c
```diff
@@ -0,0 +1,255 @@
...
+              success = TRUE;
+              for (guint64 i = 0; i < size && success; i++)
+                {
+                  FilterXObject *col = filterx_list_get_subscript(cols, i);
+                  FilterXObject *elt = filterx_object_get_subscript(csv_data, col);
+                  success = _append_to_buffer(col, elt, user_data);
+                  filterx_object_unref(col);
+                  filterx_object_unref(elt);
+                }
+            }
```
bazsi, line 137: we should handle if columns is not a list

bazsi: now, I can see that in that case, success is FALSE, so that's what we would do.

I think extracting the type specific specializations would have avoided my misunderstanding.

modules/csvparser/filterx-func-format-csv.c
```diff
@@ -0,0 +1,255 @@
...
+      else
+        {
+          gpointer user_data[] = { self, formatted };
+          success = filterx_dict_iter(csv_data, _append_to_buffer, user_data);
+        }
+    }
+  else
+    {
+      filterx_eval_push_error("input must be a dict or list. " FILTERX_FUNC_FORMAT_CSV_USAGE, s, csv_data);
+      filterx_object_unref(csv_data);
```
bazsi, line 149: we do the same in the epilogue part, as sucess is FALSE in this case.

modules/csvparser/filterx-func-format-csv.c
```diff
@@ -0,0 +1,255 @@
...
+    {
+      ScratchBuffersMarker marker;
+      GString *value_buffer = scratch_buffers_alloc_and_mark(&marker);
+
+      g_string_assign(value_buffer, buffer->str + len_before_value);
+      g_string_truncate(buffer, len_before_value);
+      g_string_append_c(buffer, '"');
+      append_unsafe_utf8_as_escaped_binary(buffer, value_buffer->str, value_buffer->len, "\"");
+      g_string_append_c(buffer, '"');
+
```
bazsi, line 79: this is kind of the "dialect" parameter we already support on the parse_csv() side. the dialect would determine what kind of quotation we produce on the output side.

bazsi (commented):
This is looking good, I had a few comments, some of which would be worth addressing before merging this.

jszigetvari:
@bshifter Just an idea:
If we try to format a set of fields to csv, and at a certain point all the remaining fields would be empty strings, then I think it would be safe to not print any more fields or separators from that point on. Thus we could stop the formatting of that message right at that place.

jszigetvari:
> @bshifter Just an idea: If we try to format a set of fields to csv, and at a certain point all the remaining fields would be empty strings, then I think it would be safe to not print any more fields or separators from that point on. Thus we could stop the formatting of that message right at that place.

After discussing this with @bshifter , it seems better to let this approach go.

modules/csvparser/filterx-func-format-csv.c
```diff
@@ -0,0 +1,261 @@
...
+  gboolean success = TRUE;
+  gpointer user_data[] = { self, formatted };
+  if (self->columns)
+    {
+      FilterXObject *cols = filterx_expr_eval(self->columns);
+      if (filterx_object_is_type(cols, &FILTERX_TYPE_NAME(list)))
+        {
+          guint64 size;
+          if (!filterx_object_len(cols, &size))
+            return FALSE;
```
alltilla, line 116: This leaks `cols`.

modules/csvparser/filterx-func-format-csv.c
```diff
@@ -0,0 +1,261 @@
...
+    {
+      FilterXObject *elt = filterx_list_get_subscript(csv_data, i);
+      success = _append_to_buffer(NULL, elt, user_data);
+      filterx_object_unref(elt);
+    }
+  return success;
+}
+
+static gboolean
+_handle_dict_input(FilterXFunctionFormatCSV *self, FilterXObject *csv_data, GString *formatted)
```
alltilla, line 105: [optional]

Usually we functions are written with early returns in axosyslog, as it is more readable. Can you reorganize this function? Thanks!

alltilla (commented):
This is only a partial review. I will continue the review tomorrow.

bazsi (approved):

PR #136, opened 2024-06-02
Add support for ssl peer fingerprint

bazsi:
This resolves https://github.com/syslog-ng/syslog-ng/issues/4978

### DIFF
--- a/lib/transport/tls-session.c
+++ b/lib/transport/tls-session.c
@@ -523,7 +523,7 @@ void
 tls_session_info_callback(const SSL *ssl, int where, int ret)
 {
   TLSSession *self = (TLSSession *)SSL_get_app_data(ssl);
-  if( !self->peer_info.found && where == (SSL_ST_ACCEPT|SSL_CB_LOOP) )
+  if (!self->peer_info.found && where == (SSL_ST_ACCEPT|SSL_CB_LOOP))
     {
       X509 *cert = SSL_get_peer_certificate(ssl);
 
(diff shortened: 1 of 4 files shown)


lib/transport/tls-session.c
```diff
@@ -195,11 +198,30 @@ tls_session_verify(TLSSession *self, int ok, X509_STORE_CTX *ctx)
...
+                         tls_context_format_location_tag(self->ctx));
+              return 0;
+            }
+        }
+      else
+        {
+          /* this is a trust anchor that forces the key to be valid */
+          if (tls_session_verify_fingerprint(ctx))
+            {
+              msg_notice("Certificate accepted due to being present on trusted-keys()",
```
MrAnno, line 220: I think we should not change the behavior of ` trusted-keys()` because, from the perspective of a user who completely understands the old behavior and just upgrades to AxoSyslog, it would be considered a security issue.

I can see that forcibly accepting an otherwise "invalid" key based on its fingerprint can be a real use case (I have some doubts about the validity of such a need), but in that case, I think it should be introduced as a new option.
I understand that the name `trusted-keys()` would be perfect for such a new option, but I think we missed this opportunity years ago.

Changing the behavior based on config-version is another option, but I wouldn't do that either because of the security nature of this matter.

MrAnno: I was not completely right here: if the fingerprint alg is chosen well, it is suitable to replace the CA verification, and that can be a perfectly valid use case.

With this new implementation, we should not default to SHA1, and we should consider not letting users set a weak hash function.
To avoid any confusion, I'd remove the old option and introduce a new one, for example:

```
trusted-fingerprints(sha1, "aa:bb:cc", ...)
```

bazsi:
I've dropped the change in trusted-keys() validation, as we agreed with @MrAnno that it should be introduced with a new option (trusted-fingerprints() or something like that).

By dropping this patch, this PR became very simple, _if_ there's trusted-keys() based validation in a config, the key fingerprint is saved into ${.tls.x509_fp} along with the CA and similar parameters.

With that this is mergable.

lib/transport/tls-session.c
bazsi: I've dropped this patch from the series and will fold it into #137

MrAnno (approved):

PR #137, opened 2024-06-02
Add support for fingerprint alg

### LINKED ISSUE #4976: trusted-keys: support a secure hash algorithm
## Description of the problem
https://syslog-ng.github.io/admin-guide/100_TLS-encrypted_message_transfer/004_TLS_options#trusted-keys says it uses SHA-1, and if I understand [the code](https://github.com/syslog-ng/syslog-ng/blob/9adfc8e80fc2b896bdea6a6412eccb6f184c9226/lib/transport/tls-session.c#L51) correctly, that's the only hash algorithm it supports. From https://en.wikipedia.org/wiki/SHA-1:

> Since 2005, SHA-1 has not been considered secure against well-funded opponents;[11] as of 2010 many organizations have recommended its replacement.[12][10][13] NIST formally deprecated use of SHA-1 in 2011 and disallowed its use for digital signatures in 2013, and declared that it should be phased out by 2030.[14] As of 2020, chosen-prefix attacks against SHA-1 are practical.[6][8] As such, it is recommended to remove SHA-1 from products as soon as possible and instead use SHA-2 or SHA-3. Replacing SHA-1 is urgent where it is used for digital signatures.

## Proposed solution
Add at least one other hash algorithm that's still considered secure. SHA-256 seems like a good candidate, but there are other good options.

## Alternatives
I guess it could work to support including the entire certificate in trusted-keys, instead of just the hash.

## Additional context
I'd guess that a practical attack would still be infeasible, assuming the attacker's goal is to create a second preimage certificate that they control the corresponding private key of.  I think they'd have to find a second preimage that is both a valid certificate and that they can get to include the public key corresponding to their private key. I'm not an expert at hash cryptanalysis, but that seems significantly more difficult to me than just a plain second preimage attack. I'm also not sure if sha-1's second preimage resistance has been successfully attacked at all yet. So this doesn't seem like a vulnerability to me, just a good idea to use hash algorithms that are generally considered secure.

bazsi:
This branch allows you to specify the fingerprint-alg() to be used for validating key fingerprints in trusted-keys()

Fixes: https://github.com/syslog-ng/syslog-ng/issues/4976


alltilla (commented):
The last commit lacks the signoff part. I don't think it is a good idea if I write it for you and force push, so please add it yourself. Thanks!

The code looks good to me. We need a newsfile entry, but I can make it for you.

bazsi:
I've rebased, fixed the signoff and added the news entry. btw, @MrAnno this PR now allows to trust specific keys, even if their X.509 validation fails. 

The code did not support it thus far, but I think this has always been the intention. Currently, trusted-keys() will only trust keys that are otherwise trusted, which does not make much sense to me.

The documentation is not that clear on this topic either (@fekete-robert).

bazsi:
I've dropped the change in trusted-keys() validation, as we agreed with @MrAnno that it should be introduced with a new option (trusted-fingerprints() or something like that).

By dropping this patch, this PR became very simple, _if_ there's trusted-keys() based validation in a config, the key fingerprint is saved into ${.tls.x509_fp} along with the CA and similar parameters.

With that this is mergable.

MrAnno:
From #136:

> I was not completely right here: if the fingerprint alg is chosen well, it is suitable to replace the CA-based verification, and that can be a perfectly valid use case.

> With this new implementation, we should not default to SHA1, and we should consider not letting users set a weak hash function.
To avoid any confusion, I'd remove the old option and introduce a new one, for example:

```
trusted-fingerprints(sha256, "aa:bb:cc", ...)
```

I can implement the missing parts next week.

bazsi:
I rebased and finished off the branch to do what was discussed in comments above.

bazsi:
This should be back on the review queue

lib/transport/tls-session.c
```diff
@@ -99,11 +100,34 @@ tls_session_find_issuer(TLSSession *self, X509 *cert)
...
+  if (!colon)
+    return FALSE;
+
+  gsize len = colon - fp;
+  if (len >= fp_alg_len - 1)
+    return FALSE;
+
+  memcpy(fp_alg, fp, len);
+  fp_alg[len] = 0;
+  if (len == 2 && isxdigit(fp_alg[0]) && isxdigit(fp_alg[1]))
```
MrAnno, line 116: Shouldn't it be `>=2`?

lib/transport/tls-session.c
```diff
@@ -297,6 +366,13 @@ tls_session_verify_callback(int ok, X509_STORE_CTX *ctx)
           break;
         }
     }
+  else if (self->cert_trusted_by_fingerprint)
+    {
+      msg_debug("Certificate invalid, but accepting due to being present on trusted-fingerprints()",
```
MrAnno, line 392: Do we want to allow trusted fingerprints for intermediate certs or the root CA too?

I have a feeling that we should not accept the connection if the fingerprint matches, for example, on the root CA.
It seems more like a misconfiguration. `cert_trusted_by_fingerprint` should only accept (or reject) the leaf in my opinion.

lib/transport/tls-session.c
bazsi: not really, this just validates that the initial field (right before the first colon) is an algorithm name or not. this is a more specific error message than "56" unknown digest algorithm.

lib/transport/tls-session.c
bazsi: Having a matching fingerprint is a very narrow and specific case, matching against the public key of the x.509 keypair.

I don't think we should do any additional validation in this case. This is the same as the ssh known_hosts authentication, where we _only_ have the fingerprint, this is very similar.

bazsi: ... maybe I misunderstood your comment, but we are only setting `cert_trusted_by_fingerprint` if the _leaf_ cert matches, see the only case where this variable is set to TRUE in the function `tls_session_verify`

bazsi: This is the chat I had with chatgpt about the topic:

https://chatgpt.com/share/68ff10ee-70e4-800d-b412-cf183bbcdbe2

Also, X509_digest() that we use validate the entire X.509 DER encoded structure, e.g. it includes expiration, and certificate properties as well. E.g. basicConstraints:CA value.

MrAnno (approved):

PR #139, opened 2024-06-03
Build arm64 images for dev branches

pepov:
We need arm64 images for testing and development on macs.


alltilla:
The `dev-main` job was created to have an image quickly, but adding arm64 would raise the job's runtime to ~2 hours. I think we should rather use the `nightly` tag, where arm64 is needed.

PR #140, opened 2024-06-03
json: fix copypaste error in cache_json_file() filterx fn

alltilla:


OverOrion (approved):
Thank you very much for the hotfix! :hot_pepper:

PR #148, opened 2024-06-06
Filterx csv parse format unify arguments

bazsi:
This PR matches up parse_csv() and format_csv() argument lists:

* the "delimiters" argument is renamed to "delimiter", even though it does allow multiple delimiters
* the default delimiter is set to the comma (as per the Comma Separated Values)


bshifter (approved):
lgtm

PR #153, opened 2024-06-07
Filterx handle unknown function arguments

bazsi:
This branch implements the handling of incorrect filterx function arguments, and also improves error handling aesthetics in multiple ways.

Here's a list of user-visible changes:
  * in some cases, the documentation/contact links were not printed on syntax errors, this was fixed
  * the primary error upon syntax error was easily lost in the log/error output, make it stand out
  * errors in function arguments are reported nicely, both when detected at compile time and at runtime

Internal changes:
  * it is now possible to set a filterx error condition where the extra info is a string and not a filterx object. Sometimes the filterx object is not available and creating one just for reporting an error would be an overhead. This should make it easier to report errors via the filterx error mechanism, instead of msg_error()
  * improve argument handling performance, use a GPtrArray to represent arg expressions, instead of a GList, which is iterated by index

### DIFF
--- a/modules/examples/filterx/example-filterx-func/example-filterx-func-plugin.c
+++ b/modules/examples/filterx/example-filterx-func/example-filterx-func-plugin.c
@@ -54,8 +54,4 @@ exit:
   return filterx_boolean_new(FALSE);
 }
 
-gpointer
-example_filterx_simple_func_construct_echo(Plugin *self)
-{
-  return (gpointer) &echo;
-}
+FILTERX_SIMPLE_FUNCTION(example_echo, echo);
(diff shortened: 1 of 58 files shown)


modules/examples/filterx/example-filterx-func/example-filterx-func-plugin.c
```diff
@@ -29,7 +29,7 @@
 #include "filterx/object-primitive.h"
 
 static FilterXObject *
-echo(GPtrArray *args)
+echo(FilterXExpr *s, GPtrArray *args)
```
alltilla, line 32: I think this change should be in `filterx: add filterx error handling to simple functions argument validation`.

alltilla (commented):
I only have one rebase related comment, otherwise LGTM.

Thanks for this!

modules/examples/filterx/example-filterx-func/example-filterx-func-plugin.c
alltilla: I have fixed this myself.

alltilla (approved):

PR #154, opened 2024-06-09
Implement rfc6587 auto detection

bazsi:
This is a pretty common misconfiguration that should be handled automatically. 

Just use syslog(transport(auto)) to automatically accept both octet counted and non-octet counted connections, this opens a TCP port without TLS, meaning it is an implementation of RFC6587.

### DIFF
--- a/lib/logproto/logproto-auto-server.c
+++ b/lib/logproto/logproto-auto-server.c
@@ -1,0 +1,147 @@
+/*
+ * Copyright (c) 2024 Balázs Scheidler <balazs.scheidler@axoflow.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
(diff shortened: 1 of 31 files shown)


lib/logproto/logproto-auto-server.c
```diff
@@ -0,0 +1,147 @@
...
+  else if (rc < 0)
+    return LPS_ERROR;
+
+  self->proto_impl = _construct_detected_proto(self, detect_buffer, rc);
+  if (self->proto_impl)
+    {
+      /* transport is handed over to the new proto */
+      self->super.transport = NULL;
+      return LPS_SUCCESS;
+    }
```
alltilla, line 121: Shouldn't we do a handshake after we autodetected the proto?

news/feature-4814.md
```diff
@@ -0,0 +1,5 @@
+`syslog()` source driver: add support for RFC6587 style auto-detection of
+octet-count based framing to avoid confusion that stems from the sender
+using a different protocol to the server.  This is automatically enabled for
+both the `transport(tcp)` and the `transport(tls)` settings.  You can disable
+auto-detection by using `transport(text)` or `transport(framed)`.
```
alltilla, line 5: We disable auto detection with `transport(force-framed)`, right?

bazsi:
This should wait until #156 is merged.

bazsi:
I have now rebased it but still need to check if it works on top of #156

alltilla:
#156 is now merged, I think we can give this another go (rebase, CI, review, ...)

bazsi:
This should now be ready for review.

Note: there's a followup PR as well, #361

bazsi:
I have now opted out of doing this by default. The auto-detection mechanism is only enabled with transport(auto)

MrAnno:
I've started reviewing this patchset, will finish it by the end of the week.

OverOrion (commented):
I've reviewed this and tested it manually as well, worked just fine :rocket:.
Is there a reason why it's still marked as `WIP`?

bazsi:
I pulled some of the later commits into this one, rebased against main and removed WIP.

@OverOrion if you can please re-approve, but note that I did pull in some newer changes into this.

PR #155, opened 2024-06-09
Logproto remove handshake in progress

bazsi:
The branch itself simplifies the LogProto handshake API and adds a unit test for the LogProtoAutoServer class (rfc6587)


bazsi:
I've rebased this now to sit on top of main, instead of the rfc6587 auto detection branch. With that this should be a lot easier to merge.

bazsi:
rebased against master. This is now a simple branch without much dependencies or consequences so should be ready to merge.

MrAnno (approved):

PR #162, opened 2024-06-11
macros: extract type information from MESSAGE, PROGRAM and HOST 

alltilla:
`MESSAGE` in particular can contain not just string data, but for example JSON as well.

This is necessary for the `$(format-json)` function to work properly with a json type `MESSAGE`.

Kudos to b1oodborne from Discord for finding this issue.


bshifter (approved):

alltilla:
@bazsi Do you agree with this change?

bazsi (approved):

PR #164, opened 2024-06-15
Filterx null coalesce operator

bshifter:
Add null coalesce operator to simplify handling of default values.  Suppresses errors occurring in the LHS argument, but logs them for debugging purposes.

```
filterx {
  key_na = "wrong";
  key = "bar";
  data = json({"foo":"1", "bar":"2", "baz":"3"});
  x = data[key_na] ?? "default";
  y = null ?? data[key];
  z = data[key_na] ?? data[key] ?? "default";
}
```

### DIFF
--- a/lib/filterx/expr-null-coalesce.c
+++ b/lib/filterx/expr-null-coalesce.c
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2024 Axoflow
+ * Copyright (c) 2024 shifter <shifter@axoflow.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "filterx/expr-null-coalesce.h"
+#include "filterx/object-null.h"
+#include "filterx/filterx-eval.h"
+#include "filterx/object-message-value.h"
+
+typedef struct _FilterXNullCoalesce FilterXNullCoalesce;
+
+struct _FilterXNullCoalesce
+{
+  FilterXBinaryOp super;
+};
+
+static FilterXObject *
+_eval(FilterXExpr *s)
+{
+  FilterXNullCoalesce *self = (FilterXNullCoalesce *) s;
+
+  FilterXObject *lhs_object = filterx_expr_eval(self->super.lhs);
+  if (!lhs_object || filterx_object_is_type(lhs_object, &FILTERX_TYPE_NAME(null))
+      || (filterx_object_is_type(lhs_object, &FILTERX_TYPE_NAME(message_value))
+          && filterx_message_value_get_type(lhs_object) == LM_VT_NULL))
+    {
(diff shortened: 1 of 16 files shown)


lib/filterx/tests/test_expr_null_coalesce.c
```diff
@@ -0,0 +1,157 @@
...
+  cr_assert(!!res);
+  cr_assert(filterx_object_is_type(res, &FILTERX_TYPE_NAME(test_unknown_object)));
+  filterx_expr_unref(coalesce);
+  filterx_object_unref(res);
+}
+
+Test(expr_null_coalesce, test_coalescing_supressing_lhs_eval_error)
+{
+  // init errorous expression which pushes eval error to context - len function without any args
+  FilterXExpr *err_expr = filterx_simple_function_new("len", filterx_function_args_new(NULL, NULL),
```
alltilla, line 72: You can use this:
`FilterXExpr *filterx_error_expr_new(void);`

lib/filterx/expr-null-coalesce.c
```diff
@@ -0,0 +1,67 @@
...
+  FilterXObject *rhs_object = NULL;
+
+  if (self->super.lhs)
+    lhs_object = filterx_expr_eval(self->super.lhs);
+  if (!lhs_object || filterx_object_is_type(lhs_object, &FILTERX_TYPE_NAME(null)))
+    {
+      if (!lhs_object)
+        {
+          msg_debug("FILTERX NULL coalesce error handling",
+                    filterx_format_last_error());
```
alltilla, line 48: I think this could be a bit more verbose, to communicate the behavior cleaner, mentioning that the lhs evaluated with an error, so the rhs will be returned.

tests/light/functional_tests/filterx/test_filterx.py
```diff
@@ -1534,3 +1534,130 @@ def test_unset_empties(config, syslog_ng):
     assert file_true.get_stats()["processed"] == 1
     assert "processed" not in file_false.get_stats()
     assert file_true.read_log() == "[{},{},[],[]]\n"
+
+
+def test_null_coalesce_happy_path(config, syslog_ng):
```
alltilla, line 1539: For the E2E tests generally:

It is the cleanest to test only one functionality in one TC, but there is a performance impact, which I think should be considered. All these TCs start syslog-ng, which has a significant overhead, we can quickly end up with minutes long E2E test runs this way.

I didn't think through if any of these TCs could be merged sanely, but this "feels" like that we should handle these cases in fever TCs for this simple feature to save test time.

tests/light/functional_tests/filterx/test_filterx.py
```diff
@@ -1534,3 +1534,130 @@ def test_unset_empties(config, syslog_ng):
...
+
+def test_null_coalesce_do_not_supress_last_error(config, syslog_ng):
+    (file_true, file_false) = create_config(
+        config, r"""
+        data = json({"foo":"1", "bar":"2", "baz":"3"});
+        def = "default";
+        key1 = "missing_key1";
+        key2 = "missing_key2";
+        key3 = "missing_key3";
+        $MSG = data[key1] ?? data[key2] ?? data[key3];
```
alltilla, line 1657: It would be nice to see a mix of `??` and `? :` in a TC. It is a corner case which can be handled in different ways, and it should match with the rest of the world, if there is a common way in other languages.

lib/filterx/tests/test_expr_null_coalesce.c
```diff
@@ -0,0 +1,157 @@
...
+#include "filterx/filterx-eval.h"
+#include "filterx/func-len.h"
+#include "filterx/expr-function.h"
+
+#include "apphook.h"
+#include "scratch-buffers.h"
+
+Test(expr_null_coalesce, test_coalescing_hard_null)
+{
+  FilterXExpr *coalesce = filterx_null_coalesce_new(NULL, filterx_non_literal_new(filterx_test_unknown_object_new()));
```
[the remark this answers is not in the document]
bshifter: NULL reprsents any filterx error in this case.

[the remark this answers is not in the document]
alltilla: This is a `NULL` expression, not a non-`NULL` expression which evaluates to `NULL`. The expression is always non-`NULL`, if I understand correctly.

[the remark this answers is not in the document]
bshifter: u r right

alltilla (approved):

alltilla:
Cool thing, good job!

PR #166, opened 2024-06-17
metrics: factor out TLS cache

alltilla:

### DIFF
--- a/lib/metrics/metrics-template.c
+++ b/lib/metrics/metrics-template.c
@@ -94,7 +94,7 @@ metrics_template_build_sck(MetricsTemplate *self,
       GString *value_buffer = scratch_buffers_alloc();
 
       label_template_format(label_template, template_options, msg, value_buffer,
-                            metrics_tls_cache_get_next_label());
+                            metrics_tls_cache_alloc_label());
     }
 
   if (self->vp)
--- a/lib/metrics/metrics-template.c
+++ b/lib/metrics/metrics-template.c
@@ -62,8 +62,9 @@ _add_dynamic_labels_vp_helper(const gchar *name, LogMessageValueType type, const
   g_string_assign(name_buffer, name);
   g_string_append_len(value_buffer, value, value_len);
 
-  StatsClusterLabel label = stats_cluster_label(name_buffer->str, value_buffer->str);
-  metrics_tls_cache_append_label(&label);
+  StatsClusterLabel *label = metrics_tls_cache_alloc_label();
+  label->name = name_buffer->str;
+  label->value = value_buffer->str;
 
   return FALSE;
 }
--- a/lib/metrics/metrics-template.c
+++ b/lib/metrics/metrics-template.c
@@ -24,79 +24,9 @@
 
 #include "metrics-template.h"
 #include "label-template.h"
+#include "metrics-tls-cache.h"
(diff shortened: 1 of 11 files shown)


MrAnno (approved):

lib/metrics/metrics-tls-cache.c
```diff
@@ -84,15 +84,15 @@ _deinit_tls_clusters_map_thread_init_hook(gpointer user_data)
 }
 
 static void
-_init_tls_clusters_map_apphook(gint type, gpointer user_data)
+_init_tls_clusters_map_apphook(void)
```
bazsi, line 87: If I understand correctly, this is not an apphook anymore, so maybe remove that from the name.

lib/metrics/metrics-tls-cache.c
```diff
@@ -144,14 +144,13 @@ metrics_tls_cache_get_labels_len(void)
 void
 metrics_tls_cache_global_init(void)
```
bazsi, line 145: maybe this is a bit too granular for apphook, and we should just have a metrics_global_init()/deinit() functions.

e.g. drop the tls_cache piece from the name. if there will be multiple metrics related global init/deinit chores, we could call all of those from a single metrics_global_init() function.

lib/metrics/metrics-tls-cache.c
```diff
@@ -0,0 +1,157 @@
...
+}
+
+void
+metrics_tls_cache_append_label(StatsClusterLabel *label)
+{
+  label_buffers = g_array_append_vals(label_buffers, label, 1);
+}
+
+StatsClusterLabel *
+metrics_tls_cache_get_next_label(void)
```
bazsi, line 126: [optional] this is more like an alloc() function. get_next_label() seems like a read only operation. I needed to check out the implementation to understand what it was doing exactly.

lib/metrics/metrics-template.c
```diff
@@ -133,7 +63,7 @@ _add_dynamic_labels_vp_helper(const gchar *name, LogMessageValueType type, const
   g_string_append_len(value_buffer, value, value_len);
 
   StatsClusterLabel label = stats_cluster_label(name_buffer->str, value_buffer->str);
-  label_buffers = g_array_append_vals(label_buffers, &label, 1);
+  metrics_tls_cache_append_label(&label);
```
bazsi, line 66: This could easily use metrics_tls_cache_get_next_label() instead of initializing a label locally and then copying it.

I don't think this is much of a performance issue as the label itself is small, but this means we have two distinct ways for accessing the label array.

lib/metrics/metrics-tls-cache.c
alltilla: Thanks, fixed.

lib/metrics/metrics-tls-cache.c
alltilla: Agree, fixed.

lib/metrics/metrics-tls-cache.c
alltilla: Good idea, fixed.

lib/metrics/metrics-template.c
alltilla: Good idea, fixed.

bshifter (approved):

PR #170, opened 2024-06-19
filterx: lower() and upper()

alltilla:

### DIFF
--- a/lib/filterx/func-str-transform.c
+++ b/lib/filterx/func-str-transform.c
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2024 Axoflow
+ * Copyright (c) 2024 Attila Szakacs
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "filterx/func-str-transform.h"
+#include "filterx/object-string.h"
+#include "filterx/filterx-eval.h"
+
+static const gchar *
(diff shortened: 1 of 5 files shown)


lib/filterx/func-str-transform.c
```diff
@@ -0,0 +1,79 @@
...
+
+FilterXObject *
+filterx_simple_function_lower(FilterXExpr *s, GPtrArray *args)
+{
+  gssize len;
+  const gchar *str = _extract_str_arg(s, args, &len);
+  if (!str)
+    return NULL;
+
+  gchar *lower = g_ascii_strdown(str, len);
```
OverOrion, line 59: I would suggest using [`utf8_strdown`](https://docs.gtk.org/glib/func.utf8_strdown.html) here, what do you think?

lib/filterx/func-str-transform.c
```diff
@@ -0,0 +1,79 @@
...
+
+FilterXObject *
+filterx_simple_function_upper(FilterXExpr *s, GPtrArray *args)
+{
+  gssize len;
+  const gchar *str = _extract_str_arg(s, args, &len);
+  if (!str)
+    return NULL;
+
+  gchar *lower = g_ascii_strup(str, len);
```
OverOrion, line 74: I would suggest using [`utf8_strup`](https://docs.gtk.org/glib/func.utf8_strup.html) here, what do you think?

lib/filterx/func-str-transform.c
bazsi, line 59: I'd try to avoid allocations here if possible.

alltilla: Nice catch, fixed.

lib/filterx/func-str-transform.c
alltilla: Nice catch, fixed.

lib/filterx/func-str-transform.c
alltilla: From `g_utf8_strup()`:
> The exact manner that this is done depends on the current locale, and may result in the number of characters in the string increasing. (For instance, the German ess-zet will be changed to SS.)

The only way I see to optimize this would be to use `ScratchBuffers` and a `GString` API that converts the string in-place (we would save the allocation of the new string with `ScratchBuffers`, but would still need to copy), but there is none for UTF-8 unfortunately.

I am open for ideas, if there's any.

bazsi: I don't think we want UTF-8 here. Nothing ensures that a string is actually utf8.

bazsi: Check if the size of the string is less than 256 chars and use a local buffer if it is.

But let's see if this shows up in our performance test.

OverOrion (approved):

lib/filterx/func-str-transform.c
alltilla: The problem is that for UTF-8 there is no in-place replace function. For ASCII we have one in `GString::`.

alltilla: We do the same in the respective template function: https://github.com/alltilla/axosyslog/blob/1f82d7d4991c1e9e40d4060e0cec7830418cc5f8/modules/basicfuncs/str-funcs.c#L358
But I am okay to using ASCII.

bazsi: Ok then

PR #172, opened 2024-06-20
filterx: fix crash if cache-json-file() fails

OverOrion:
Signed-off-by: Szilard Parrag <szilard.parrag@axoflow.com>

### DIFF
--- a/modules/json/filterx-cache-json-file.c
+++ b/modules/json/filterx-cache-json-file.c
@@ -121,11 +121,10 @@ _load_json_file(const gchar *filepath, GError **error)
     }
 
   FilterXObject *result = NULL;
-  if (object)
-    {
-      result = filterx_json_new_from_object(object);
-      filterx_object_make_readonly(result);
-    }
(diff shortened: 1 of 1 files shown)


news/bugfix-172.md
```diff
@@ -0,0 +1 @@
+filterx: fix crash in `case cache-json-file()` object creation fails
```
alltilla, line 1: We don't communicate filterx for now, so we don't need this.

modules/json/filterx-cache-json-file.c
```diff
@@ -124,7 +124,8 @@ _load_json_file(const gchar *filepath, GError **error)
   if (object)
     {
       result = filterx_json_new_from_object(object);
-      filterx_object_make_readonly(result);
+      if (result)
+        filterx_object_make_readonly(result);
```
alltilla, line 128: If we don't get a `FilterXObject` here, we should free `object`, or it will be leaked.

alltilla (approved):

PR #175, opened 2024-06-24
filterx/expr-conditional: refactor 

alltilla:
Although the original version is shorter it is a bit difficult for me to understand its flow.

For example the using or freeing of conditional_value or the "fallthrough"-like implementation of the evaluation of statements.

In some places it was hard to wrap my head around why we are returning something, and it was caused by the same code handling the basic if-elif-else code and ternary as well.

Hopefully the new version makes it easier to read and is less error-prone.


alltilla:
This is just a bandaid to a bigger problem. This code needs a bigger refactor. Closing.

PR #181, opened 2024-06-27
lexer fix for slash strings

bshifter:
lexer was unable to handle escaped slashes in slash strings ( i.e: \foo/\bar\ ) this lesser refinement of lexer rule for slash strings allows the lexer to recognize escaped slashes

### DIFF
--- a/tests/light/functional_tests/filterx/test_filterx.py
+++ b/tests/light/functional_tests/filterx/test_filterx.py
@@ -1680,3 +1680,60 @@ def test_null_coalesce_precedence_versus_ternary(config, syslog_ng):
     assert file_true.get_stats()["processed"] == 1
     assert "processed" not in file_false.get_stats()
     assert file_true.read_log() == '{"a":"default","b":"2","c":"2","d":"3","e":"1"}\n'
+
+
+def test_slash_string_features(config, syslog_ng):
+    cfg = r"""
+            $MSG = json();
+            $MSG.base = /foo bar/;
+            $MSG.line_break = /foo
+bar/;
+            $MSG.joint_lines = /foo \
+bar/;
+            ## escaped characters
+            $MSG.escaped_backslash = /foo\\bar/;
+            $MSG.escaped_slash = /foo\/bar/;
+            $MSG.non_escaped_single_quotes = /foo'bar/;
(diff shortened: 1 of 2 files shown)


alltilla:
I have tested this manually with this awk script based on the E2E tests:
```
BEGIN { RS=""; FS="\\n" }
{
  gsub(/pattern goes here/, "almafa");
  print
}
```

I have found 2 differences between awk's regexp constants and axosyslog's slash strings:
1. awk did not allow line breaks in the pattern, while axosyslog did, but I think it is a useful feature to have in axosyslog
2. awk has `\nnn` for octal, while axosyslog has `\onnn`, but this is the historical behavior of axosyslog, so we should keep it as is

tests/light/functional_tests/filterx/test_filterx.py
```diff
@@ -1680,3 +1680,60 @@ def = "default";
...
+            $MSG = json();
+            $MSG.base = /foo bar/;
+            $MSG.line_break = /foo
+bar/;
+            $MSG.joint_lines = /foo \
+bar/;
+            ## escaped characters
+            $MSG.escaped_backslash = /foo\\bar/;
+            $MSG.escaped_slash = /foo\/bar/;
+            $MSG.non_escaped_double_quotes = /foo"bar/;
```
alltilla, line 1696: It would be nice to see `'` tested, too.

tests/light/functional_tests/filterx/test_filterx.py
```diff
@@ -1680,3 +1680,60 @@ def = "default";
...
+            $MSG.form_feed = /foo\fbar/;
+            $MSG.backspace = /foo\bbar/;
+            $MSG.alert = /foo\abar/;
+
+            ##
+            $MSG.hexadecimal_value = /foo\x40bar/;
+            $MSG.octal_value = /foo\o100bar/;
+        """
+
+    # assert cfg == ""
```
alltilla, line 1712: Is this needed?

alltilla (commented):
2 minor comments, otherwise approve.

alltilla:
Thanks for cleaning up my mess :)

alltilla (approved):

PR #192, opened 2024-07-05
otel,loki,bigquery: add `headers()` option to dest 

alltilla:
Example config:
```
opentelemetry(
  ...
  headers(
    "organization" => "Axoflow"
    "stream-name" => "axo-stream"
  )
);
```


bazsi:
I tested the headers() option for Otel as I had a good enough test environment locally and it seems to work fine. It did lowercase the header as expected and the tcpdump capture did contain the header added.

bazsi (approved):

MrAnno (approved):

PR #198, opened 2024-07-08
Filterx parse csv delimiter fix

### LINKED ISSUE #197: `parse_csv()` allow to set delimiters to empty string if string_delimiters is not-empty

bshifter:
Since the string-delimiters feature was added, the delimiter argument must be allowed to be set to empty to avoid parsing side effects.
+function usage description update

closes #197


jszigetvari (approved):
Looks okay to me.

MrAnno (approved):

PR #203, opened 2024-07-12
Filterx regexp subst

bshifter:
Complete implementation of regex_subst filterx function based on #195.
usage:
```
regexp_subst(string, pattern, replacement)
```

example:
```
filterx {
  $MSG = regexp_subst($MSG, "u*", "X");
}
```

### DIFF
--- a/lib/filterx/expr-regexp.c
+++ b/lib/filterx/expr-regexp.c
@@ -29,6 +29,8 @@
 #include "filterx/object-dict-interface.h"
 #include "compat/pcre.h"
 
+#define FILTERX_FUNC_REGEXP_SUBST_USAGE "regexp_subst(string, pattern, replacement)"
+
 typedef struct FilterXReMatchState_
 {
   pcre2_match_data *match_data;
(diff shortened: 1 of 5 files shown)


lib/filterx/expr-regexp.c
```diff
@@ -417,3 +438,290 @@ filterx_expr_regexp_search_generator_new(FilterXExpr *lhs, const gchar *pattern)
...
+      g_set_error(error, FILTERX_FUNCTION_ERROR, FILTERX_FUNCTION_ERROR_CTOR_FAIL,
+                  "argument must be a string literal: replacement. " FILTERX_FUNC_REGEXP_SUBST_USAGE);
+      return NULL;
+    }
+
+  return g_strdup(replacement);
+}
+
+static gboolean
+_extract_optional_flags(FilterXFuncRegexpSubst *self, FilterXFunctionArgs *args, GError **error)
```
alltilla, line 581: I like the idea of separating the matching options to a separate entity, we should lean into it more IMO.
`FilterXRegexpFlags` could be a separate class, and this function should operate on it only. This way we will be able to reuse it for other functions as well, e.g. for `regexp_search()` when it becomes a real function.

lib/filterx/expr-regexp.c
```diff
@@ -28,6 +28,9 @@
 #include "filterx/object-list-interface.h"
 #include "filterx/object-dict-interface.h"
 #include "compat/pcre.h"
+#include "scratch-buffers.h"
+
+#define FILTERX_FUNC_REGEXP_SUBST_USAGE "regexp_subst(string, pattern, replacement)"
```
alltilla, line 33: We should list the optional arguments, too.

lib/filterx/expr-regexp.c
```diff
@@ -28,6 +28,14 @@
 #include "filterx/object-list-interface.h"
 #include "filterx/object-dict-interface.h"
 #include "compat/pcre.h"
+#include "scratch-buffers.h"
+
+#define FILTERX_FUNC_REGEXP_SUBST_USAGE "regexp_subst(string, pattern, replacement, [" \
```
alltilla, line 33: To me `[]` implies that these are passed in an array, I think we should remove the brackets, the rest is good.

lib/filterx/expr-regexp.h
```diff
@@ -26,8 +26,27 @@
...
+#define FILTERX_FUNC_REGEXP_SUBST_FLAG_UTF8_NAME "utf8"
+#define FILTERX_FUNC_REGEXP_SUBST_FLAG_IGNORECASE_NAME "ignorecase"
+#define FILTERX_FUNC_REGEXP_SUBST_FLAG_NEWLINE_NAME "newline"
+
+typedef struct FilterXFuncRegexpSubstOpts_
+{
+  gboolean global;
+  gboolean jit;
+  gboolean utf8;
+  gboolean ignorcase;
```
alltilla, line 42: typo:
```diff
-  gboolean ignorcase
+  gboolean ignorecase
```

alltilla (commented):
nice feature! :)

2 small findings, after these we can merge.

alltilla (approved):

PR #208, opened 2024-07-15
[dbld] Handle already exist USER_ID in docker environment

mitzkia:
Original issue:
 - While going through the local dbld dev environment creation with:
   - `dbld/rules image`
   - `dbld/rules bootstrap`
 - I've got the following error:
```
...
sudo: unknown user micek
sudo: error initializing audit plugin sudoers_audit
make: *** [dbld/rules:101: bootstrap-ubuntu-noble] Error 1
```

Root cause:
 - After ubuntu:kinetic general "ubuntu" user was added to the base image with USER_ID: 1000
```
micek@DESKTOP-0E9OKG3:~$ docker run -it ubuntu:jammy bash -c "grep 1000 /etc/passwd"
micek@DESKTOP-0E9OKG3:~$ docker run -it ubuntu:kinetic bash -c "grep 1000 /etc/passwd"
micek@DESKTOP-0E9OKG3:~$ docker run -it ubuntu:lunar bash -c "grep 1000 /etc/passwd"
ubuntu:x:1000:1000:Ubuntu:/home/ubuntu:/bin/bash
micek@DESKTOP-0E9OKG3:~$ docker run -it ubuntu:mantic bash -c "grep 1000 /etc/passwd"
ubuntu:x:1000:1000:Ubuntu:/home/ubuntu:/bin/bash
micek@DESKTOP-0E9OKG3:~$ docker run -it ubuntu:noble bash -c "grep 1000 /etc/passwd"
ubuntu:x:1000:1000:Ubuntu:/home/ubuntu:/bin/bash
```
 - The else branch was not handled while checking the USER_ID exitance in `dbld/images/entrypoint.sh`  before the user creation and based on this own user can not be created.
 - This line causes the error: `exec sudo --preserve-env -Hu "${USER_NAME}" "$@"`

Solution:
 - With removing the unused 'ubuntu' user we can add our own user with id 1000

### DIFF
--- a/dbld/images/entrypoint.sh
+++ b/dbld/images/entrypoint.sh
@@ -25,10 +25,14 @@ function create_user() {
 if [[ "$USER_ID" -eq 0 ]]; then
     "$@"
 else
-    if ! getent passwd $USER_ID > /dev/null
+    if getent passwd $USER_ID > /dev/null
     then
+        echo "USER_ID: $USER_ID already exist in passwd database, performing cleanup"
+        userdel --remove $(getent passwd $USER_ID | cut -d":" -f1)
+	create_user
+    else
         create_user
     fi
-
+    echo "Added new user: $USER_NAME"
     exec sudo --preserve-env -Hu "${USER_NAME}" "$@"
 fi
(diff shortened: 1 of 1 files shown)


bazsi:
why wasn't this an issue with our current setup? the CI does not use USERID 1000?

bazsi:
ah. seems I am using the devshell which is debian testing based.

bazsi:
I could reproduce this now. Indeed any interactive work in our ubuntu based images would not work because of the missing user. For some reason, building the packages still works, but I think it makes sense to merge this.

bazsi (approved):
Thanks for the PR.

bazsi:
NOTE: it works to create packages, as that one runs commands as root, so no need for the creation of the user.

PR #210, opened 2024-07-16
README: fix forked off from version

alltilla:


MrAnno (approved):
I've already fixed one, I missed the second one :(

PR #213, opened 2024-07-17
appmodel: filterx-related fixes

MrAnno:


MrAnno:
@bazsi @alltilla Is it a problem if I set the "legacy" `.app.name` kv and the `.app.*` tag unconditionally even in modern filterx use cases?

If so, I think we should add a syntax that allows marking application adapters as `modern`, because relying on the top-level type is not practical.

bazsi (approved):

PR #214, opened 2024-07-19
filterx: support message_value in lower() and upper()

alltilla:


alltilla:
Closing in favor of #229

PR #216, opened 2024-07-19
Null-pointer fixes

MrAnno:
Patches are originally coming from syslog-ng/syslog-ng#4132 (@OverOrion do you want to close that PR?)

Fixes syslog-ng/syslog-ng#5025
Fixes syslog-ng/syslog-ng#5026

(Contains an alternative implementation of syslog-ng/syslog-ng#5032 and syslog-ng/syslog-ng#5031)


MrAnno:
I've added @nivelus to the news entry as a thank-you for finding some of these issues.

bazsi (approved):

PR #217, opened 2024-07-19
FilterXObject class method add implementation for base types and list/dict variables #1

bshifter:
This pull request introduces a significant refactor as part of merging the Filterx lists and dictionaries. Specifically, it addresses the implementation of the + operator using the 'add' class method to ensure it behaves as expected.

Current Changes:

* redesign `expr-plus` operator
* extend every base class with `add` method

Future Work:

This is a partial solution; generators are not yet handled.
Part 2 will focus on integrating generator handling and refining generator grammar definitions.

example:
```
filterx {
    $MSG = json();
    $MSG.string = "foo" + "bar" + "baz";
    $MSG.bytes = string(bytes("\xCA") + bytes("\xFE"));
    $MSG.datetime_datetime = string(strptime("2000-01-01T00:00:00Z", "%Y-%m-%dT%H:%M:%S%z") + strptime("1970-01-01T02:00:00Z", "%Y-%m-%dT%H:%M:%S%z"));
    $MSG.datetime_integer = string(strptime("2000-01-01T00:00:00Z", "%Y-%m-%dT%H:%M:%S%z") + 3600000000);
    $MSG.datetime_double = string(strptime("2000-01-01T00:00:00Z", "%Y-%m-%dT%H:%M:%S%z") + 3600.0);
    $MSG.integer_integer = 3 + 4 + 5;
    $MSG.integer_double = 3 + 0.5;
    $MSG.double_integer = 3.5 + 2;
    $MSG.double_double = 3.14 + 0.86;
    js1 = json_array(["foo","bar"]);
    js2 = json_array(["baz","other"]);
    $MSG.list_list = js1 + js2;
    dict1 = json({"foo":"bar"});
    dict2 = json({"baz":"other"});
    $MSG.dict_dict = dict1 + dict2;
}
```

### DIFF
--- a/lib/filterx/expr-plus.c
+++ b/lib/filterx/expr-plus.c
@@ -35,37 +35,21 @@ _eval(FilterXExpr *s)
 {
   FilterXOperatorPlus *self = (FilterXOperatorPlus *) s;
 
-  FilterXObject *lhs_object = filterx_expr_eval(self->super.lhs);
+  FilterXObject *lhs_object = filterx_expr_eval_typed(self->super.lhs);
   if (!lhs_object)
-    {
-      return NULL;
-    }
+    return NULL;
 
-  FilterXObject *rhs_object = filterx_expr_eval(self->super.rhs);
+  FilterXObject *rhs_object = filterx_expr_eval_typed(self->super.rhs);
   if (!rhs_object)
     {
       filterx_object_unref(lhs_object);
       return NULL;
     }
 
-  if (filterx_object_is_type(lhs_object, &FILTERX_TYPE_NAME(string)) &&
-      filterx_object_is_type(rhs_object, &FILTERX_TYPE_NAME(string)))
-    {
-      gsize lhs_len, rhs_len;
-      const gchar *lhs_value = filterx_string_get_value(lhs_object, &lhs_len);
-      const gchar *rhs_value = filterx_string_get_value(rhs_object, &rhs_len);
-      GString *buffer = scratch_buffers_alloc();
-
-      g_string_append_len(buffer, lhs_value, lhs_len);
-      g_string_append_len(buffer, rhs_value, rhs_len);
-      /* FIXME: support taking over the already allocated space */
-      return filterx_string_new(buffer->str, buffer->len);
-    }
-
-  filterx_eval_push_error("operator+ only works on strings", s, NULL);
+  FilterXObject *res = filterx_object_add_object(lhs_object, rhs_object);
   filterx_object_unref(lhs_object);
   filterx_object_unref(rhs_object);
-  return NULL;
+  return res;
 }
 
 FilterXExpr *
--- a/lib/filterx/object-datetime.c
+++ b/lib/filterx/object-datetime.c
@@ -205,6 +205,28 @@ _repr(FilterXObject *s, GString *repr)
   return datetime_repr(&ut, repr);
 }
 
+static FilterXObject *
+_add(FilterXObject *self, FilterXObject *object)
+{
+  UnixTime result;
+  UnixTime base = filterx_datetime_get_value(self);
+  if (filterx_object_is_type(object, &FILTERX_TYPE_NAME(integer)))
+    {
+      GenericNumber gn = filterx_primitive_get_value(object);
+      result = unix_time_add_duration(base, gn_as_int64(&gn));
+    }
+  else if (filterx_object_is_type(object, &FILTERX_TYPE_NAME(double)))
+    {
+      GenericNumber gn = filterx_primitive_get_value(object);
+      result = unix_time_add_duration(base, (gint64)(gn_as_double(&gn) * USEC_PER_SEC));
+    }
+  else
+    return NULL;
+
+
+  return filterx_datetime_new(&result);
+}
+
 const gchar *
 _strptime_get_time_str_from_object(FilterXObject *obj, gsize *len)
 {
@@ -381,4 +403,5 @@ FILTERX_DEFINE_TYPE(datetime, FILTERX_TYPE_NAME(object),
                     .map_to_json = _map_to_json,
                     .marshal = _marshal,
                     .repr = _repr,
+                    .add = _add,
                    );
--- a/lib/filterx/object-dict-interface.c
+++ b/lib/filterx/object-dict-interface.c
@@ -207,6 +207,24 @@ filterx_dict_init_instance(FilterXDict *self, FilterXType *type)
   self->support_attr = TRUE;
 }
 
+static FilterXObject *
+_add(FilterXObject *lhs_object, FilterXObject *rhs_object)
+{
+  if (!filterx_object_is_type(lhs_object, &FILTERX_TYPE_NAME(dict)) ||
+      !filterx_object_is_type(rhs_object, &FILTERX_TYPE_NAME(dict)))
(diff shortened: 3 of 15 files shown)


lib/filterx/object-datetime.c
```diff
@@ -205,6 +205,38 @@ _repr(FilterXObject *s, GString *repr)
...
+    {
+      GenericNumber gn = filterx_primitive_get_value(rhs_object);
+      result = unix_time_add_unix_epoch(base, (gint64)(gn_as_double(&gn) * USEC_PER_SEC));
+    }
+  else
+    goto error;
+
+
+  return filterx_datetime_new(&result);
+error:
```
alltilla, line 236: [optional]

Using goto for a simple `return NULL` seems unnecessary, but I understand the general pattern so it is completely okay to keep it like this.

lib/filterx/object-datetime.c
```diff
@@ -205,6 +205,38 @@ _repr(FilterXObject *s, GString *repr)
   return datetime_repr(&ut, repr);
 }
 
+static FilterXObject *
+_add(FilterXObject *lhs_object, FilterXObject *rhs_object)
+{
+  if (!filterx_object_is_type(lhs_object, &FILTERX_TYPE_NAME(datetime)))
+    goto error;
```
alltilla, line 212: If this is not `datetime`, it is a programming error, so this should be an assert or we could just simply remove it. This applies to the rest of the types' `add()`, too.

lib/filterx/object-datetime.c
```diff
@@ -205,6 +205,38 @@ _repr(FilterXObject *s, GString *repr)
   return datetime_repr(&ut, repr);
 }
 
+static FilterXObject *
+_add(FilterXObject *lhs_object, FilterXObject *rhs_object)
```
alltilla, line 209: This is a virtual method for adding an other object to `self`, I think the signature should represent that.
```c
_add(FilterXObject *self, FilterXObject *object)
```
Changing the signature would make it more natural to see that `lhs_object` / `self` is always `datetime` typed. This applies to the rest of the types' add(), too.

lib/filterx/object-dict-interface.c
```diff
@@ -207,6 +207,24 @@ filterx_dict_init_instance(FilterXDict *self, FilterXType *type)
...
+static FilterXObject *
+_add(FilterXObject *lhs_object, FilterXObject *rhs_object)
+{
+  if (!filterx_object_is_type(lhs_object, &FILTERX_TYPE_NAME(dict)) ||
+      !filterx_object_is_type(rhs_object, &FILTERX_TYPE_NAME(dict)))
+    return NULL;
+
+  FilterXObject *cloned = filterx_object_clone(lhs_object);
+
+  if(!filterx_dict_merge(cloned, rhs_object))
```
alltilla, line 219: [nitpick]

A space is missing after the `if`.

lib/filterx/object-primitive.c
```diff
@@ -63,6 +63,26 @@ _integer_map_to_json(FilterXObject *s, struct json_object **object, FilterXObjec
...
+{
+  if (!filterx_object_is_type(lhs_object, &FILTERX_TYPE_NAME(integer)))
+    goto error;
+  GenericNumber base = filterx_primitive_get_value(lhs_object);
+  if (filterx_object_is_type(rhs_object, &FILTERX_TYPE_NAME(integer)))
+    {
+      GenericNumber add = filterx_primitive_get_value(rhs_object);
+      return filterx_integer_new(gn_as_int64(&base) + gn_as_int64(&add));
+    }
+  else if (filterx_object_is_type(rhs_object, &FILTERX_TYPE_NAME(double)))
```
alltilla, line 77: [nitpick]

This `else if` could be a simple `if`, as you did an early return above.

lib/filterx/object-primitive.c
```diff
@@ -63,6 +63,26 @@ _integer_map_to_json(FilterXObject *s, struct json_object **object, FilterXObjec
...
+    {
+      GenericNumber add = filterx_primitive_get_value(rhs_object);
+      return filterx_double_new(gn_as_double(&base) + gn_as_double(&add)); // implicit typecast, is that ok?
```
alltilla, line 80: double + int and int + double is currently different:
```
return filterx_double_new(gn_as_double(&base) + gn_as_int64(&add));
return filterx_double_new(gn_as_double(&base) + gn_as_double(&add));
```
I guess the result is the same but it feels strange. Also from a clean code perspective it would be better to not duplicate these codes. Maybe we can take advantage of the commutativity of adding, and write one function that handles both.

lib/filterx/expr-plus.c
```diff
@@ -37,9 +37,7 @@ _eval(FilterXExpr *s)
 
   FilterXObject *lhs_object = filterx_expr_eval(self->super.lhs);
```
alltilla, line 38: We should either make these `filterx_expr_eval_typed()` or support `message_value` in all the `add()` implementations.

lib/filterx/object-primitive.c
```diff
@@ -63,6 +63,26 @@ _integer_map_to_json(FilterXObject *s, struct json_object **object, FilterXObjec
   return TRUE;
 }
 
+static FilterXObject *
+_integer_add(FilterXObject *lhs_object, FilterXObject *rhs_object)
```
alltilla, line 67: We should handle bool as well, I think.

lib/timeutils/unixtime.c
```diff
@@ -368,3 +368,22 @@ unix_time_to_unix_epoch(const UnixTime ut)
...
 }
+
+UnixTime
+unix_time_add_unix_time(UnixTime time, UnixTime add)
+{
+  UnixTime ut =
+  {
+    .ut_sec = time.ut_sec + add.ut_sec,
+    .ut_usec = time.ut_usec + add.ut_usec,
+    .ut_gmtoff = time.ut_gmtoff, //TODO: figure this out
```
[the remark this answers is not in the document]
alltilla: Or maybe we should also fail in case both datetime's have timezones. If we do that, essentially this becomes a datetime + timedelta (or timedelta + datetime) operation, as the only difference between a datetime and timedelta is that timedelta cannot have a timezone.

alltilla:
Please change the title of the PR to represent what is happening. This is in a way both less and more than list and dict merge. (we dont have generator support, but have other type support, like int, str, ...)

lib/timeutils/unixtime.c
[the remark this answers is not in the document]
bshifter: this is the case when we use rhs datetime as timedelta, simply skipping its timezone.

lib/filterx/object-primitive.c
alltilla: We decided not to handle them, just like how go and rust works.

lib/timeutils/unixtime.c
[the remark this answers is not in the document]
alltilla: We decided not to support adding datetimes. We will eventually have a duration type that could be added. Also we will have a `combine_date_and_time()` function or sth similar for combining two datetimes.

alltilla (commented):
Approve for the code.

The light test fails. After fixing it we can merge.

alltilla (approved):

PR #219, opened 2024-07-21
Filterx parse csv refinements

bazsi:
Another round of csv parsing refinements:

* strip_whitespaces argument renamed to strip_whitespace 
* never fold the delimiter argument into being a whitespace (e.g. when parsing tab separated values)
* strip_whitespace should not default to TRUE


bazsi:
this still breaks the tests and it is indeed a bug (e.g. whitespace within quotation marks that should be stripped). don't merge yet.

bshifter (approved):

bazsi:
@bshifter can you please re-approve?

bshifter (approved):

PR #220, opened 2024-07-22
filterx: `update_metric()`

alltilla:
Example:
```
update_metric("filterx_metric", labels={"msg": $MSG, "foo": "foovalue"}, level=1, increment=$INCREMENT)
```

### DIFF
--- a/lib/filterx/filterx-metrics-labels.c
+++ b/lib/filterx/filterx-metrics-labels.c
@@ -1,0 +1,361 @@
+/*
+ * Copyright (c) 2024 Axoflow
+ * Copyright (c) 2024 Attila Szakacs <attila.szakacs@axoflow.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "filterx-metrics-labels.h"
+#include "filterx-eval.h"
+#include "expr-literal.h"
+#include "expr-literal-generator.h"
+#include "object-string.h"
+#include "object-dict-interface.h"
+#include "metrics/metrics-tls-cache.h"
+#include "stats/stats-cluster-single.h"
+#include "scratch-buffers.h"
+
+#include <string.h>
+
+typedef struct _FilterXMetricsLabel
+{
+  gchar *name;
+  struct
+  {
+    FilterXExpr *expr;
+    gchar *str;
+  } value;
+} FilterXMetricsLabel;
+
+static gchar *
+_format_str_obj(FilterXObject *obj)
+{
+  GString *str = scratch_buffers_alloc();
+  return filterx_object_repr(obj, str) ? str->str : NULL;
+}
+
+static gchar *
+_format_value_expr(FilterXExpr *expr)
+{
+  FilterXObject *obj = filterx_expr_eval_typed(expr);
+  if (!obj)
+    return NULL;
+
+  gchar *result = _format_str_obj(obj);
+  if (!result)
+    filterx_eval_push_error("failed to format metrics label value, repr() failed", expr, obj);
+
+  filterx_object_unref(obj);
+  return result;
+}
+
+static gboolean
+_label_format(FilterXMetricsLabel *self, StatsClusterLabel *sc_label)
+{
+  sc_label->name = self->name;
+
+  if (self->value.str)
+    {
+      sc_label->value = self->value.str;
+      return TRUE;
+    }
+
+  sc_label->value = _format_value_expr(self->value.expr);
+
+  return !!sc_label->value;
+}
+
+static gboolean
+_label_is_const(FilterXMetricsLabel *self)
+{
+  return !!self->value.str;
+}
+
+static gint
+_label_cmp(gconstpointer a, gconstpointer b)
+{
+  const FilterXMetricsLabel *lhs = *(const FilterXMetricsLabel **) a;
+  const FilterXMetricsLabel *rhs = *(const FilterXMetricsLabel **) b;
+
+  return strcmp(lhs->name, rhs->name);
+}
+
+static void
+_label_free(FilterXMetricsLabel *self)
+{
+  g_free(self->name);
+  filterx_expr_unref(self->value.expr);
+  g_free(self->value.str);
+
+  g_free(self);
+}
+
+static gchar *
+_init_label_name(FilterXExpr *name)
+{
+  if (!filterx_expr_is_literal(name))
+    {
+      filterx_eval_push_error("failed to initialize metrics label name, name must be a string literal", name, NULL);
+      return NULL;
+    }
+
+  FilterXObject *obj = filterx_expr_eval_typed(name);
+  gchar *str = g_strdup(filterx_string_get_value(obj, NULL));
+  if (!str)
+    filterx_eval_push_error_info("failed to initialize metrics label name, name must be a string literal", name,
+                                 g_strdup_printf("got %s instead", obj->type->name), TRUE);
+
+  filterx_object_unref(obj);
+  return str;
+}
+
+static gboolean
+_init_label_value(FilterXMetricsLabel *self, FilterXExpr *value)
+{
+  if (!filterx_expr_is_literal(value))
+    {
+      self->value.expr = filterx_expr_ref(value);
+      return TRUE;
+    }
+
+  ScratchBuffersMarker marker;
+  scratch_buffers_mark(&marker);
+  self->value.str = g_strdup(_format_value_expr(value));
+  scratch_buffers_reclaim_marked(marker);
+
+  return !!self->value.str;
+}
+
+static FilterXMetricsLabel *
+_label_new(FilterXExpr *name, FilterXExpr *value)
+{
+  FilterXMetricsLabel *self = g_new0(FilterXMetricsLabel, 1);
+
+  self->name = _init_label_name(name);
+  if (!self->name)
+    goto error;
+
+  if (!_init_label_value(self, value))
+    goto error;
+
+  return self;
+
+error:
+  _label_free(self);
+  return NULL;
+}
+
+
+struct _FilterXMetricsLabels
+{
+  FilterXExpr *expr;
+  GPtrArray *literal_labels;
+  gboolean is_const;
+};
+
+static void
+_check_label_is_const(gpointer data, gpointer user_data)
+{
+  FilterXMetricsLabel *label = (FilterXMetricsLabel *) data;
+  gboolean *is_const = (gboolean *) user_data;
+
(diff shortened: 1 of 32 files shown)


alltilla:
Ready for review.

tests/light/functional_tests/filterx/test_filterx_update_metric.py
```diff
@@ -0,0 +1,189 @@
...
+        stats_level=2,
+    )
+
+    syslog_ng.reload(config)
+    network_source.write_log("foo\n")
+    file_destination.read_log()
+
+    samples = config.get_prometheus_samples([MetricFilter("syslogng_metric", {})])
+    assert len(samples) == 1
+    assert int(samples[0].value) == 1
```
OverOrion, line 189: I think a `syslog_ng.stop()` might be missing here

tests/light/functional_tests/filterx/test_filterx_update_metric.py
```diff
@@ -0,0 +1,189 @@
...
+    syslog_ng.stop()
+
+
+def test_filterx_update_metric_increment(config, port_allocator, syslog_ng):
+    network_source, file_destination = create_config(
+        config,
+        port_allocator,
+        r"""
+            update_metric("const", increment=3);
+            # TODO: fix int cast from message_value
```
OverOrion, line 117: Note for the future, or wanted to fix this in this PR?

lib/filterx/filterx-metrics-labels.c
```diff
@@ -0,0 +1,348 @@
...
+}
+
+static gchar *
+_format_value_expr(FilterXExpr *expr)
+{
+  FilterXObject *obj = filterx_expr_eval_typed(expr);
+  if (!obj)
+    return NULL;
+
+  gchar *result = _format_str_obj(obj);
```
MrAnno, line 61: Is this okay that we accept everything here?
For example, a deeply nested dict may not be something we want as a label value.
It's probably not a real issue as almost everything is accepted as a label value by Prometheus.

tests/light/functional_tests/filterx/test_filterx_update_metric.py
alltilla: Will be fixed in https://github.com/axoflow/axosyslog/pull/229

lib/filterx/filterx-metrics-labels.c
```diff
@@ -0,0 +1,348 @@
...
+
+  if (self->expr)
+    {
+      self->is_const = FALSE;
+    }
+  else
+    {
+      self->is_const = TRUE;
+      g_ptr_array_foreach(self->literal_labels, _check_label_is_const, &self->is_const);
+    }
```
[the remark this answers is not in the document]
alltilla: I have refactored this part completely, what do you think about this version?

lib/filterx/filterx-metrics-labels.c
```diff
@@ -0,0 +1,348 @@
...
+  if (!name_str)
+    {
+      filterx_eval_push_error_info("failed to format label name, name must be string", NULL,
+                                   g_strdup_printf("got %s instead", key->type->name), TRUE);
+      return FALSE;
+    }
+
+  const gchar *value_str = _format_str_obj(value);
+  if (!value_str)
+    return FALSE;
```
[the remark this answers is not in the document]
alltilla: Done thanks!

tests/light/functional_tests/filterx/test_filterx_update_metric.py
alltilla: Done, thanks :)

lib/filterx/filterx-metrics-labels.c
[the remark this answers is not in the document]
OverOrion: Perfect!

OverOrion (dismissed):

alltilla:
rebased to main

lib/filterx/filterx-metrics-labels.c
```diff
@@ -0,0 +1,361 @@
...
+  if (!(*success))
+    return;
+
+  *success = _label_format(label, metrics_cache_alloc_label(metrics_tls_cache()));
+}
+
+gboolean
+filterx_metrics_labels_format(FilterXMetricsLabels *self, StatsClusterLabel **labels, gsize *len)
+{
+  metrics_cache_reset_labels(metrics_tls_cache());
```
bazsi, line 260: is there a point in saving the metrics_tls_cache() value and not do it 3x?

lib/filterx/filterx-metrics-labels.c
```diff
@@ -0,0 +1,361 @@
...
+
+  const gchar *value_str = _format_str_obj(value);
+  if (!value_str)
+    {
+      filterx_eval_push_error_info("failed to format label value, object does not support string casting", NULL,
+                                   "", FALSE);
+      return FALSE;
+    }
+
+  StatsClusterLabel *label = metrics_cache_alloc_label(metrics_tls_cache());
```
bazsi, line 211: actually it might even make sense to cache the metrics_tls_cache() across the entire iteration. Getting the value of a TLS variable is not free, especially when wrapped into an external function, like in this case.

lib/filterx/filterx-metrics-labels.c
```diff
@@ -0,0 +1,361 @@
...
+#include <string.h>
+
+typedef struct _FilterXMetricsLabel
+{
+  gchar *name;
+  struct
+  {
+    FilterXExpr *expr;
+    gchar *str;
+  } value;
```
bazsi, line 44: isn't this a union? my understanding that only one of these is set at a time.

lib/filterx/filterx-metrics.c
```diff
@@ -0,0 +1,241 @@
...
+  FilterXObject *key_obj = filterx_expr_eval_typed(key);
+  if (!filterx_object_is_type(key_obj, &FILTERX_TYPE_NAME(string)))
+    {
+      filterx_eval_push_error_info("failed to init metrics key, key must be a string", key,
+                                   g_strdup_printf("got %s instread", key_obj->type->name), TRUE);
+      filterx_object_unref(key_obj);
+      return FALSE;
+    }
+
+  self->key.str = g_strdup(filterx_string_get_value(key_obj, NULL));
```
bazsi, line 219: probably filterx_object_extract_string() would make sense here (depends on PR #229)

lib/filterx/filterx-metrics.c
```diff
@@ -0,0 +1,241 @@
...
+#include "metrics/metrics-tls-cache.h"
+#include "scratch-buffers.h"
+
+struct _FilterXMetrics
+{
+  struct
+  {
+    FilterXExpr *expr;
+    gchar *str;
+  } key;
```
bazsi, line 41: again, probably a union.

bazsi: but if this was a union, we would need to indicate which one is used. forget it.

lib/filterx/filterx-metrics.c
```diff
@@ -0,0 +1,241 @@
...
+  gsize len;
+  const gchar *name = filterx_string_get_value(key_obj, &len);
+  if (len == 0)
+    {
+      filterx_eval_push_error("failed to format metrics key: key must be a non-empty string", self->key.expr, key_obj);
+      goto exit;
+    }
+
+  GString *name_buffer = scratch_buffers_alloc();
+  g_string_append(name_buffer, name);
```
bazsi, line 75: I guess you were trying to avoid a g_strdup() here. in that case, it makes sense to use g_string_append_len(), otherwise you'd be doing an strlen() as a part of the copy, while the length is already available.

modules/metrics-probe/filterx/func-update-metric.c
```diff
@@ -22,3 +22,196 @@
...
+static gboolean
+_get_increment(FilterXFunctionUpdateMetric *self, gint64 *increment)
+{
+  if (!self->increment.expr)
+    {
+      *increment = self->increment.value;
+      return TRUE;
+    }
+
+  FilterXObject *increment_obj = filterx_expr_eval_typed(self->increment.expr);
```
bazsi, line 55: extract_integer.

modules/metrics-probe/filterx/func-update-metric.c
```diff
@@ -22,3 +22,196 @@
...
+  if (!filterx_metrics_get_stats_counter(self->metrics, &counter))
+    goto exit;
+
+  stats_counter_add(counter, increment);
+  success = TRUE;
+
+exit:
+  if (!success)
+    {
+      msg_error("FilterX: Failed to process update_metric()", filterx_format_last_error());
```
bazsi, line 88: I think it is a bad idea to start generating a message in these cases. I think it'd be much better to add an "error" counter along with all update_metric() counters and increment that every time something fails.

With that we could even add limits on how many counters we would be adding in case there's a cardinality explosion.

it does not need to be done now, just an idea.

lib/filterx/filterx-metrics.c
```diff
@@ -110,6 +145,17 @@ filterx_metrics_get_stats_counter(FilterXMetrics *self, StatsCounterItem **count
       return TRUE;
     }
 
+  /*
+   * We need to delay the optimization to the first get() call,
+   * as we don't have stats options when FilterXExprs are being created.
+   */
+  if (!self->is_optimized)
+    {
+      if (!_optimize(self))
```
bazsi, line 154: I think this needs to be protected by a lock, quite probably we can reuse the starts_lock() in this case.

lib/filterx/filterx-metrics-labels.c
bazsi: scratch that, the only thing that indicates if one is set if str is NULL or not.

lib/filterx/filterx-metrics-labels.c
bazsi: Just to clarify, I meant that it would make sense to pass the value around to callbacks that iterate possibly multiple times per message to avoid the TLS overhead.

lib/filterx/filterx-metrics.c
bazsi: maybe the plan is to merge this first, in that case of course #229 should address this.

modules/metrics-probe/filterx/func-update-metric.c
bazsi: would depend on #229, so depends on merge order.

bazsi (dismissed):

modules/metrics-probe/filterx/func-update-metric.c
alltilla: I agree, changed the log level to debug in a new commit. We can introduce a counter later.

lib/filterx/filterx-metrics.c
alltilla: Thanks, fixed and squashed to the original commit.

lib/filterx/filterx-metrics-labels.c
alltilla: I did not think about the overhead. Thanks, fixed this in a new commit.

lib/filterx/filterx-metrics-labels.c
alltilla: Thanks, fixed in a new commit.

lib/filterx/filterx-metrics.c
alltilla: Nice catch! Fixed in a new commit.

alltilla:
Thanks for the thorough review @bazsi.

I have fixed all the comments, most of them went to new commits to make it easier to review, starting with the commit [filterx/metrics-labels: reduce the number of tls cache acquirals](https://github.com/axoflow/axosyslog/pull/220/commits/4a86b091d93ae09060f28df5ac05d2badd508f12)

The order of this PR and https://github.com/axoflow/axosyslog/pull/229 is undecided, I will update the one which gets merged later.

lib/filterx/filterx-metrics.c
alltilla: In this specific scenario, we don't need to call the extract function, as we check before that it is a literal, and there are no literal message values.

modules/metrics-probe/filterx/func-update-metric.c
alltilla: Done.

bazsi (approved):

PR #225, opened 2024-07-25
Filterx expr plus operator for generators and += operator (part 2)

bshifter:
continue #217 with expr-plus-generator operator for handling mixed generator/variable operations

This commit continues the work from #217 by introducing the expr-plus-generator 
operator. This new operator is designed to handle plus operations over generators 
and mixed generator/variable operations, ensuring that both types of elements are 
processed seamlessly as a single generator.

### DIFF
--- a/lib/filterx/expr-plus-generator.c
+++ b/lib/filterx/expr-plus-generator.c
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2024 Axoflow
+ * Copyright (c) 2024 shifter
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+#include "expr-plus.h"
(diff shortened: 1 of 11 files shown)


lib/filterx/expr-plus-generator.c
```diff
@@ -0,0 +1,117 @@
...
+#include "filterx-eval.h"
+#include "scratch-buffers.h"
+#include "expr-generator.h"
+#include "object-list-interface.h"
+#include "object-dict-interface.h"
+
+typedef struct FilterXOperatorPlusGenerator
+{
+  FilterXExprGenerator super;
+  FilterXBinaryOp elts;
```
alltilla, line 35: This seems like a double inheritance, which is not fully supported.

Both `FilterXExprGenerator` and `FilterXBinaryOp` are `FilterXExpr`s, but `FilterXBinaryOp`'s eval is never used. I think it would be cleaner to just store the 2 `FilterXExpr`s and not use `FilterXBinaryOp` at all.

lib/filterx/expr-plus-generator.c
```diff
@@ -0,0 +1,117 @@
...
+_generate_obj(FilterXObject *obj, FilterXObject *fillable)
+{
+  if (filterx_object_is_type(fillable, &FILTERX_TYPE_NAME(list)))
+    return filterx_list_merge(fillable, obj);
+
+  if (filterx_object_is_type(fillable, &FILTERX_TYPE_NAME(dict)))
+    return filterx_dict_merge(fillable, obj);
+
+  msg_error("FilterX: Failed to merge objects, invalid fillable type",
+            evt_tag_str("type", fillable->type->name));
```
alltilla, line 48: `filterx_eval`'s error reporting would be more suitable here, I think.

lib/filterx/filterx-grammar.ym
```diff
@@ -346,6 +346,9 @@ expr_generator_unchecked
 	| list_generator
 	| regexp_search
 	| '(' expr_generator ')'		{ $$ = $2; }
+	| expr '+' expr_generator		{ $$ = filterx_operator_plus_generator_new($1, $3); }
+	| expr_generator '+' expr 		{ $$ = filterx_operator_plus_generator_new($1, $3); }
+	| expr_generator '+' expr_generator 	{ $$ = filterx_operator_plus_generator_new($1, $3); }
```
alltilla, line 372: [optional]

Factoring these out to `plus_generator` or something similar would make this a bit more tidy.

alltilla (commented):
Nice feature, good job :)

alltilla (approved):

PR #229, opened 2024-08-01
filterx: cleanup `message_value` handling

alltilla:
A common coding error we introduce is to forget to handle `message_value` types in our functions.

This PR introduces object-extractor which handles the value extraction for all built-in types and their `message_value` counterparts. I also replaced nearly every occurrence of getting an object's value.

The code change fixes some missing `message_value` handlings implicitly.

Also I think we improve the performance and add more functionality in some cases, e.g.:
* in some places we now skip calling `unmarshal()`/`eval_typed()` for creating a temporary object that we immediately unref (notably in subscript, attr and add expressions)
* now we support `message_value`s in subscript expression keys

### DIFF
--- a/lib/filterx/expr-comparison.c
+++ b/lib/filterx/expr-comparison.c
@@ -65,8 +65,8 @@ _convert_filterx_object_to_generic_number(FilterXObject *obj, GenericNumber *gn)
   UnixTime utime;
   if (filterx_object_extract_datetime(obj, &utime))
     {
-      uint64_t unix_epoch = unix_time_to_unix_epoch(utime);
-      gn_set_int64(gn, (uint64_t) unix_epoch);
+      guint64 unix_epoch = unix_time_to_unix_epoch(utime);
+      gn_set_int64(gn, (gint64) MIN(unix_epoch, G_MAXINT64));
       return;
     }
 
--- a/lib/filterx/expr-comparison.c
+++ b/lib/filterx/expr-comparison.c
@@ -24,6 +24,7 @@
 #include "filterx/expr-comparison.h"
 #include "filterx/object-datetime.h"
 #include "filterx/filterx-globals.h"
+#include "filterx/object-extractor.h"
 #include "filterx/object-primitive.h"
 #include "filterx/object-null.h"
 #include "filterx/object-string.h"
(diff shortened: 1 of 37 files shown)


alltilla:
rebased to master.

lib/filterx/expr-comparison.c
```diff
@@ -44,43 +45,43 @@ typedef struct _FilterXComparison
...
+      return;
     }
-  else
+
+  UnixTime utime;
+  if (filterx_object_extract_datetime(obj, &utime))
     {
-      gn_set_nan(gn);
+      uint64_t unix_epoch = unix_time_to_unix_epoch(utime);
+      gn_set_int64(gn, (uint64_t) unix_epoch);
```
OverOrion, line 69: I think you might have wanted to cast this to `gint64`

lib/filterx/expr-comparison.c
```diff
@@ -44,43 +45,43 @@ typedef struct _FilterXComparison
...
+      gn_set_int64(gn, 0);
+      return;
     }
-  else
+
+  UnixTime utime;
+  if (filterx_object_extract_datetime(obj, &utime))
     {
-      gn_set_nan(gn);
+      uint64_t unix_epoch = unix_time_to_unix_epoch(utime);
```
OverOrion, line 68: [Nit]
Why not `guint64`?

alltilla:
rebased to main

lib/filterx/object-datetime.c
```diff
@@ -126,28 +127,22 @@ filterx_typecast_datetime(FilterXExpr *s, GPtrArray *args)
       filterx_object_ref(object);
       return object;
     }
-  else if (filterx_object_is_type(object, &FILTERX_TYPE_NAME(integer)))
+
+  gint64 i;
+  if (filterx_object_extract_integer(object, &i))
```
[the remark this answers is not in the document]
alltilla: Thanks, fixed.

lib/filterx/expr-comparison.c
alltilla: Thanks, fixed.

lib/filterx/expr-comparison.c
alltilla: Thanks, fixed.

bazsi:
I like where this is going. I did not review all the call-site conversions in detail, but you were right, it was time to extract the these extractions into a separate set of functions.

One thing that we may want to consider (after merging this, not trying to block it), is to see if it would make sense to move these functions to the modules of the respective types, instead of a "magnet" module, which will need to grow every time we add a new type, e.g.
* filterx_object_extract_string() would go to the object-string module, while extract_datetime() would go to the object-datetime module.
* also, I'd look at if the typecast functions (which are now filterx simple functions) could be refactored to be the opposite of the extraction (because extract basically unwraps an object, while a typecast wraps generic value to a wrapped value)

lib/filterx/object-datetime.c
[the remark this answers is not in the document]
bshifter: i think int64_t would be enough here. it was probably my bad decision to use uint64_t

bshifter (approved):
what a great improvement of code quality!

PR #230, opened 2024-08-01
lib/filterx/object-json: fix missing json_double workaround in _deep_copy_filterx_object_ref

OverOrion:


alltilla (approved):
Approving but another review would be necessary as I was taking part of implementing the fix.

MrAnno (approved):

PR #232, opened 2024-08-02
github-actions: temporarily disable smpt

alltilla:
Until libesmtp is fixed in Debian Testing.

### DIFF
--- a/.github/workflows/devshell.yml
+++ b/.github/workflows/devshell.yml
@@ -49,6 +49,7 @@ jobs:
             --disable-java-modules
             --enable-ebpf
             --with-python=3
+            --disable-smtp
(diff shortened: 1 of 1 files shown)


alltilla:
well it seems like cmake has an amqp problem...

alltilla:
Also distcheck fails, I don't really want to work on more workarounds because of libesmtp. Hopefully it gets fixed relatively quickly in testing :/

MrAnno:
The CI should be green. If this is considered temporary, the workaround is straightforward:
#233

(Otherwise, we would have to install libesmtp in another way or change platform, but this seems unnecessary)

alltilla:
closing in favor of https://github.com/axoflow/axosyslog/pull/233

PR #241, opened 2024-08-07
filterx: generator functions support

alltilla:
This is a refactor PR, which introduces generator functions, and port `regexp_search()`.
In separate PRs we should port all of or parser functions to be generator functions.

---

Generator functions have the exact same syntax as regular functions, but we need to write separate rules in the grammar
to handle them, but if we don't know whether an identifier is a regular function name or generator function name, the
rules will conflict.

I solved this by preprocessing `LL_IDENTIFIER`s in the lexer, and trying to find a corresponding function or generator function for them, and returning separate tokens if we have found one: `LL_FILTERX_FUNC` and `LL_FILTERX_GEN_FUNC`.

### DIFF
--- a/lib/filterx/filterx-grammar.ym
+++ b/lib/filterx/filterx-grammar.ym
@@ -120,6 +120,7 @@ construct_template_expr(LogTemplate *template)
 %type <node> expr_value
 %type <node> expr_generator
 %type <node> expr_generator_unchecked
+%type <node> expr_plus_generator
 %type <node> generator_function_call
 %type <node> function_call
 %type <ptr> arguments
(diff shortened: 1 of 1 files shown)


lib/cfg-lexer.c
```diff
@@ -313,6 +314,29 @@ cfg_lexer_map_word_to_token(CfgLexer *self, CFG_STYPE *yylval, const CFG_LTYPE *
...
+      /*
+       * Regular functions and generator functions have the exact same syntax,
+       * but we need to have different tokens for them as they are handled in
+       * completely different grammar rules.
+       */
+
+      PluginContext *plugin_context = &self->cfg->plugin_context;
+
+      if ((self->cfg && plugin_is_plugin_available(plugin_context, LL_CONTEXT_FILTERX_FUNC, token)) ||
+          (self->cfg && plugin_is_plugin_available(plugin_context, LL_CONTEXT_FILTERX_SIMPLE_FUNC, token)) ||
```
bazsi, line 333: this would not be needed at all, if

1) we had a single context for all kinds of functions (why do we need separate ones?)
2) we pushed the right context from the grammar
3) builtin functions would also be registered as plugins

then we would have a single context based plugin lookup. 

I might be missing something though, so let's discuss.

bazsi:
This looks great, but let's discuss why we can't use the context based plugin lookup mechanism to cover all kinds of functions.

lib/cfg-lexer.c
alltilla: Probably the bottleneck is my understanding of the lexer/parser. I did not find the right spot to push the context, but it might be trivial. Sure, let's talk about this IRL, maybe there is a cleaner solution.

bazsi (approved):

lib/cfg-lexer.c
alltilla: Fixed in the way we have decided IRL.

OverOrion (approved):

PR #242, opened 2024-08-07
filterx: add SDATA helper functions

OverOrion:
```
@version: 4.8
@include "scl.conf"

log {

  source { example-msg-generator(num(1)
    template("[Originator@6876 sub=Vimsvc.ha-eventmgr opID=esxui-13c6-6b16 sid=5214bde6 user=root][anotherSDID@32473 iut=4 eventSource=Application eventID=1012]")
  ); };
  parser{sdata-parser();};
  filterx {
    bar = has_sdata();
    foo = is_sdata_from_enterprise("6876");
    # asd = get_sdata();
    vars();
    };
  destination { file("/dev/stdout"); };
};

```

```
[2024-08-07T14:57:58.003231] Setting value; name='.SDATA.Originator@6876.sub', value='Vimsvc.ha-eventmgr', type='string', msg='0x59da85103170', rcptid='1872'
[2024-08-07T14:57:58.003231] Setting value; name='.SDATA.Originator@6876.opID', value='esxui-13c6-6b16', type='string', msg='0x59da85103170', rcptid='1872'
[2024-08-07T14:57:58.003231] Setting value; name='.SDATA.Originator@6876.sid', value='5214bde6', type='string', msg='0x59da85103170', rcptid='1872'
[2024-08-07T14:57:58.003231] Setting value; name='.SDATA.Originator@6876.user', value='root', type='string', msg='0x59da85103170', rcptid='1872'
[2024-08-07T14:57:58.003231] Setting value; name='.SDATA.anotherSDID@32473.iut', value='4', type='string', msg='0x59da85103170', rcptid='1872'
[2024-08-07T14:57:58.003231] Setting value; name='.SDATA.anotherSDID@32473.eventSource', value='Application', type='string', msg='0x59da85103170', rcptid='1872'
[2024-08-07T14:57:58.003231] Setting value; name='.SDATA.anotherSDID@32473.eventID', value='1012', type='string', msg='0x59da85103170', rcptid='1872'
[2024-08-07T14:57:58.003231] <<<<<< parser rule evaluation result; result='accepted', rule='#anon-parser0', location='/source/install_dir/etc/asd.conf:9:10', msg='0x59da85103170', rcptid='1872'
[2024-08-07T14:57:58.003231] >>>>>> filterx rule evaluation begin; rule='#anon-filter0', location='/source/install_dir/etc/asd.conf:10:10', msg='0x59da85103170', rcptid='1872'
[2024-08-07T14:57:58.003231] FILTERX ESTEP; expr='/source/install_dir/etc/asd.conf:11:5|        bar = has_sdata()', value='true', truthy='1', type='boolean'
[2024-08-07T14:57:58.003231] FILTERX ESTEP; expr='/source/install_dir/etc/asd.conf:12:5|        foo = is_sdata_from_enterprise("6876")', value='true', truthy='1', type='boolean'
```

### DIFF
--- a/lib/filterx/func-sdata.c
+++ b/lib/filterx/func-sdata.c
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2024 Axoflow
+ * Copyright (c) 2024 Szilard Parrag
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "func-sdata.h"
+#include "filterx/object-primitive.h"
+#include "filterx/filterx-eval.h"
+#include "filterx/object-string.h"
+#include "filterx/object-extractor.h"
+
+#include <string.h>
+
+
+FilterXObject *
+filterx_simple_function_is_sdata_from_enterprise(FilterXExpr *s, GPtrArray *args)
+{
+  if (args == NULL || args->len != 1)
+    {
+      filterx_simple_function_argument_error(s, "Requires exactly one argument", FALSE);
+      return NULL;
+    }
+
+  FilterXObject *obj = filterx_typecast_string(NULL, args);
+  if (!obj)
+    return NULL;
+  const gchar *str_value;
+  gsize len;
+  if (!filterx_object_extract_string(obj, &str_value, &len))
+    return NULL;
+
+  gboolean contains = FALSE;
+  FilterXEvalContext *context = filterx_eval_get_context();
+  LogMessage *msg = context->msgs[0];
+
+  for (guint8 i = 0; i < msg->num_sdata && !contains; i++)
+    {
+      const gchar *value = log_msg_get_value_name(msg->sdata[i], NULL);
+      gchar *at_sign = strchr(value, '@');
+      if (!at_sign)
+        continue;
+      contains = strncmp(at_sign+1, str_value, len) == 0;
+    }
+  filterx_object_unref(obj);
+  return filterx_boolean_new(contains);
+}
--- a/lib/filterx/func-sdata.c
+++ b/lib/filterx/func-sdata.c
@@ -33,22 +33,43 @@
 
 #define SDATA_PREFIX_LEN (7)
 
-FilterXObject *
-filterx_simple_function_is_sdata_from_enterprise(FilterXExpr *s, GPtrArray *args)
+#define FILTERX_FUNC_IS_SDATA_FROM_ENTERPRISE_USAGE "Usage: is_sdata_from_enteprise(\"32473\")"
+
+typedef struct FilterXFunctionIsSdataFromEnteprise_
 {
-  if (args == NULL || args->len != 1)
+  FilterXFunction super;
+  gchar *str_literal;
+  gsize str_literal_len;
+} FilterXFunctionIsSdataFromEnteprise;
+
+static gboolean
+_extract_args(FilterXFunctionIsSdataFromEnteprise *self, FilterXFunctionArgs *args, GError **error)
+{
+  gsize len = filterx_function_args_len(args);
+  if (len != 1)
     {
-      filterx_simple_function_argument_error(s, "Requires exactly one argument", FALSE);
-      return NULL;
+      g_set_error(error, FILTERX_FUNCTION_ERROR, FILTERX_FUNCTION_ERROR_CTOR_FAIL,
+                  "invalid number of arguments. " FILTERX_FUNC_IS_SDATA_FROM_ENTERPRISE_USAGE);
+      return FALSE;
     }
 
-  FilterXObject *obj = filterx_typecast_string(NULL, args);
-  if (!obj)
-    return NULL;
-  const gchar *str_value;
-  gsize len;
-  if (!filterx_object_extract_string(obj, &str_value, &len))
-    return NULL;
+  gsize str_literal_len;
+  const gchar *str_literal = filterx_function_args_get_literal_string(args, 0, &str_literal_len);
+  if (!str_literal)
+    {
+      g_set_error(error, FILTERX_FUNCTION_ERROR, FILTERX_FUNCTION_ERROR_CTOR_FAIL,
+                  "argument must be a string literal. " FILTERX_FUNC_IS_SDATA_FROM_ENTERPRISE_USAGE);
+      return FALSE;
+    }
+  self->str_literal = g_strdup(str_literal);
+  self->str_literal_len = str_literal_len;
+  return TRUE;
+}
+
+static FilterXObject *
+_eval(FilterXExpr *s)
+{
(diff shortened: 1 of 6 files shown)


lib/filterx/func-sdata.c
```diff
@@ -0,0 +1,60 @@
...
+      filterx_simple_function_argument_error(s, "Requires exactly one argument", FALSE);
+      return NULL;
+    }
+
+  FilterXObject *obj = filterx_typecast_string(NULL, args);
+  if (!obj)
+    return NULL;
+  const gchar *str_value = filterx_string_get_value(obj, NULL);
+  if (!str_value)
+    return NULL;
```
bazsi, line 46: I think the filterx_object_extract_string() function would make this simpler, #229

lib/filterx/func-sdata.c
```diff
@@ -0,0 +1,75 @@
...
+
+  gboolean contains = FALSE;
+  FilterXEvalContext *context = filterx_eval_get_context();
+  LogMessage *msg = context->msgs[0];
+
+  for (guint8 i = 0; i < msg->num_sdata && !contains; i++)
+    {
+      const gchar *value = log_msg_get_value_name(msg->sdata[i], NULL);
+      gchar *res = strstr(value, str_value);
```
bazsi, line 55: I think we should add the "@" in front of the string value.

lib/filterx/func-sdata.c
```diff
@@ -0,0 +1,75 @@
...
+
+FilterXObject *
+filterx_simple_function_is_sdata_from_enterprise(FilterXExpr *s, GPtrArray *args)
+{
+  if (args == NULL || args->len != 1)
+    {
+      filterx_simple_function_argument_error(s, "Requires exactly one argument", FALSE);
+      return NULL;
+    }
+
```
bazsi, line 40: I expect this to be used a lot to narrow down messages to come from a given vendor.

so it would be a good improvement later (not now), to check if the argument is a literal string, because I'd expect in most cases it would be.  With that, we could be a lot faster, especially as the enterpriseId is usually in the first sdata.

bazsi (approved):
I'd love to see the new extract functions being used in the implementation, but it's approved from my part.

lib/filterx/func-sdata.c
OverOrion: Not sure about this as [RFC424](https://datatracker.ietf.org/doc/html/rfc5424#section-6.3.2) says that there are names without @:
>Names that do not contain an at-sign ("@", ABNF %d64) are reserved
      to be assigned by IETF Review as described in [BCP26](https://datatracker.ietf.org/doc/html/bcp26) [[RFC5226](https://datatracker.ietf.org/doc/html/rfc5226)].
      Currently, these are the names defined in [Section 7](https://datatracker.ietf.org/doc/html/rfc5424#section-7).

lib/filterx/func-sdata.c
OverOrion: Good point, switched to them!

lib/filterx/func-sdata.c
bazsi: Yeah, but the function is there to find sdata coming from a specific enterprise.

We may need a function eventually to check for a specific block, but then that should not be named is_sdata_from_enterprise()

OverOrion: Yes, thank you, forgot about that part for a moment! :sweat_smile:

lib/filterx/func-sdata.c
```diff
@@ -0,0 +1,64 @@
...
+#include "filterx/object-primitive.h"
+#include "filterx/filterx-eval.h"
+#include "filterx/object-string.h"
+#include "filterx/object-extractor.h"
+
+#include <string.h>
+
+
+FilterXObject *
+filterx_simple_function_is_sdata_from_enterprise(FilterXExpr *s, GPtrArray *args)
```
alltilla, line 34: I'd rather implement this now as a non-simple function, and enforce the argument to be a string literal. We could cache its `char *` and skip a lot of `eval()` calls.

lib/filterx/func-sdata.c
```diff
@@ -0,0 +1,297 @@
...
+static gboolean
+_generate(FilterXExprGenerator *s, FilterXObject *fillable)
+{
+  FilterXEvalContext *context = filterx_eval_get_context();
+  LogMessage *msg = context->msgs[0];
+
+  ScratchBuffersMarker marker;
+  GString *current_sd_id = scratch_buffers_alloc_and_mark(&marker);
+  GString *next_sd_id = scratch_buffers_alloc();
+  GString *param_name = scratch_buffers_alloc();
```
alltilla, line 197: Copying to `GString`s is slow and unnecessary. The `sd_id` and `sd_param` strings could be `const gchar *`s with `gsize` len, kind of "Slices" in the original string.

lib/filterx/func-sdata.c
```diff
@@ -0,0 +1,290 @@
...
+
+static FilterXObject *
+_create_container(FilterXExprGenerator *s, FilterXExpr *fillable_parent)
+{
+  FilterXObject *fillable_parent_obj = filterx_expr_eval_typed(fillable_parent);
+  if (!fillable_parent_obj)
+    return NULL;
+
+  return filterx_object_create_dict(fillable_parent_obj);
+}
```
alltilla, line 258: Shouldn't we `unref()` the `fillable_parent_obj`?

lib/filterx/func-sdata.c
```diff
@@ -0,0 +1,290 @@
...
+
+static gboolean
+_extract_sd_components(const gchar *key, gssize key_len, const gchar **sd_id_start, gsize *sd_id_len,
+                       const gchar **sd_param_name_start, gsize *sd_param_name_len)
+{
+  /* skip '.SDATA.' */
+  if (key_len < SDATA_PREFIX_LEN)
+    return FALSE;
+
+  const gchar *last_dot = strrchr(key, '.');
```
alltilla, line 153: This expects a NUL terminated string which is not guaranteed here.

lib/filterx/func-sdata.c
```diff
@@ -0,0 +1,290 @@
...
+  gboolean res = filterx_object_set_subscript(dict, fob_key, &fob_value);
+
+  filterx_object_unref(fob_key);
+  filterx_object_unref(fob_value);
+
+  return res;
+}
+
+static gboolean
+_generate(FilterXExprGenerator *s, FilterXObject *fillable)
```
alltilla, line 181: This function became really complex with the double loop and multiple iterator indexes being incremented and set manually. I now understand what is happening, but it could and should be a lot cleaner to be easily maintainable. Please refactor this function. I can help if you get stuck.

lib/filterx/func-sdata.c
```diff
@@ -0,0 +1,313 @@
...
+        goto error;
+      i += num_insertions + 1;
+
+      FilterXObject *sd_id_key = filterx_string_new(current_sd_id_start, current_sd_id_len);
+      filterx_object_set_subscript(fillable, sd_id_key, &inner_dict);
+      filterx_object_unref(inner_dict);
+      filterx_object_unref(sd_id_key);
+    }
+
+error:
```
alltilla, line 267: [nitpick]

It is strange to see that it is possible to go to the error returning part with `success = TRUE`, which does not return an error. I think we should either call this `exit`, or make `error` not depend on the value of `success`, or simply get rid of the gotos.

I would prefer the latter: usually we need the `goto`s, so it is easier to destroy local variables, but that is not needed here, so we ca just `return FALSE` instead of the lot of `goto`s. We might even be able to remove the `success` variable, too.

jszigetvari:
subscribe .

lib/filterx/func-sdata.c
```diff
@@ -0,0 +1,313 @@
...
+
+static FilterXObject *
+_create_container(FilterXExprGenerator *s, FilterXExpr *fillable_parent)
+{
+  FilterXObject *fillable_parent_obj = filterx_expr_eval_typed(fillable_parent);
+  if (!fillable_parent_obj)
+    return NULL;
+
+  filterx_object_unref(fillable_parent_obj);
+  return filterx_object_create_dict(fillable_parent_obj);
```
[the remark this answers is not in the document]
alltilla: This is my only must fix comment.

lib/filterx/func-sdata.c
OverOrion: You are absolutely right there, once I got rid of the things that would have required a cleanup this goto became obsolete, thanks!

alltilla (approved):
Nice feature, thanks! :)

PR #243, opened 2024-08-08
Fix disappearing `metrics-probe()` counters

MrAnno:
The TLS cache of `metrics-probe()` has been extended with a global cache in order not to make metrics orphaned whenever an ivykis worker stops (after 10 seconds of inactivity).

The global cache requires a lock, but it is only updated on thread shutdown. It is purged during reload.

Refactor overview:
- metrics-cache --> dyn-metrics-store
- metrics-tls-cache --> dyn-metrics-cache

I've also tried to decouple metrics-template from the cache functionality, but it would have had performance implications, so I only added a comment about the coupled functionality.


MrAnno:
@kira-syslogng do stresstest

bazsi (approved):
This looks good. Great work.

PR #249, opened 2024-08-10
csvparser: fix filterx race condition

MrAnno:
CSVScannerOptions must be local when it is modified by _eval().


modules/csvparser/filterx-func-parse-csv.c
```diff
@@ -88,6 +88,27 @@ _parse_list_argument(FilterXFunctionParseCSV *self, FilterXExpr *list_expr, GLis
...
+    {
+      csv_scanner_options_copy(local_opts, &self->options);
+      opts = local_opts;
+    }
+
+  if (string_delimiters)
+    csv_scanner_options_set_string_delimiters(local_opts, string_delimiters);
+
+  if (cols)
+    csv_scanner_options_set_expected_columns(local_opts, g_list_length(cols));
```
bazsi, line 107: that g_list_length() is going to bite us, but I guess that's coming in a separate PR.

bazsi (dismissed):

modules/csvparser/filterx-func-parse-csv.c
MrAnno: Let's fix it here :)

modules/csvparser/filterx-func-parse-csv.c
```diff
@@ -114,21 +134,17 @@ _eval(FilterXExpr *s)
...
-      csv_scanner_options_set_expected_columns(&self->options, g_list_length(cols));
-      result = filterx_json_object_new_empty();
-    }
+    result = filterx_json_object_new_empty();
   else
     result = filterx_json_array_new_empty();
 
-  csv_scanner_init(&scanner, &self->options, input);
+  CSVScanner scanner;
+  CSVScannerOptions local_opts = {0};
```
bazsi, line 146: if we are not initializing scanner, we might get away with not initializing local_opts either.

MrAnno: I need to null-init the scanner, because I call its destructor unconditionally.

modules/csvparser/filterx-func-parse-csv.c
MrAnno: Fixed.

modules/csvparser/filterx-func-parse-csv.c
```diff
@@ -71,6 +72,9 @@ _parse_list_argument(FilterXFunctionParseCSV *self, FilterXExpr *list_expr, GLis
   if (!filterx_object_len(list_obj, &size))
     return FALSE;
 
+  if (list_length)
+    *list_length = size;
+
```
bazsi, line 77: I don't see why we need to extract the list argument into a GList, we could just as easily use the list directly, couldn't we?

Earlier, extracting the string values from a list was cumbersome, but now with filterx_object_extract_string() this is not any more difficult than iterating a GList with strdupped strings on it.

But like I said with my previous comment, we should probably push this into a separate PR.

PR #251, opened 2024-08-12
filterx: `parse_xml()`

alltilla:

### DIFF
--- a/modules/xml/filterx-parse-xml.c
+++ b/modules/xml/filterx-parse-xml.c
@@ -1,0 +1,701 @@
+/*
+ * Copyright (c) 2024 Axoflow
+ * Copyright (c) 2024 Attila Szakacs <attila.szakacs@axoflow.com>
+ *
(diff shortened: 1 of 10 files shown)


bshifter (dismissed):

modules/xml/filterx-parse-xml.c
```diff
@@ -0,0 +1,702 @@
...
+      gchar *error_info = g_strdup(error ? error->message : "unknown error");
+      filterx_eval_push_error_info("failed to parse xml", &self->super.super.super, error_info, TRUE);
+      if (error)
+        g_error_free(error);
+      goto exit;
+    }
+
+exit:
+  _elem_context_free(root_elem_context);
+  g_queue_free(obj_stack);
```
MrAnno, line 613: In case of errors, the XmlElemContexts in the queue need to be freed (unclosed tags).

alltilla: Thanks, fixed.

PR #253, opened 2024-08-12
csvparser-filterx: do not traverse column list 3 times

MrAnno:
Depends on #249

csvparser-filterx: do not traverse column list 3 times

- constructing list
- g_list_length()
- traversing list

This also fixes a possible bug where a name-value pair passed to `filterx_string_new()` may not have been null-terminated.

### DIFF
--- a/modules/csvparser/filterx-func-parse-csv.c
+++ b/modules/csvparser/filterx-func-parse-csv.c
@@ -141,17 +141,11 @@ static inline gboolean
 _fill_object_col(FilterXObject *cols, gint64 index, CSVScanner *scanner, FilterXObject *result)
 {
   FilterXObject *col = filterx_list_get_subscript(cols, index);
-  const gchar *col_name;
-  gsize col_name_len;
-  filterx_object_extract_string(col, &col_name, &col_name_len);
-
-  FilterXObject *key = filterx_string_new(col_name, col_name_len);
   FilterXObject *val = filterx_string_new(csv_scanner_get_current_value(scanner),
                                           csv_scanner_get_current_value_len(scanner));
 
-  gboolean ok = filterx_object_set_subscript(result, key, &val);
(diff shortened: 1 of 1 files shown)


bazsi:
:+1:

bazsi (dismissed):

bazsi (dismissed):

bazsi (approved):

alltilla:
I think this is a good step forward.

Recently we have added a support for passing list/dict literals to function arguments. With that we don't need to evaluate the list over and over again, and we can initialize even a `GPtrArray` with the column names in init time. The next step should be supporting this IMO.

For example see: https://github.com/MrAnno/axosyslog/blob/10376b4a6a158999699bb9ba05ec11ead88a2a6e/lib/filterx/filterx-metrics-labels.c#L328

modules/csvparser/filterx-func-parse-csv.c
```diff
@@ -103,12 +103,61 @@ _init_scanner(FilterXFunctionParseCSV *self, GList *string_delimiters, GList *co
...
+
+static inline gboolean
+_fill_col_object_field(FilterXObject *cols, gint64 index, CSVScanner *scanner, FilterXObject *result)
+{
+  FilterXObject *col = filterx_list_get_subscript(cols, index);
+  const gchar *col_name;
+  gsize col_name_len;
+  filterx_object_extract_string(col, &col_name, &col_name_len);
+
+  FilterXObject *key = filterx_string_new(col_name, col_name_len);
```
alltilla, line 148: Why don't we just use `col`? It seems like we have recreated the same string. `set_subscript()` support `message_value`s as well, so there is no need to ensure `FilterXString`.

MrAnno: I didn't notice about this, I just refactored this part. 
I've removed the unnecessary string recreation.

PR #254, opened 2024-08-13
fix(lib): do not insert "nulls" as values for __VARARGS__ config arguments

### LINKED ISSUE #244: crash while passing an empty parameter through chain of scl blocks
Dear maintainers, I noticed there is some kind of crash when I want to pass an `""` parameter through multiple embedded scl blocks. Could you take a look?

Reproduction

```
@version: current

block destination second(...) {
      `__VARARGS__`
};

block destination first(...) {
      second(`__VARARGS__`);
};

log {
   destination{ first(some-param("")) };
};
```

Backtrace:

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7e4967b in _report_generator_args (key=0x5555556075b0, value=0x0, user_data=0x55555561dd40)
    at ../lib/cfg-block-generator.c:42
42	  for (const gchar *c = (const gchar *) value; *c; c++)
(gdb) bt
#0  0x00007ffff7e4967b in _report_generator_args (key=0x5555556075b0, value=0x0, user_data=0x55555561dd40)
    at ../lib/cfg-block-generator.c:42
#1  0x00007ffff74fb9bb in g_hash_table_foreach (hash_table=0x5555555b8420, 
    func=0x7ffff7e493f3 <_report_generator_args>, user_data=0x55555561dd40) at ../glib/glib/ghash.c:2117
#2  0x00007ffff7e484d3 in cfg_args_foreach (self=0x55555561dd80, func=0x7ffff7e493f3 <_report_generator_args>, 
    user_data=0x55555561dd40) at ../lib/cfg-args.c:53
#3  0x00007ffff7e497a8 in cfg_block_generator_generate (self=0x5555556099b0, cfg=0x555555584790, 
    args=0x55555561dd80, result=0x55555561dd40, reference=0x7fffffff99a0 "etc/syslog-ng.conf:8:7")
    at ../lib/cfg-block-generator.c:61
#4  0x00007ffff7e4c238 in cfg_lexer_parse_and_run_block_generator (self=0x55555560cb60, p=0x555555609a10, 
    yylval=0x7fffffff9d50) at ../lib/cfg-lexer.c:1083
#5  0x00007ffff7e4c4b5 in cfg_lexer_preprocess (self=0x55555560cb60, tok=10430, yylval=0x7fffffff9d50, 
    yylloc=0x7fffffff9d90) at ../lib/cfg-lexer.c:1151
#6  0x00007ffff7e4c693 in cfg_lexer_lex (self=0x55555560cb60, yylval=0x7fffffff9d50, yylloc=0x7fffffff9d90)
    at ../lib/cfg-lexer.c:1224
#7  0x00007ffff7e4e289 in main_lex (yylval=0x7fffffff9d50, yylloc=0x7fffffff9d90, lexer=0x55555560cb60)
    at ../lib/cfg-parser.c:233
#8  0x00007ffff7ea508f in main_parse (lexer=0x55555560cb60, dummy=0x7fffffffdaf8, arg=0x0) at lib/cfg-grammar.c:3867
#9  0x00007ffff7e4e521 in cfg_parser_parse (self=0x7ffff7fb6060 <main_parser>, lexer=0x55555560cb60, 
    instance=0x7fffffffdaf8, arg=0x0) at ../lib/cfg-parser.c:299
#10 0x00007ffff7e47a19 in cfg_run_parser (self=0x555555584790, lexer=0x55555560cb60, 
    parser=0x7ffff7fb6060 <main_parser>, result=0x7fffffffdaf8, arg=0x0) at ../lib/cfg.c:534
#11 0x00007ffff7e48039 in cfg_read_config (self=0x555555584790, fname=0x555555563320 "etc/syslog-ng.conf", 
    preprocess_into=0x0) at ../lib/cfg.c:655
#12 0x00007ffff7e71e43 in main_loop_read_and_init_config (self=0x7ffff7fba3c0 <main_loop>) at ../lib/mainloop.c:680
#13 0x0000555555556ac7 in main (argc=1, argv=0x7fffffffdcd8) at ../syslog-ng/main.c:319
```

OverOrion:
```
@version: current

block destination second(...) {
      `__VARARGS__`
};

block destination first(...) {
      second(`__VARARGS__`);
};

log {
   destination{ first(some-param("")) };
};
```

After the fix:

```
[2024-10-15T09:49:23.089776] Unknown argument, adding it to __VARARGS__; argument='some_param', value='', reference='../install_dir/etc/empty.conf:12:17'
[2024-10-15T09:49:23.089776] Unknown argument, adding it to __VARARGS__; argument='some_param', value='', reference='../install_dir/etc/empty.conf:8:7'

Error parsing log statement: syntax error, unexpected LL_IDENTIFIER, expecting '}'

In ../install_dir/etc/empty.conf:4:7-4:17:
-1      
0       #Start Block block destination second() at ../install_dir/etc/empty.conf:3
1       ## some_param=
2       @line "../install_dir/etc/empty.conf" 3 31
3       
4----->       some_param("") 
4----->       ^^^^^^^^^^
5       
6       #End Block block destination second() at ../install_dir/etc/empty.conf:3
7       

Included from ../install_dir/etc/empty.conf:8:7-8:30:
3       
4       #Start Block block destination first() at ../install_dir/etc/empty.conf:7
5       ## some_param=
6       @line "../install_dir/etc/empty.conf" 7 30
7       
8----->       second(some_param("") );
8----->       ^^^^^^^^^^^^^^^^^^^^^^^
9       
10      #End Block block destination first() at ../install_dir/etc/empty.conf:7
11      

Included from ../install_dir/etc/empty.conf:12:17-12:38:
7       block destination first(...) {
8             second(`__VARARGS__`);
9       };
10      
11      log {
12---->    destination{ first(some-param("")) };
12---->                 ^^^^^^^^^^^^^^^^^^^^^
13      };
```

Fixes #244.

### DIFF
--- a/lib/cfg-block-generator.c
+++ b/lib/cfg-block-generator.c
@@ -39,6 +39,7 @@ _report_generator_args(gpointer key, gpointer value, gpointer user_data)
 {
   GString *result = (GString *) user_data;
   g_string_append_printf(result, "## %s=", (gchar *) key);
+  value = value ? : "";
   for (const gchar *c = (const gchar *) value; *c; c++)
     {
       if (*c == '\n' && *(c + 1))


furiel (approved):

OverOrion:
On a second thought this might not be a proper fix:

```
[2024-08-13T09:19:13.979827] Unknown argument, adding it to __VARARGS__; argument='some_param', value='', reference='/source/install_dir/etc/foo.conf:12:17'
[2024-08-13T09:19:13.979827] Unknown argument, adding it to __VARARGS__; argument='some_param', value='(null)', reference='/source/install_dir/etc/foo.conf:8:7'
```

The value gets lost somewhere :eyes:

MrAnno (changes requested):
As far as I remember this only happens through multi-level passes of `__VARARGS__` with empty values, so the real fix might be somewhere else and will be something about how we handle the empty string parameter passed through these varargs.

alltilla:
Another config that crashes at the same location:
```
@version: current

block destination a(filtr(def)) {
  channel {
    filter {
      `filtr`
      and 1 == 1};
  };
};

destination d { a(filtr()); };
```

lib/cfg-block-generator.c
```diff
@@ -39,6 +39,8 @@ _report_generator_args(gpointer key, gpointer value, gpointer user_data)
 {
   GString *result = (GString *) user_data;
   g_string_append_printf(result, "## %s=", (gchar *) key);
+  if (!value)
+    return;
```
orymate, line 43: don't we need the trailing \n?

OverOrion: I opted for fixing this at a different layer in the end (I think this is the right patch).

OverOrion:
Updated the PR to properly handle empty `___VARARG__`  values instead.

MrAnno:
This is getting closer, it fixes the crash with the above config snippet but changes the output if you type `some-param()` to `some-param("")` inside the block (but it crashes before that because of the same null-ptr deref issue).

I think the issue is in the lexer, it removes `""` when passing parameters.
I'm investigating this and propose my patch soon.

MrAnno:
```
<block>\({white}*\)        { yy_pop_state(yyscanner); yylval->cptr = NULL; return LL_BLOCK; }
<block>\({white}*\"\"{white}*\) { yy_pop_state(yyscanner); yylval->cptr = g_strdup(""); return LL_BLOCK; }
<block>\({white}*\'\'{white}*\) { yy_pop_state(yyscanner); yylval->cptr = g_strdup(""); return LL_BLOCK; }
```

These are the ones removing what we need.
The problem here is probably the context of where we use these:
- we need to remove quotes in the declaration of block params
- we should NOT remove quotes when we pass arguments to blocks (when we "call" them)

MrAnno:
I can't fix this without reimplementing https://github.com/axoflow/axosyslog/commit/f0650bb78fecbc999d5400d95a7542505aa59896 completely, which would be a lot more work.

https://github.com/syslog-ng/syslog-ng/pull/2088#issuecomment-396034252 says the following:
> As "" is not mapped into "" anymore, this is a problem needs to be addressed. Luckily, we could break the cycle here, do not need to find another symbol that would be mapped to "", of which again find a symbol that maps to the underlying symbol etc. This can be addressed by modifying such scl. I show an example for this:

Unfortunately, this is not so lucky with varargs. With the current implementation, it is impossible to pass an empty string parameter through `varargs`. With your current patch, it is possible, but then it breaks the fully empty case.

So in order to fix the crash, I think we could go with this (which is very similar to your original patch, except for the new-line ending):
```diff
diff --git a/lib/cfg-block-generator.c b/lib/cfg-block-generator.c
index 9374c58e8..00811f278 100644
--- a/lib/cfg-block-generator.c
+++ b/lib/cfg-block-generator.c
@@ -39,6 +39,7 @@ _report_generator_args(gpointer key, gpointer value, gpointer user_data)
 {
   GString *result = (GString *) user_data;
   g_string_append_printf(result, "## %s=", (gchar *) key);
+  value = value ? : "";
   for (const gchar *c = (const gchar *) value; *c; c++)
     {
       if (*c == '\n' && *(c + 1))
```

Because of this:
https://github.com/axoflow/axosyslog/blob/f074c0fc1181fd87cbbba37b75b264d75545947b/lib/cfg-lexer-subst.c#L257

And we can open a separate issue about the impossibility to pass empty strings through varargs.
@OverOrion Is that okay with you?

MrAnno (approved):

PR #257, opened 2024-08-23
Fix race condition in FilterX simple funcs and perf improvements

MrAnno:
2 race conditions have been fixed around `FilterXFunctionArgs`:
- `retrieved` field should not be changed during eval
- expression `ref()/unref()` calls should not be called during eval

Depends on #252

### DIFF
--- a/lib/filterx/expr-function.c
+++ b/lib/filterx/expr-function.c
@@ -47,12 +47,42 @@ typedef struct _FilterXSimpleFunction
   FilterXSimpleFunctionProto function_proto;
 } FilterXSimpleFunction;
 
+struct _FilterXFunctionArgs
+{
+  GPtrArray *positional_args;
+  GHashTable *named_args;
+};
+
 void
 filterx_simple_function_argument_error(FilterXExpr *s, gchar *error_info, gboolean free_info)
 {
   FilterXSimpleFunction *self = (FilterXSimpleFunction *) s;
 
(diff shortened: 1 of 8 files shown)


lib/filterx/expr-function.c
```diff
@@ -115,23 +136,53 @@ static void
...
 {
   FilterXSimpleFunction *self = g_new0(FilterXSimpleFunction, 1);
 
   filterx_function_init_instance(&self->super, function_name);
   self->super.super.eval = _simple_eval;
   self->super.super.free_fn = _simple_free;
-  self->args = args;
   self->function_proto = function_proto;
 
+  self->args = g_ptr_array_new_full(8, (GDestroyNotify) filterx_expr_unref);
```
bshifter, line 174: this 8 seems a little bit ad hoc to me, especially you can have the number of arguments there

bshifter (approved):

lib/filterx/expr-function.c
MrAnno: It's based on this:
`self->positional_args = g_ptr_array_new_full(8, (GDestroyNotify) filterx_function_arg_free);`

8 is just the initial capacity here.

MrAnno: I understood. Fixed.

bshifter (approved):

PR #258, opened 2024-08-27
filterx: perf improvement and fix race condition in readonly FilterXJson::to_json_literal()

MrAnno:
json_object_to_json_string_ext() writes/cached into the given JSON object, so it's not thread safe.

### DIFF
--- a/lib/filterx/object-json-array.c
+++ b/lib/filterx/object-json-array.c
@@ -41,6 +41,9 @@ struct FilterXJsonArray_
   FilterXList super;
   FilterXWeakRef root_container;
   struct json_object *jso;
+
+  GMutex lock;
+  const gchar *cached_ro_literal;
 };
 
 static gboolean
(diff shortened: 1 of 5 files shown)


alltilla (approved):
Nice catch, sorry for introducing this problem.

What would you think about making `make_readonly()` virtual and implementing the json specific parts there, so we don't need separate `json_make_readonly()` functions?

MrAnno:
Much better, thanks.

alltilla:
I think we could replace the make_readonly calls with an assert, and make sure to take care of it in the json and otel filterx object implementations. Maybe the assert could be in the list and dict interface getters.

The json object caches all the values, so we can do a deep traversal and make everything readonly when the root object becomes readonly, otel does not cache, it creates a new object, so we can must it readonly instantly.

Let's talk about this IRL.

Thanks for the investigation!

lib/filterx/object-json-array.c
```diff
@@ -231,13 +243,29 @@ _unset_index(FilterXList *s, guint64 index)
   return TRUE;
 }
 
+static void
+_make_readonly(FilterXObject *s)
+{
+  FilterXJsonArray *self = (FilterXJsonArray *) s;
+
+  /* racy, but the result will be consistent */
```
alltilla, line 251: [optional]

We could check `g_atomic_pointer_get(&self->cached_ro_literal)` again after locking to eliminate this raciness.

alltilla (approved):
One minor suggestion, but it is completely okay for me to merge this as is. Let me know if you don't want to change the code based on my suggestion, and I will merge the PR.

Thanks!

lib/filterx/object-json-array.c
MrAnno: Will do

alltilla (approved):
Awesome, this turned out really nice :)

PR #261, opened 2024-08-29
cfg-lexer: fix handling of glob return values on musl

bazsi:
In case an error happens during glob(), glibc returns GLOB_NOMATCH while musl returns GLOB_ABORTED. Handle both properly.

### DIFF
--- a/lib/cfg-lexer.c
+++ b/lib/cfg-lexer.c
@@ -686,9 +686,7 @@ cfg_lexer_include_file_glob_at(CfgLexer *self, CfgIncludeLevel *level, const gch
 
   r = glob(pattern, GLOB_NOMAGIC, _cfg_lexer_glob_err, &globbuf);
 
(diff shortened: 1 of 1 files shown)


bazsi:
One additional context: glibc seems to ignore ENOTDIR errors, while musl does not. So to trigger the issue, one needs to 

```
@include "foo/*/*.conf"
```

and then place a symlink to "foo/" that points to an existing file. In this case, glob finds the symlink, tries to use it as a directory, fails with ENOTDIR. On musl, this calls the glob error callback and prints:

```
Error processing path for inclusion; path='/etc/syslog-ng/foo/symlink/', errno='Not a directory (20)'
```
and then returns GLOB_ABORTED.

Our wrapper function was incorrectly returning success in this case, without the include file list being populated and then it crashes.

MrAnno (approved):

PR #272, opened 2024-09-03
otel/ipv6: respect ipv6 compile switch

MrAnno:
Cherry-picked from syslog-ng/syslog-ng#5084 (@HofiOne)

### DIFF
--- a/modules/grpc/otel/otel-protobuf-parser.hpp
+++ b/modules/grpc/otel/otel-protobuf-parser.hpp
@@ -23,6 +23,8 @@
 #ifndef OTEL_PROTOBUF_PARSER_HPP
 #define OTEL_PROTOBUF_PARSER_HPP
 
+#include "syslog-ng.h"
+
 #include "otel-protobuf-parser.h"
 
 #include "compat/cpp-start.h"
--- a/modules/grpc/otel/otel-protobuf-parser.cpp
+++ b/modules/grpc/otel/otel-protobuf-parser.cpp
@@ -255,14 +255,16 @@ _extract_saddr(const grpc::string &peer)
       else
         host = peer.substr(first + 1, last - first - 1);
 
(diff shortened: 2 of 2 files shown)


modules/grpc/otel/otel-protobuf-parser.hpp
```diff
@@ -26,6 +26,7 @@
 #include "otel-protobuf-parser.h"
 
 #include "compat/cpp-start.h"
+#include "syslog-ng.h"
```
HofiOne, line 29: just curious, all the other `.hpp` headers include `syslog.h` outside the compact block, why the extern "C" is needed in this case?

MrAnno: I will check this. The cpp-helper says this:

'syslog-ng.h needs to come first even in cpp/hpp files.  You don't need cpp-start/end wrapper around syslog-ng.h.`

But I am not sure this is correct, because syslog-ng.h pulls glib and glib does some macro magic completely differently in case of C++ and C.

MrAnno: I remembered completely the opposite of the truth.
We need syslog-ng.h outside of the external C block especially because of the glib behavior.

Thanks for your comment. I will drop the last patch.

HofiOne: also, i can see this in the `syslog-ng.h`

```
 * Header ordering:
 *   - syslog-ng.h needs to come first even in cpp/hpp files.  You don't
 *     need cpp-start/end wrapper around syslog-ng.h.  This include can
 *     happen indirectly, e.g.  if you include any other header that should
 *     take care of syslog-ng.h
```

HofiOne (approved):

alltilla (approved):

PR #275, opened 2024-09-05
filterx: rework of unset_empties

bshifter:
- Default empty set: "", null, [], {}.
    - Optional named arguments:
      - recursive: Enables recursive processing of nested dictionaries.
      - ignorecase: Enables case-insensitive matching.
      - replacement: Specifies a filterxobject to replace target elements instead of removing them.
      - targets: A list of elements to identify for removal or replacement, clearing the default empty set.
    - UTF-8 support.
    - ignorecase and recursive args are defaults to TRUE
    
    This change removes elements from the given dictionary or list that match
    the empty set. If the recursive argument is provided, the function will
    process nested dictionaries as well. The replacement argument allows
    replacing target elements with a specified object, and the targets
    argument customizes which elements are removed or replaced, overriding
    the default empty set.

```
Usage: unset_empties(object, recursive=bool, ignorecase=bool, targets=[objects...], replacement=str);
```

example:
```
unset_empties(js1, targets=["foo", "bar", null, "", [], {}], ignorecase=false, replacement="N/A", recursive=false);
```

### DIFF
--- a/lib/filterx/expr-literal-generator.c
+++ b/lib/filterx/expr-literal-generator.c
@@ -302,3 +302,10 @@ filterx_expr_is_literal_list_generator(FilterXExpr *s)
   FilterXExprGenerator *generator = (FilterXExprGenerator *) s;
   return filterx_expr_is_generator(s) && generator->create_container == filterx_generator_create_list_container;
 }
+
+guint
+filterx_expr_literal_generator_len(FilterXExpr *s)
+{
+  FilterXExprLiteralGenerator *self = (FilterXExprLiteralGenerator *) s;
+  return g_list_length(self->elements);
+}
--- a/lib/filterx/expr-literal-generator.c
+++ b/lib/filterx/expr-literal-generator.c
@@ -290,17 +290,37 @@ filterx_literal_inner_list_generator_new(FilterXExpr *root_literal_generator, GL
 }
 
 gboolean
+_filterx_expr_is_inner_dict_generator(FilterXExpr *s)
+{
+  return s && (s->eval == _inner_dict_generator_eval);
+}
+
+gboolean
+_filterx_expr_is_inner_list_generator(FilterXExpr *s)
+{
+  return s && (s->eval == _inner_list_generator_eval);
+}
+
+gboolean
 filterx_expr_is_literal_dict_generator(FilterXExpr *s)
 {
   FilterXExprGenerator *generator = (FilterXExprGenerator *) s;
-  return filterx_expr_is_generator(s) && generator->create_container == filterx_generator_create_dict_container;
+  return (filterx_expr_is_generator(s) && generator->create_container == filterx_generator_create_dict_container)
+         || _filterx_expr_is_inner_dict_generator(s);
 }
 
 gboolean
 filterx_expr_is_literal_list_generator(FilterXExpr *s)
 {
   FilterXExprGenerator *generator = (FilterXExprGenerator *) s;
-  return filterx_expr_is_generator(s) && generator->create_container == filterx_generator_create_list_container;
+  return (filterx_expr_is_generator(s) && generator->create_container == filterx_generator_create_list_container)
+         || _filterx_expr_is_inner_list_generator(s);
+}
+
+gboolean
+filterx_expr_is_literal_generator(FilterXExpr *s)
+{
+  return filterx_expr_is_literal_list_generator(s) || filterx_expr_is_literal_dict_generator(s);
 }
 
 guint
--- a/lib/filterx/func-flags.h
+++ b/lib/filterx/func-flags.h
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) 2023 Axoflow
+ * Copyright (c) 2024 shifter
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#ifndef FUNC_FLAGS_H
+#define FUNC_FLAGS_H
+
+#include "syslog-ng.h"
+
+#define MAX_ENUMS 64
+
+#define STATIC_ASSERT(COND) G_STATIC_ASSERT(COND)
+
+#define DEFINE_FUNC_FLAGS(ENUM_NAME, ...) \
+    typedef enum { __VA_ARGS__, ENUM_NAME##_MAX} ENUM_NAME; \
+    STATIC_ASSERT(ENUM_NAME##_MAX <= MAX_ENUMS) \
+
+#define FUNC_FLAGS_ITER(ENUM_NAME, CODE) for (guint64 enum_elt = 0; enum_elt < ENUM_NAME##_MAX; enum_elt++) { CODE; };
+
+#define FLAG_VAL(ID) (1 << ID)
+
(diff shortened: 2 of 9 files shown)


lib/filterx/func-unset-empties.c
```diff
@@ -46,20 +52,16 @@ static gboolean _process_list(FilterXFunctionUnsetEmpties *self, FilterXObject *
...
     {
       return TRUE;
     }
 
+  gsize str_len;
+  const gchar *str;
+  if (filterx_object_extract_string(obj, &str, &str_len) && str_len == 0)
+    return TRUE;
+
   if (filterx_object_is_type(obj, &FILTERX_TYPE_NAME(dict)) ||
```
alltilla, line 65: It's a bit strange for me that the user cannot change the behavior of handling `list`s, `dict`s and `null`s.
I don't know if they want to do it, but this seems asymmetric. If as a user I were to set the target strings, and `null`s and empty `dict`s were getting removed, I would be surprised.

lib/filterx/func-unset-empties.c
```diff
@@ -216,28 +249,105 @@ _extract_object_expr(FilterXFunctionArgs *args, GError **error)
...
+                  FILTERX_FUNC_UNSET_EMPTIES_ARG_NAME_REPLACEMENT" argument must be literal. " FILTERX_FUNC_UNSET_EMPTIES_USAGE);
+      return FALSE;
+    }
+
+  self->replacement = value;
+  return TRUE;
+}
+
+static gboolean
+_extract_target_objects(FilterXFunctionUnsetEmpties *self, FilterXFunctionArgs *args, GError **error)
```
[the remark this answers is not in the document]
bshifter: The feeature does not allow to remove targets at the moment, this way i think it's more handy to not to set default targets except the real empties.

lib/filterx/func-unset-empties.c
```diff
@@ -68,6 +70,19 @@ _should_unset(FilterXFunctionUnsetEmpties *self, FilterXObject *obj)
...
 
+  if (!str)
+    return FALSE;
+
+  guint num_targets = self->targets ? self->targets->len : 0;
+
+  for (guint i = 0; i < num_targets; i++)
+    {
+      const gchar *target = g_ptr_array_index(self->targets, i);
+      if ((self->ignorecase ? strcasecmp(str, target) : strcmp(str, target)) == 0)
```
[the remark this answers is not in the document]
bshifter: that's true. 
originally utf8 wasn't in the scope, since non of our other string related function supporting utf8 atm. supporting utf8 could be useful, indeed.

lib/filterx/func-unset-empties.c
bshifter: what would be the proper way to handle that case?

lib/filterx/func-unset-empties.c
```diff
@@ -216,24 +280,154 @@ _extract_object_expr(FilterXFunctionArgs *args, GError **error)
...
+    {
+      g_assert(filterx_object_len(target, &len));
+      if (len == 0)
+        set_flag(&self->flags, FILTERX_FUNC_UNSET_EMPTIES_FLAG_REPLACE_EMPTY_LIST, TRUE);
+    }
+  else if (filterx_object_is_type(target, &FILTERX_TYPE_NAME(dict)))
+    {
+      g_assert(filterx_object_len(target, &len));
+      if (len == 0)
+        set_flag(&self->flags, FILTERX_FUNC_UNSET_EMPTIES_FLAG_REPLACE_EMPTY_DICT, TRUE);
```
alltilla, line 361: I think we should fail and throw an error to the user, if the len is not 0.

lib/filterx/func-unset-empties.c
```diff
@@ -216,24 +280,154 @@ _extract_object_expr(FilterXFunctionArgs *args, GError **error)
...
+    }
+  return TRUE;
+}
+
+static gboolean
+_target_iterator(FilterXExpr *key, FilterXExpr *value, gpointer user_data)
+{
+  FilterXFunctionUnsetEmpties *self = ((gpointer *) user_data)[0];
+  GError **error = ((gpointer *) user_data)[1];
+
```
alltilla, line 390: First let's check whether the expression itself is literal, because if it is a function or variable, we should fail.

lib/filterx/func-unset-empties.c
```diff
@@ -216,24 +280,154 @@ _extract_object_expr(FilterXFunctionArgs *args, GError **error)
...
+      g_set_error(error, FILTERX_FUNCTION_ERROR, FILTERX_FUNCTION_ERROR_CTOR_FAIL,
+                  FILTERX_FUNC_UNSET_EMPTIES_ARG_NAME_TARGETS" argument must be literal list. " FILTERX_FUNC_UNSET_EMPTIES_USAGE);
+      return FALSE;
+    }
+
+  guint64 num_targets = filterx_expr_literal_generator_len(targets);
+  if (num_targets > 0)
+    reset_flags(&self->flags, self->flags & (FILTERX_FUNC_UNSET_EMPTIES_FLAG_IGNORECASE |
+                                             FILTERX_FUNC_UNSET_EMPTIES_FLAG_RECURSIVE));
+  self->targets = g_ptr_array_new_full(num_targets, (GDestroyNotify)free);
```
alltilla, line 421: GLib allocated memory should be freed with `g_free`.

lib/filterx/func-unset-empties.c
```diff
@@ -46,29 +66,54 @@ static gboolean _process_list(FilterXFunctionUnsetEmpties *self, FilterXObject *
...
     }
 
-  if (filterx_object_is_type(obj, &FILTERX_TYPE_NAME(null)))
+  gsize str_len = 0;
+  const gchar *str = NULL;
+  gchar *casefold_str = NULL;
+  if (!filterx_object_extract_string(obj, &str, &str_len))
+    return FALSE;
+  if (!str)
+    return FALSE;
```
alltilla, line 89: Shouldn't this be an assert?

lib/filterx/func-unset-empties.c
```diff
@@ -46,29 +66,54 @@ static gboolean _process_list(FilterXFunctionUnsetEmpties *self, FilterXObject *
...
-      return str_len == 0 ||
-             strcasecmp(str, "n/a") == 0 ||
-             strcmp(str, "-") == 0;
+      guint64 len;
+      filterx_object_len(obj, &len);
+      return len == 0;
     }
 
-  if (filterx_object_is_type(obj, &FILTERX_TYPE_NAME(null)))
+  gsize str_len = 0;
```
alltilla, line 83: [optional]

The string part got quite large, it could go to a separate function for readability.

lib/filterx/func-flags.h
```diff
@@ -0,0 +1,72 @@
...
+ * COPYING for details.
+ *
+ */
+
+#ifndef FUNC_FLAGS_H
+#define FUNC_FLAGS_H
+
+#define MAX_ENUMS 64
+
+#define STATIC_ASSERT(COND, MSG) _Static_assert(COND, MSG)
```
MrAnno, line 30: Let's use `G_STATIC_ASSERT` instead, it's more compatible with older C versions.
(`syslog-ng.h` should be included instead of glib directly due to some glib compat headers we have)

lib/filterx/expr-literal-generator.c
```diff
@@ -302,3 +302,10 @@ filterx_expr_is_literal_list_generator(FilterXExpr *s)
   FilterXExprGenerator *generator = (FilterXExprGenerator *) s;
   return filterx_expr_is_generator(s) && generator->create_container == filterx_generator_create_list_container;
 }
+
+guint
+filterx_expr_literal_generator_len(FilterXExpr *s)
```
MrAnno, line 327: `g_list_length()` is `O(n)`, but we won't use this function at runtime (hopefully).

lib/filterx/func-unset-empties.c
```diff
@@ -216,24 +282,171 @@ _extract_object_expr(FilterXFunctionArgs *args, GError **error)
...
+    {
+      const gchar *str = filterx_string_get_value(target, &len);
+      if (len == 0)
+        set_flag(&self->flags, FILTERX_FUNC_UNSET_EMPTIES_FLAG_REPLACE_EMPTY_STRING, TRUE);
+      else
+        {
+          if (check_flag(self->flags, FILTERX_FUNC_UNSET_EMPTIES_FLAG_IGNORECASE))
+            g_ptr_array_add(self->targets, g_utf8_casefold(str, len));
+          else
+            g_ptr_array_add(self->targets, g_strdup(str));
```
MrAnno, line 385: Optional: reordering the conditions and using early return for the `TRUE` paths could reduce some "cyclomatic" complexity here.

MrAnno (approved):

alltilla (approved):
Cool feature :)

PR #277, opened 2024-09-06
lib/filterx: add startswith / endswith functions

OverOrion:
```
@version: 4.8

log {
  source {stdin(flags(no-parse));};
  filterx {
    declare needle = "bar";
    declare path = "";

    if (startswith($MSG, "foo", ignorecase=true)) {
      path = "starts with foo";
    }
    elif (startswith($MSG, needle)) {
      path = "starts with bar";
    }
    elif (endswith($MSG, "foo", ignorecase=true)) {
      path = "ends with foo";
    }
    elif (endswith($MSG, needle, ignorecase=true)) {
      path = "ends with bar";
    }
    else {path = "NOPE";};
  };
  filterx {vars();};
  destination { file("/dev/stdout"); };
};

```

### DIFF
--- a/lib/filterx/func-str.c
+++ b/lib/filterx/func-str.c
@@ -1,0 +1,244 @@
+/*
+ * Copyright (c) 2024 Axoflow
+ * Copyright (c) 2024 Szilard Parrag
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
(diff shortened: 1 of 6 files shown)


lib/filterx/func-str.c
```diff
@@ -0,0 +1,244 @@
...
+
+static void
+_filterx_function_startswith_free(FilterXExpr *s)
+{
+  FilterXFuncStartsWith *self = (FilterXFuncStartsWith *) s;
+
+  filterx_expr_unref(self->haystack);
+  if (self->needle.expr)
+    filterx_expr_unref(self->needle.expr);
+
```
alltilla, line 216: Who frees `self->needle.literal.str`?

lib/filterx/func-str.h
```diff
@@ -0,0 +1,51 @@
...
+} FilterXExprOrLiteral;
+
+typedef struct _FilterXFuncStartsWith
+{
+  FilterXFunction super;
+  FilterXExpr *haystack;
+  FilterXExprOrLiteral needle;
+  gsize needle_len;
+  gboolean ignore_case;
+} FilterXFuncStartsWith;
```
alltilla, line 47: These could be "private" == defined in the `.c` file.

PR #156, opened 2024-06-09
Proxy protocol moves tolog transport

bazsi:
This is now a complete LogTransport based implementation of the HAProxy protocol so it applies cleanly to any kind of LogProto implementation.

PR #278, opened 2024-09-09
transport: get rid of the TransportFactoryId abstraction and the associated registry

bazsi:
TransportFactoryId was used to encapsulate a "name" and a unique identifier for each transport type, from which we had two:
1) socket
2) tls

Since the switching between transports will only happen a couple of times within the lifetime of a connection, we don't need to optimize the lookup of the required transport to switch to.

This patch simply changes the TransportFactoryId with a string and the lookup by ID to a lookup by string. The string identifiers are encapsulated in macros, so misspelling on the call sites is not likely. And we only have one single call site.

By removing the TransportFactoryId and TransportFactoryRegistry I can easily get rid off 400 lines of complex, difficult to read code.


bazsi:
I've updated #156 to be based on top of this and finished off the conversion from LogProto to LogTransport when supporting the HAProxy protocol.

MrAnno (approved):

PR #282, opened 2024-09-13
filterx: `parse_windows_eventlog_xml()`

alltilla:
I tried to do this cleanly with minimal code duplication, and this is what I could achieve. I am still not completely satisfied with the result, but I don't see how could this be improved without making it less clean.

### DIFF
--- a/modules/xml/filterx-parse-windows-eventlog-xml.c
+++ b/modules/xml/filterx-parse-windows-eventlog-xml.c
@@ -1,0 +1,318 @@
+/*
+ * Copyright (c) 2024 Axoflow
+ * Copyright (c) 2024 Attila Szakacs <attila.szakacs@axoflow.com>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 as published
+ * by the Free Software Foundation, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "filterx-parse-windows-eventlog-xml.h"
+#include "filterx/object-string.h"
+#include "filterx/object-dict-interface.h"
+#include "scratch-buffers.h"
+
+static void _set_error(GError **error, const gchar *format, ...) G_GNUC_PRINTF(2, 0);
+static void
+_set_error(GError **error, const gchar *format, ...)
+{
+  if (!error)
+    return;
+
+  va_list va;
+  va_start(va, format);
+  *error = g_error_new_valist(g_quark_from_static_string("filterx-parse-windows-eventlog-xml"), 0, format, va);
+  va_end(va);
+}
+
+
+typedef enum FilterXParseWEVTPos_
+{
+  WEVT_POS_NONE,
+  WEVT_POS_EVENT,
+  WEVT_POS_EVENT_DATA,
+  WEVT_POS_DATA,
+} FilterXParseWEVTPos;
+
+typedef struct FilterXParseWEVTState_
+{
+  FilterXParseXmlState super;
+  FilterXParseWEVTPos position;
+  gboolean has_named_data;
+  GString *last_data_name;
+} FilterXParseWEVTState;
+
+static FilterXParseXmlState *
+_state_new(void)
+{
+  FilterXParseWEVTState *self = g_new0(FilterXParseWEVTState, 1);
+  filterx_parse_xml_state_init_instance(&self->super);
+  self->last_data_name = scratch_buffers_alloc();
+  self->position = WEVT_POS_NONE;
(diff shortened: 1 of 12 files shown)


modules/xml/filterx-parse-xml.c
```diff
@@ -97,50 +97,60 @@ _set_error(GError **error, const gchar *format, ...)
...
-} FilterXGeneratorFunctionParseXml;
+  self->xml_elem_context_stack = g_queue_new();
+  self->free_fn = filterx_parse_xml_user_data_free_method;
+}
+
+void
+filterx_parse_xml_user_data_free_method(FilterXParseXmlUserData *self)
+{
+  g_queue_free_full(self->xml_elem_context_stack, (GDestroyNotify) xml_elem_context_free);
+  g_free(self);
```
bazsi, line 143: I usually have the g_free() call in the wrapper that calls the free_fn() method, so the free_fn() can be optional.

modules/xml/filterx-parse-windows-eventlog-xml.c
```diff
@@ -0,0 +1,277 @@
...
+  va_end(va);
+}
+
+
+typedef struct FilterXParseWindowsEventlogXmlUserData_
+{
+  FilterXParseXmlUserData super;
+  gboolean in_event, in_event_data, in_data, has_named_data;
+  GString *last_data_name;
+} FilterXParseWindowsEventlogXmlUserData;
```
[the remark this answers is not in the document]
bazsi: Just to be constructive as well, FilterXParseXMLState could be a name for the parent and
FilterXParseWEVTState could be the other, and similarly name structs for the functions.

modules/xml/filterx-parse-windows-eventlog-xml.c
```diff
@@ -0,0 +1,277 @@
...
+}
+
+static void
+_end_elem(FilterXGeneratorFunctionParseXml *s,
+          GMarkupParseContext *context, const gchar *element_name,
+          FilterXParseXmlUserData *u, GError **error)
+{
+  FilterXParseWindowsEventlogXmlUserData *user_data = (FilterXParseWindowsEventlogXmlUserData *) u;
+
+  if (user_data->in_data && g_strcmp0(element_name, "Data") == 0)
```
bazsi, line 215: this list of items is pretty close to what we have on the opening tag (e.g. _store_relevant_position()), we could potentially use the same function, with an explicit value instead of a hard-coded TRUE or FALSE.

modules/xml/filterx-parse-windows-eventlog-xml.c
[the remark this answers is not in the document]
alltilla: Yes, you understand it right. There are things that is needed for `parse_xml()`, but `parse_windows_eventlog_xml()` needs some other fields on top of that.

The thing is that this user data is ugly by nature. You will see 4 different types of user_data passing/casting in the code. I did not want to have another user_data in `FilterXParseXmlUserData` for the descendant class to fill, because casting turned out to be a hell, so I did this object oriented approach, which seemed better.

I can rename it, its not a problem. I can abbreviate the `WindowsEventlog` parts in the new code with `WinEvent`.

modules/xml/filterx-parse-windows-eventlog-xml.c
alltilla: Yes, but you need to have a reverse order of checking on the closing part, at least I could only make it work that way. Maybe some refactor of this small part could make it work, I'll think about it.

modules/xml/filterx-parse-windows-eventlog-xml.c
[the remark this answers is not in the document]
alltilla: I have just seen your suggestions, I will rename accordingly.

modules/xml/filterx-parse-windows-eventlog-xml.c
bazsi: ah, ok. I didn't notice the different order. Also the starting part does not check that the booleans are false. Isn't that a problem?

alltilla: I'm rewriting this part with an enum and a very basic state machine, I think it is cleaner. That will fix the false checking problem.

modules/xml/filterx-parse-xml.c
```diff
@@ -97,50 +97,60 @@ _set_error(GError **error, const gchar *format, ...)
...
 
 
-typedef struct FilterXGeneratorFunctionParseXml_
+void
+filterx_parse_xml_user_data_init_instance(FilterXParseXmlUserData *self)
 {
-  FilterXGeneratorFunction super;
-  FilterXExpr *xml_expr;
-} FilterXGeneratorFunctionParseXml;
+  self->xml_elem_context_stack = g_queue_new();
```
bazsi, line 135: I was just wondering why this GQueue is not a GArray with the element type set to to XMLElemContext. That should get rid off a lot of allocations (the XMLElemContext itself and the associated GList structs that GQueue is using). We are only adding elemnts to the last position and taking them off from there, don't we?

modules/xml/filterx-parse-xml.c
alltilla: Thanks, fixed.

modules/xml/filterx-parse-windows-eventlog-xml.c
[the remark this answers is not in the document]
alltilla: Fixed.

modules/xml/filterx-parse-windows-eventlog-xml.c
alltilla: Fixed.

modules/xml/filterx-parse-xml.c
alltilla: Good idea, fixed.

alltilla:
@bazsi added the quick root node check as we discussed.

bazsi (approved):

PR #283, opened 2024-09-16
filterx: make 0 sec 0 usec datetime falsy

alltilla:
It is particularly useful when the otel log record does not have a timestamp set.

With this we can write expressions, like:

```
  date_str = log.time_unix_nano ?
             string(log.time_unix_nano) :
             string(log.observed_time_unix_nano);
```

### DIFF
--- a/lib/filterx/object-datetime.c
+++ b/lib/filterx/object-datetime.c
@@ -50,7 +50,8 @@ typedef struct _FilterXDateTime
 static gboolean
 _truthy(FilterXObject *s)
 {
-  return TRUE;
+  FilterXDateTime *self = (FilterXDateTime *) s;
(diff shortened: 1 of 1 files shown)


MrAnno (approved):

lib/filterx/object-datetime.c
```diff
@@ -50,7 +50,8 @@ typedef struct _FilterXDateTime
 static gboolean
 _truthy(FilterXObject *s)
 {
-  return TRUE;
+  FilterXDateTime *self = (FilterXDateTime *) s;
+  return self->ut.ut_sec != 0 || self->ut.ut_usec != 0;
```
MrAnno, line 54: Shouldn't it be falsy only when both are 0?

PR #284, opened 2024-09-16
filterx: support casting datetime to integer or double

alltilla:

### DIFF
--- a/lib/filterx/object-primitive.c
+++ b/lib/filterx/object-primitive.c
@@ -30,6 +30,7 @@
 #include "cfg.h"
 #include "filterx-globals.h"
 #include "str-utils.h"
+#include "timeutils/misc.h"
 
 #include "compat/json.h"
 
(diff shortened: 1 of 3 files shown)


lib/filterx/object-primitive.c
```diff
@@ -307,6 +308,10 @@ filterx_typecast_integer(FilterXExpr *s, GPtrArray *args)
         return filterx_integer_new(val);
     }
 
+  UnixTime ut;
+  if (filterx_object_extract_datetime(object, &ut))
+    return filterx_integer_new(ut.ut_sec * USEC_PER_SEC + ut.ut_usec);
```
bazsi, line 313: is this going to fit in a 64 bit integer?

bazsi: one ore question, I was thinking about increasing time resolution to nano seconds in the future, even in LogStamp, would that also fit? or we would lose some resolution in that case?

alltilla: Yes, it works with nanos, too for another ~240 years:

image

MrAnno, line 313: Do we need microsecond precision? What's our use-case here, is this needed to send timestamps to somewhere where usec epoch is the required format or do we just need it for internal reasons?

If the former is the case, we would probably need a conversion function supporting multiple epoch precisions.

bazsi: I was thinking more in the other direction, e.g. using nanoseconds precision instead.

The conversion to msec or sec is simple, just divide it by 1000 or 1000000

I'd merge this as is and then maybe do a followup based on the result of the discussion.

MrAnno (approved):

PR #295, opened 2024-09-18
logqueue-fifo: remove legacy flow-control+log-fifo-size() coupling

MrAnno:
I was investigating a log source hang, and it occurred to me in this context that it's time to part ways with the legacy force-suspend functionality.


bazsi (approved):
we probably should not have implemented the version compatibility at all.

I am happy to see this gone.

MrAnno:
In average configurations we have seen, the possible memory usage would have been doubled or tripled. We needed to avoid that scenario with this compatibility warning.

PR #297, opened 2024-09-19
lib/filterx: add startswith / endswith / includes functions

OverOrion:
Init time optimized version of https://github.com/axoflow/axosyslog/pull/277.
I opted for closing that because the changes are really different.

### DIFF
--- a/lib/filterx/func-str.c
+++ b/lib/filterx/func-str.c
@@ -391,3 +391,18 @@ error:
   filterx_expr_unref(&self->super.super);
   return NULL;
 }
+
+static gboolean
+_function_startswith_process(const gchar *haystack, gsize haystack_len, const gchar *needle, gsize needle_len)
+{
+  if (needle_len > haystack_len)
+    return FALSE;
+  return memcmp(haystack, needle, needle_len) == 0;
+}
+
+FilterXExpr *
+filterx_function_startswith_new(FilterXFunctionArgs *args, GError **error)
+{
+  return _function_affix_new(args, "startswith", _function_startswith_process,
+                             FILTERX_FUNC_STARTSWITH_USAGE, error);
+}
--- a/lib/filterx/func-str.c
+++ b/lib/filterx/func-str.c
@@ -39,6 +39,9 @@
 #define FILTERX_FUNC_ENDSWITH_USAGE "Usage: endswith(string, suffix, ignorecase=true)" \
 "or endswith(string, [suffix_1, suffix_2, ..], ignorecase=true)"
 
+#define FILTERX_FUNC_INCLUDES_USAGE "Usage: includes(string, substring, ignorecase=true)" \
+"or includes(string, [substring_1, substring_2, ..], ignorecase=true)"
+
 typedef struct _FilterXStringWithCache
 {
   FilterXExpr *expr;
@@ -422,5 +425,20 @@ FilterXExpr *
 filterx_function_endswith_new(FilterXFunctionArgs *args, GError **error)
 {
   return _function_affix_new(args, "endswith", _function_endswith_process, FILTERX_FUNC_ENDSWITH_USAGE,
+                             error);
+}
+
+static gboolean
+_function_includes_process(const gchar *haystack, gsize haystack_len, const gchar *needle, gsize needle_len)
+{
+  if (needle_len > haystack_len)
+    return FALSE;
+  return g_strstr_len(haystack, haystack_len, needle) != NULL;
+}
+
+FilterXExpr *
+filterx_function_includes_new(FilterXFunctionArgs *args, GError **error)
+{
+  return _function_affix_new(args, "includes", _function_includes_process, FILTERX_FUNC_INCLUDES_USAGE,
                              error);
 }
--- a/lib/filterx/func-str.c
+++ b/lib/filterx/func-str.c
@@ -36,6 +36,9 @@
 #define FILTERX_FUNC_STARTSWITH_USAGE "Usage: startswith(string, prefix, ignorecase=true)" \
 "or startswith(string, [prefix_1, prefix_2, ..], ignorecase=true)"
 
+#define FILTERX_FUNC_ENDSWITH_USAGE "Usage: endswith(string, suffix, ignorecase=true)" \
+"or endswith(string, [suffix_1, suffix_2, ..], ignorecase=true)"
+
 typedef struct _FilterXStringWithCache
 {
   FilterXExpr *expr;
@@ -405,4 +408,19 @@ filterx_function_startswith_new(FilterXFunctionArgs *args, GError **error)
 {
   return _function_affix_new(args, "startswith", _function_startswith_process,
                              FILTERX_FUNC_STARTSWITH_USAGE, error);
+}
+
+static gboolean
+_function_endswith_process(const gchar *haystack, gsize haystack_len, const gchar *needle, gsize needle_len)
+{
+  if (needle_len > haystack_len)
+    return FALSE;
+  return memcmp(haystack + haystack_len - needle_len, needle, needle_len) == 0;
+}
+
+FilterXExpr *
+filterx_function_endswith_new(FilterXFunctionArgs *args, GError **error)
+{
+  return _function_affix_new(args, "endswith", _function_endswith_process, FILTERX_FUNC_ENDSWITH_USAGE,
+                             error);
 }
--- a/lib/filterx/func-str.c
+++ b/lib/filterx/func-str.c
@@ -1,0 +1,393 @@
+/*
+ * Copyright (c) 2024 Axoflow
+ * Copyright (c) 2024 Szilard Parrag
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "filterx/func-str.h"
+#include "filterx/object-primitive.h"
+#include "filterx/expr-literal.h"
+#include "filterx/expr-literal-generator.h"
+#include "filterx/filterx-eval.h"
+#include "filterx/object-extractor.h"
+#include "filterx/object-string.h"
+#include "filterx/object-message-value.h"
+#include "filterx/object-list-interface.h"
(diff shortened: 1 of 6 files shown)


lib/filterx/func-str.c
```diff
@@ -0,0 +1,387 @@
...
+  if(*len > G_MAXSSIZE)
+    return FALSE;
+
+  *dst = g_utf8_casefold(src, (gssize) *len);
+  *len = g_utf8_strlen(*dst, -1);
+  return TRUE;
+}
+
+static gboolean
+_filterx_string_with_cache_eval_uncached(FilterXStringWithCache *self, gboolean ignore_case, const gchar **dst,
```
bazsi, line 88: can we move methods of FilterXStringWithCache closer to the struct so that these methods and those related to FilterXExprAffix interleave?

By keeping them separate, and lower level code coming first, circular dependencies can be avoided. Also it becomes a lot more clean what is what.

lib/filterx/func-str.c
```diff
@@ -0,0 +1,387 @@
...
+                                                gsize needle_len);
+typedef struct _FilterXExprAffix
+{
+  FilterXFunction super;
+  gboolean ignore_case;
+  FilterXStringWithCache *haystack;
+  struct
+  {
+    FilterXExpr *expr;
+    GPtrArray *needle_objects_with_cache;
```
bazsi, line 64: maybe we could drop the needle prefix from the name of the array? we already have that in the enclosing struct.

lib/filterx/func-str.c
```diff
@@ -0,0 +1,387 @@
...
+      *dst = self->borrowed_str_value;
+      *len = self->str_len;
+      return TRUE;
+    }
+
+  return _filterx_string_with_cache_eval_uncached(self, ignore_case, dst, len);
+}
+
+static gboolean
+_cache_string_literal(FilterXExpr *expr, const gchar *expression_name, gboolean ignore_case,
```
bazsi, line 143: I am finding the different prefixes to these functions difficult to read, as I am trying to find out why some of them has a _filterx_ prefix while others do not.

The many "cache" related functions and the FilterXStringWithCache struct makes it difficult to judge which class the specific method is related to.

As I see most of this function should be delegated to FilterXStringWithCache as it directly changes members of that struct.

lib/filterx/func-str.c
```diff
@@ -0,0 +1,512 @@
...
+  g_free(self);
+}
+
+static gboolean
+filterx_string_with_cache_do_casefold(FilterXStringWithCache *self)
+{
+  if(self->str_len > G_MAXSSIZE)
+    return FALSE;
+
+  self->owned_str_value = g_utf8_casefold(self->borrowed_str_value, (gssize) self->str_len);
```
OverOrion, line 109: The string comparisons might need `g_utf8_normalize()` as well besides case insensitive form.
I did not use it yet as I wanted a confirmation about the UTF8-ness of strings in this context.
@bazsi @alltilla  What do you think? :eyes:

OverOrion: The official docs say for the deprecated [`strncasecmp`](https://docs.gtk.org/glib/func.strncasecmp.html):
> [...] There are therefore two replacement techniques:
> [g_ascii_strncasecmp()](https://docs.gtk.org/glib/func.ascii_strncasecmp.html), which only works on ASCII and is not locale-sensitive,
> 
> and [**g_utf8_casefold**()](https://docs.gtk.org/glib/func.utf8_casefold.html) **followed by strcmp() on the resulting strings, which is good for case-insensitive sorting of UTF-8.**

lib/filterx/func-str.c
```diff
@@ -0,0 +1,459 @@
...
+
+static GPtrArray *
+_expr_affix_eval_needle(FilterXExprAffix *self)
+{
+  FilterXObject *needle_obj = filterx_expr_eval(self->needle.expr);
+  if (!needle_obj)
+    return NULL;
+  GPtrArray *needles = NULL;
+
+  if (filterx_object_is_type(needle_obj, &FILTERX_TYPE_NAME(string)))
```
alltilla, line 215: `filterx_expr_eval()` might return a `message_value` of `LM_VT_STRING`, which won't pass this condition, but should.

When you fix this, please extend the E2E testcase to cover this.

lib/filterx/func-str.c
```diff
@@ -0,0 +1,459 @@
...
+  struct
+  {
+    FilterXExpr *expr;
+    GPtrArray *cached_strings;
+  } needle;
+  FilterXExprAffixProcessFunc process;
+} FilterXExprAffix;
+
+static GString *
+_format_str_obj(FilterXObject *obj)
```
alltilla, line 67: All the format functions return a `GString *`, but your cache is a `gchar *` (or it is `const gchar *` right now, but should probably be a `gchar *`). This discrepancy will become obvious when you fix the `fill_cache()` method.

lib/filterx/func-str.c
```diff
@@ -0,0 +1,459 @@
...
+{
+  GString *str = scratch_buffers_alloc();
+
+  if(!filterx_object_repr(obj, str))
```
alltilla, line 71: I think we should be stricter and only accept `string` and `message_value` with `LM_VT_STRING`. I recommend using the `object_extractor` here.

lib/filterx/func-str.c
```diff
@@ -0,0 +1,459 @@
...
+    }
+  return TRUE;
+}
+
+static void
+_expr_affix_free(FilterXExpr *s)
+{
+  FilterXExprAffix *self = (FilterXExprAffix *) s;
+  g_ptr_array_free(self->needle.cached_strings, TRUE);
+
```
alltilla, line 203: `self->needle.expr` and `self->haystack` are leaked.

lib/filterx/func-str.c
```diff
@@ -0,0 +1,459 @@
...
+                  "needle must be set. %s", usage);
+      goto error;
+    }
+
+  filterx_function_init_instance(&self->super, affix_name);
+  self->ignore_case = ignore_case;
+  self->super.super.eval = _expr_affix_eval;
+  self->super.super.free_fn = _expr_affix_free;
+
+  self->needle.cached_strings = g_ptr_array_new();
```
alltilla, line 396: Missing:
```
g_ptr_array_set_free_func(self->needle.cached_strings, (GDestroyNotify) _string_with_cache_free);
````

File: lib/filterx/func-str.c [hunk 10]
```diff
@@ -0,0 +1,459 @@
...
+    {
+      GString *str_value = _obj_format(needle_obj, self->ignore_case);
+      filterx_object_unref(needle_obj);
+      if (!str_value)
+        return NULL;
+      needles = g_ptr_array_sized_new(1);
+      g_ptr_array_add(needles, str_value);
+      return needles;
+    }
+  else if (filterx_object_is_type(needle_obj, &FILTERX_TYPE_NAME(list)))
```
alltilla, line 225: `filterx_expr_eval()` might return a `message_value` of `LM_VT_JSON` or `LM_VT_LIST`, which won't pass this condition, but should.

When you fix this, please extend the E2E testcase to cover this.

lib/filterx/func-str.c
```diff
@@ -0,0 +1,459 @@
...
+
+  g_ptr_array_add(cached_strings, obj_with_cache);
+  return TRUE;
+}
+
+
+static gboolean
+_expr_affix_init_needle(FilterXExprAffix *self, FilterXExpr *needle)
+{
+  self->needle.expr = filterx_expr_ref(needle);
```
alltilla, line 181: You already have a reference to this expr, given by this call, no need to ref it again:
```c
FilterXExpr *haystack = filterx_function_args_get_expr(args, 0);
```

lib/filterx/func-str.c
```diff
@@ -0,0 +1,459 @@
...
+    return TRUE;
+
+  return !!_expr_format(self->expr, ignore_case);
+}
+
+static FilterXStringWithCache *
+_string_with_cache_new(FilterXExpr *expr, gboolean ignore_case)
+{
+  FilterXStringWithCache *self = g_new0(FilterXStringWithCache, 1);
+  self->expr = expr;
```
alltilla, line 135: Let's ref this here, and unref in the `free()` method (which is missing right now).

alltilla (commented):
Looks good!

There are a couple of things to resolve, then after another review round it will be probably merge ready :)

alltilla:
Sorry, there were some review comments stuck in from a previous session. Resolved everything that is already fixed.

lib/filterx/func-str.c
alltilla: Actually, I think we can leave this as is, it would be really strange if this came from a message...

lib/filterx/func-str.c
OverOrion: That was intentional, so that `FilterXStringWithCache::free` won't be able to call `free()` on `ScratchBuffers` backed memory. Should I have just kept it as a `GString`?

alltilla: filterxstringwithcache should own the cache if it is available, right? it only gets filled init time once.

lib/filterx/func-str.c
alltilla: This is still not fixed and tested properly. This config reproduces the problem:

```
log {
  source { example-msg-generator(num(1) template("fooBARbAz") values("msg_val" => "foo")); };
  filterx {
    result = json();
    var = $msg_val;
    result.startswith_foo = startswith($MSG, var);
  };
  destination { stdout(); };
};
```

lib/filterx/func-str.c
```diff
@@ -0,0 +1,473 @@
...
+          if (!str_value)
+            goto error;
+          g_ptr_array_add(needles, str_value);
+        }
+      filterx_object_unref(needle_obj);
+      return needles;
+    }
+
+error:
+  g_ptr_array_free(needles, TRUE);
```
alltilla, line 273: This is called upon a `NULL` pointer if none of the types match.

lib/filterx/func-str.c
```diff
@@ -0,0 +1,473 @@
...
+    {
+      GString *str_value = _obj_format(needle_obj, self->ignore_case);
+      filterx_object_unref(needle_obj);
+      if (!str_value)
+        return NULL;
+      needles = g_ptr_array_sized_new(1);
+      g_ptr_array_add(needles, str_value);
+      return needles;
+    }
+  else if (filterx_object_is_type(needle_obj, &FILTERX_TYPE_NAME(list)))
```
alltilla, line 246: [nitpick]

There is no need to `else if`, you early return above. An `if` is enough.

lib/filterx/func-str.c
```diff
@@ -0,0 +1,473 @@
...
+{
+  FilterXExpr *needle = filterx_function_args_get_expr(args, 1);
+  if (!needle)
+    {
+      g_set_error(error, FILTERX_FUNCTION_ERROR, FILTERX_FUNCTION_ERROR_CTOR_FAIL,
+                  "needle must be set. %s", function_usage);
+      return NULL;
+    }
+  return needle;
+}
```
alltilla, line 351: [nitpick]

These are guaranteed by the `FilterXFunctionArgs` class to succeed, because we checked the length before. I would recommend either removing the validations or changing them to assertions.

OverOrion: Merged them into a single `_extract_args` method to better align with the codebase.

lib/filterx/func-str.c
OverOrion: Thanks, this ought to take care of it now:
```
   log_msg_value = ${values.str};
    if (includes(log_msg_value, log_msg_value))
    {
        result.works_with_message_value = true;
    };
```

alltilla (approved):

PR #299, opened 2024-09-20
logreader: fix spurious wakeups by updating the suspended flag on I/O poll

MrAnno:


lib/logreader.c
```diff
@@ -335,6 +335,7 @@ log_reader_update_watches(LogReader *self)
   switch (prepare_action)
     {
     case LPPA_POLL_IO:
+      self->suspended = FALSE;
```
alltilla, line 338: [nitpick/question]

Although `LPPA_POLL_IO` was the only case where setting/unsetting `self->suspended` was missing, I _feel_ like the truthy state of `self->suspended` is not decided at this point, but earlier at `free_to_send`. Of course `ProtoServer` can override it with `LPPA_SUSPEND`, but as soon as `log_source_free_to_send()` returns true, we could set `self->suspended` to false. Am I understanding this right?

This is just a strange feeling though, I think everything is behaving as expected with your change.

alltilla (approved):

lib/logreader.c
MrAnno: `log_reader_update_watches()` is responsible for deciding about the suspended state of reader.
We could set the suspended flag to FALSE anywhere in this function, and `log_reader_suspend_until_awoken()` can reset to TRUE if needed, but adding an explicit suspended setting to each branch of the `switch (prepare_action)` looked the cleanest for me.

alltilla: Okay, thanks :)

PR #303, opened 2024-09-22
`file()`, `stdout()`: fix log sources getting stuck

MrAnno:
If a partial write occurs, the written messages must be acknowledged.

It is important to note that a message may be written partially.
In that case, the message will be acknowledged once the complete write is successful. If write errors occur during this process, the partially written message may be duplicated.

### DIFF
--- a/modules/affile/logproto-file-writer.c
+++ b/modules/affile/logproto-file-writer.c
@@ -122,6 +122,8 @@ _process_partial_write(LogProtoFileWriter *self, gsize written)
 
   self->partial_pos = 0;
   self->partial_messages = self->buf_count - first_non_written_chunk_index;
+
+  log_proto_client_msg_ack(&self->super, self->buf_count - self->partial_messages);
 }
 
 /*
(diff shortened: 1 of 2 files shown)


modules/affile/logproto-file-writer.c
```diff
@@ -45,6 +45,48 @@ typedef struct _LogProtoFileWriter
...
+    {
+      self->partial_pos += rc;
+      *status = LPS_PARTIAL;
+      return FALSE;
+    }
+
+  log_proto_client_msg_ack(&self->super, self->partial_messages);
+  g_free(self->partial);
+  self->partial = NULL;
+  self->partial_messages = 0;
```
alltilla, line 86: [optional]

As far as I can see, this was not part of the original code. I guess it is needed, but probably should be introduced in another commit.

alltilla (approved):
Awesome!

I had one small comment which you can ignore.

modules/affile/logproto-file-writer.c
MrAnno: It's unnecessary, it was just part of my thought process. I can either remove it or commit it separately.

alltilla: I am okay with either, or keeping it like this, too. Let me know if I should merge it this way :)

MrAnno: I've extracted it to a separate commit. Thanks.

alltilla (approved):

PR #308, opened 2024-09-25
light test: add EXTRA_ARGS processing

bshifter:
The make light-check command appends the EXTRA_ARGS environment variable to the command line arguments. Its primary purpose is to filter light tests during execution, but by not enforcing the -k argument, its use becomes more versatile.

example:
```make light-check EXTRA_ARGS="-k test_unset_empties"```


MrAnno (commented):
Will this work with other generators, for example, with `-G Ninja`?

bshifter:
> Will this work with other generators, for example, with `-G Ninja`?

Yes, it does.

MrAnno (approved):
Thanks :)

PR #309, opened 2024-09-25
Filterx invalid utf8 fix for unset_empties

bshifter:
- pre-validation of utf8 input string on eval
- targets literal generator list parse-time utf8 validation


alltilla (dismissed):
Shouldn't we also replace the compare function to be `g_utf8_collate` even for `ignorecase=false`? Thinking about the different ways an utf-8 string can be represented but have the same meaning.

alltilla:
Please rebase to main to drop the first commit.

alltilla:
Let's talk about UTF-8 IRL, this PR is good for now.

alltilla (approved):

alltilla (approved):

PR #312, opened 2024-09-26
cfg: fix config-version incosistencies

MrAnno:
Template functions, for example, are constructed right when AxoSyslog parses the configuration, so the `@version` directive must be placed at the top of the config file to make it consistent.

This patch fixes the implicit "current version" mechanism (when no `@version` is specified) by moving it back to the lexer, and warns the user about misplaced or redefined `@version` directives.

### DIFF
--- a/lib/cfg-lexer.c
+++ b/lib/cfg-lexer.c
@@ -26,6 +26,7 @@
 #include "cfg-lexer-subst.h"
 #include "cfg-block-generator.h"
 #include "cfg-grammar.h"
+#include "cfg.h"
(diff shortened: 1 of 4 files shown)


MrAnno:
Reproducer: Place a `template("$(format-json asd=asd)"` anywhere in the configuration without an explicit `@version` declaration or place the `@version` declaration after the config block containing the mentioned template. A compatibility warning will be produced and AxoSyslog will operate in compat mode.

OverOrion (approved):

PR #315, opened 2024-09-30
filterx: fix use after free in `strptime()` and macro caching

alltilla:

### DIFF
--- a/lib/filterx/object-datetime.c
+++ b/lib/filterx/object-datetime.c
@@ -255,10 +255,10 @@ _strptime_eval(FilterXExpr *s)
   const gchar *time_str;
   gsize time_str_len;
   gboolean extract_success = filterx_object_extract_string(time_str_obj, &time_str, &time_str_len);
-  filterx_object_unref(time_str_obj);
 
   if (!extract_success)
     {
+      filterx_object_unref(time_str_obj);
       filterx_eval_push_error("First argument must be string typed. " FILTERX_FUNC_STRPTIME_USAGE, s, NULL);
       return NULL;
     }
@@ -289,6 +289,7 @@ _strptime_eval(FilterXExpr *s)
   else
     result = filterx_null_new();
 
+  filterx_object_unref(time_str_obj);
   return result;
 }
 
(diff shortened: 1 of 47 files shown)


lib/filterx/object-datetime.c
```diff
@@ -255,10 +255,10 @@ _strptime_eval(FilterXExpr *s)
   const gchar *time_str;
   gsize time_str_len;
   gboolean extract_success = filterx_object_extract_string(time_str_obj, &time_str, &time_str_len);
-  filterx_object_unref(time_str_obj);
 
   if (!extract_success)
     {
+      filterx_object_unref(time_str_obj);
```
MrAnno, line 261: We have a lot of `filterx_object_extract_*()` functions, the naming for most of them seems okay, except for `string`, `protobuf`, and `bytes`. For those types, the extracted value is not really extracted but borrowed.
That may have been one of the reasons why this use-after-free was introduced.

I would rename these functions to denote this difference. If we insist on keeping the `extractor` idea in the name, adding `ref` may work too, for example, `filterx_object_extract_string_ref()`.

MrAnno (commented):
Nice catch!

lib/filterx/object-datetime.c
OverOrion: Just chiming in, how about `filterx_object_as_string()`?
(Inspired by [Rust conventions](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv))

alltilla: Good idea. Added `ref` to several getter functions that have this behavior.

alltilla: To me `filterx_object_as_string()` still does not communicate clearly that it is a borrowed reference to the underlying string. Somehow we should incorporate `ref` to the function name.

alltilla: Oh my bad the "as" word communicates it.

MrAnno (approved):

Issue #317, opened 2024-09-30
`s3()` destination high memory usage caused by reload


alltilla:
# AxoSyslog
## Version of AxoSyslog
4.8.1

# Issue
## Failure
The memory usage of the `syslog-ng` process increases with every `syslog-ng-ctl reload` indefinitely.
It seems like that this is not a memory leak, but the `botocore` python library is caching some session related data.

The following python snippet also reproduces something similar, but the memory usage stops after 5-10 "reloads":
```py
from boto3 import Session, client

while True:
    input(": ")

    # This does not "leak"
    # c = client("s3")
    # c.close()

    s = Session()
    c = s.client("s3")
    c.close()
```

The caching happens around the `Session` object, if we create the client directly, the memory usage problem does not happen.

This might be related: https://github.com/boto/boto3/issues/3614

PR #318, opened 2024-10-01
s3: workaround on-reload memleak when no role is set

orymate:
AxoSyslog leaks an amount of memory every time it's reloaded if an s3
destination is configured (#317). As a quick workaround we partially revert the
patch introducing the use of `boto3.Session` when no `role()` is set.


alltilla (commented):
Thanks for the fix! :)

Please `git commit --amend -s`. Otherwise approve.

pepov:
A comment would probably be worth to explain the scenario when a session could leak memory in case of frequent reloads.

alltilla (approved):

PR #324, opened 2024-10-05
Filterx parse cef/leef #1

bshifter:
Introduces the `parse_cef` and `parse_leef` functions to the filterx module. 
Currently, `parse_leef` supports only LEEF version 1.0. Support for LEEF 2.0 will be included in a future update.

examples:
```
filterx {
   leef = parse_leef(leef_message);
   cef = parse_cef(cef_message);
}
```


modules/cef/event-format-parser.c
```diff
@@ -0,0 +1,323 @@
...
+  csv_scanner_init(&csv_scanner, &self->csv_opts, log);
+
+  guint64 i = 0;
+  while (csv_scanner_scan_next(&csv_scanner))
+    {
+      if (i >= num_fields)
+        break;
+
+      const gchar *input = csv_scanner_get_current_value(&csv_scanner);
+      APPEND_ZERO(input, input, csv_scanner_get_current_value_len(&csv_scanner));
```
alltilla, line 178: Why don't we pass the `len` down the call chain? We could save a lot of `memcpy`s caused by `APPEND_ZERO` and some `strlen`s down the line.

modules/cef/event-format-parser.c
```diff
@@ -0,0 +1,323 @@
...
+  FilterXObject *dict_val = filterx_string_new(value, value_len);
+
+  gboolean ok = filterx_object_set_subscript(out, dict_key, &dict_val);
+
+  filterx_object_unref(dict_key);
+  filterx_object_unref(dict_val);
+  return ok;
+}
+
+gboolean unescape_value_separators(KVScanner *self)
```
alltilla, line 96: [optional]

This could be private. Also please move the function name to a new line for better `grep`-ability.

modules/cef/filterx-func-parse-cef.c
```diff
@@ -0,0 +1,94 @@
...
+#include "parser/parser-expr.h"
+#include "scratch-buffers.h"
+#include "str-utils.h"
+#include "scanner/kv-scanner/kv-scanner.h"
+#include "event-format-parser.h"
+
+static Field cef_fields[] =
+{
+  { .name = "version", .field_parser = parse_version},
+  { .name = "deviceVendor"},
```
alltilla, line 51: [optional]

We usually use `snake_case` (see the otel filterx parts). I think it would be better to stay consistent.

modules/cef/event-format-parser.c
```diff
@@ -0,0 +1,323 @@
...
+    return FALSE;
+
+  return TRUE;
+}
+
+static FilterXObject *
+_create_container(FilterXExprGenerator *s, FilterXExpr *fillable_parent)
+{
+  return filterx_generator_create_dict_container(s, fillable_parent);
+}
```
alltilla, line 280: [nitpick]

You can set this directly:
```c
  self->super.super.create_container = filterx_generator_create_dict_container;
```

alltilla (commented):
Really solid PR!

alltilla:
I only had 2 really minor comments, otherwise approve. Very nice PR :)

modules/cef/event-format-parser.c
```diff
@@ -0,0 +1,311 @@
...
+
+  filterx_object_unref(dict_key);
+  filterx_object_unref(dict_val);
+  return ok;
+}
+
+static gboolean
+_unescape_value_separators(KVScanner *self)
+{
+  gchar escaped_separator[3] = {'\\', self->value_separator};
```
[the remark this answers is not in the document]
alltilla: oh, it seems like having 3 here was a functional thing by accident.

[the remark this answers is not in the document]
bshifter: it was a null-terminated string originally. tried to remove tailing zero, but forgot to decrease array size. it was 0 by the circumastances.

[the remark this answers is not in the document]
alltilla: I completelty missed, that the 3rd arg is a string, and you are preparing that.

alltilla (approved):

alltilla (approved):

PR #329, opened 2024-10-07
Minor FilterX cleanups

OverOrion:
- **lib/filterx: refactor _pull_variable_from_message**
- **lib/filterx: remove obsolete filterx_eval_sync_scope_and_message() function declaration**

### DIFF
--- a/lib/filterx/expr-variable.c
+++ b/lib/filterx/expr-variable.c
@@ -48,14 +48,10 @@ _pull_variable_from_message(FilterXVariableExpr *self, FilterXEvalContext *conte
       return NULL;
     }
 
-  FilterXObject *msg_ref;
   if (log_msg_is_value_from_macro(value))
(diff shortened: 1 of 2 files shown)


lib/filterx/expr-variable.c
```diff
@@ -84,7 +80,14 @@ _eval(FilterXExpr *s)
     }
 
   if (!filterx_variable_handle_is_floating(self->handle))
-    return _pull_variable_from_message(self, context, context->msgs[0]);
+    {
+      FilterXObject *msg_ref = _pull_variable_from_message(self, context, context->msgs[0]);
+      if(!msg_ref)
+        goto error;
```
alltilla, line 86: I think we should directly return with `NULL` here, to retain the eval error message set in `_pull_variable_from_message()`. Jumping to `error` rewrites it.

alltilla (approved):

PR #334, opened 2024-10-11
otel, loki: support dynamic `headers()`

alltilla:

### DIFF
--- a/modules/grpc/loki/loki-dest.cpp
+++ b/modules/grpc/loki/loki-dest.cpp
@@ -93,9 +93,14 @@ DestinationDriver::init()
     }
 
   if (log_template_is_literal_string(worker_partition_key))
-    log_template_unref(worker_partition_key);
+    {
+      log_template_unref(worker_partition_key);
+    }
   else
-    log_threaded_dest_driver_set_worker_partition_key_ref(&this->super->super.super.super, worker_partition_key);
+    {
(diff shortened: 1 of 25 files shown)


bazsi:
I did not test this explicitly, but went through the commits and they all looked fine to me.

bazsi (approved):

modules/grpc/loki/loki-dest.cpp
```diff
@@ -93,9 +93,14 @@ DestinationDriver::init()
...
   if (log_template_is_literal_string(worker_partition_key))
-    log_template_unref(worker_partition_key);
+    {
+      log_template_unref(worker_partition_key);
+    }
   else
-    log_threaded_dest_driver_set_worker_partition_key_ref(&this->super->super.super.super, worker_partition_key);
+    {
+      log_threaded_dest_driver_set_worker_partition_key_ref(&this->super->super.super.super, worker_partition_key);
+      log_threaded_dest_driver_set_flush_on_worker_key_change(&this->super->super.super.super, TRUE);
```
MrAnno: I remember that this option was deliberately not set for Loki.
I've validated your thoughts on the labels, labels are set once per batch, so it seems it is the right thing to do to flush batches. I don't remember the original reason for not setting this though.

MrAnno (approved):

PR #336, opened 2024-10-11
configure.ac: add options to override thread local storage

bazsi:


bazsi:
this may not solve the problem, so don't merge it yet.

PR #337, opened 2024-10-11
Fix std::transform and std::tolower usage in GRPCDest

OverOrion:
`std::tolower` is not an addressible function, the intended way to use it is within a lambda.

See the following for more information:
https://devblogs.microsoft.com/oldnewthing/20241007-00/?p=110345
https://en.cppreference.com/w/cpp/string/byte/tolower


alltilla (approved):
Nice catch! Thanks

PR #338, opened 2024-10-12
grpc: factor out schema related functionality from bigquery to common

alltilla:
This will be useful for other drivers that want to support setting the message schema from config.

Depends on #334


alltilla:
rebased to main

OverOrion (approved):
LGTM, thank you!

PR #327, opened 2024-10-06
Syslog ng ctl attach debugger

bazsi:
This is a continuation of #326 to add interactive debugger functionality over syslog-ng-ctl

PR #339, opened 2024-10-13
Debugger interrupt handling

bazsi:
This is a continuation o #327 to add interrupt handling to the debugger. At the moment it only works on the original console, syslog-ng-ctl attach debugger does not propagate ctrl+c signals yet.


alltilla:
Rebased to main and pushed 2 commits that fix the `help` text.

alltilla (dismissed):

alltilla (approved):

PR #340, opened 2024-10-13
Debugger step and follow commands

bazsi:
The last of my debugger improvements that follows #339

### DIFF
--- a/lib/debugger/debugger.c
+++ b/lib/debugger/debugger.c
@@ -29,6 +29,7 @@
 #include "mainloop.h"
 #include "timeutils/misc.h"
 #include "compat/time.h"
+#include "scratch-buffers.h"
 
 #include <stdio.h>
 #include <unistd.h>
@@ -44,6 +45,7 @@ struct _Debugger
   LogPipe *current_pipe;
   gboolean drop_current_message;
   struct timespec last_trace_event;
+  GThread *interactive_thread;
 };
 
 static gboolean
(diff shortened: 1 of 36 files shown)


lib/logpipe.h
```diff
@@ -74,6 +74,8 @@
 /* sync filterx state to message in right before calling queue() */
 #define PIF_SYNC_FILTERX_TO_MSG      0x0400
 
+#define PIF_BREAKPOINT        0x0400
```
MrAnno: This is an already used value, I'll fix it.

lib/debugger/debugger.c
```diff
@@ -165,8 +168,10 @@ _cmd_help(Debugger *self, gint argc, gchar *argv[])
              "  info, i                  Display information about the current execution state\n"
              "  list, l                  Display source code at the current location\n"
              "  continue, c              Continue until the next breakpoint\n"
+             "  step, s                  Single step\n"
+             "  follow, f                Follow this message, ignoring any other breakpoints\n"
              "  display                  Set the displayed message template\n"
-             "  trace, t                 Display timing information as the message traverses the config\n"
+             "  trace, t                 Trace this message along the configuration\n"
```
MrAnno, line 174: `breakpoint_site` tracking is somehow incorrect at the beginning, because I don't see these commands after the first message arrives and we stop at the first breakpoint.

PR #349, opened 2024-10-21
Fix mainloop hang in case job submission fails

bazsi:


alltilla:
@bazsi
I have extracted the last commit and rebased it to main.
Your previous version can be found here, if we want to revert to it: https://github.com/alltilla/axosyslog/tree/bazsi-fix-mainloop-hang-in-case-job-submission-fails

lib/logreader.c
```diff
@@ -573,7 +573,11 @@ log_reader_io_handle_in(gpointer s)
   log_reader_disable_watches(self);
   if ((self->options->flags & LR_THREADED))
     {
-      main_loop_io_worker_job_submit(&self->io_job, NULL);
+      if (!main_loop_io_worker_job_submit(&self->io_job, NULL))
```
MrAnno, line 576: This can only fail if `main_loop_workers_quit`, but in that case, we should not retry.

If my assumption is correct, this patch depends on work on a different branch.

alltilla: That is possible. Talked IRL with Bazsi, and concluded that this PR is meaningful on its own, but I might have misunderstood something. Sorry if so.

MrAnno: We talked about this IRL, registering a task in the debugger would hopefully resolve the issue in a cleaner way.

bazsi:
we could not reproduce the problem and clearly this is not the right solution.

PR #351, opened 2024-10-21
Add exit on eof flag to logreader

bazsi:


MrAnno (approved):

MrAnno:
(We'll need to add a news entry, the flag seems useful in general)

PR #354, opened 2024-10-22
grpc: add clickhouse dest

alltilla:
TODO:
 - [ ] newsfile entry
 - [ ] PR description

Depends on #338


alltilla:
rebased to main and fixed the copyright check

modules/grpc/clickhouse/clickhouse-dest.cpp
```diff
@@ -52,6 +52,8 @@ DestDriver::init()
       return false;
     }
 
+  this->query = "INSERT INTO " + this->database + "." + this->table + " FORMAT Protobuf";
```
alltilla, line 56: We might need escaping here.

modules/grpc/clickhouse/clickhouse-dest-worker.cpp
```diff
@@ -24,25 +24,188 @@
...
+  query_info.set_user_name(owner_->get_user());
+  query_info.set_password(owner_->get_password());
+  query_info.set_query(owner_->get_query());
+  query_info.set_input_data(this->query_data.str());
+}
+
+static LogThreadedResult
+_map_grpc_status_to_log_threaded_result(const ::grpc::Status &status)
+{
+  // TODO: this is based on OTLP, we should check how the ClickHouse gRPC server behaves
```
alltilla, line 117: Fix this todo.

modules/grpc/clickhouse/clickhouse-dest.cpp
alltilla: Note: I checked and unfortunately the ClickHouse gRPC server does not support query params with the provided proto, so we cannot use this: https://clickhouse.com/docs/en/sql-reference/syntax#defining-and-using-query-parameters

The best we can do is escaping with backtick or double quote: https://clickhouse.com/docs/en/sql-reference/syntax#identifiers

Additionally we could move the database name from the query to the request's database field, so we only escape the table part.

alltilla: FYI: https://github.com/ClickHouse/ClickHouse/pull/71080

alltilla: Done.

OverOrion (commented):
Overall LGTM, nice!

modules/grpc/clickhouse/clickhouse-dest-worker.cpp
```diff
@@ -36,6 +36,104 @@ DestWorker::DestWorker(GrpcDestWorker *s)
...
+    default:
+      g_assert_not_reached();
+    }
+
+temporary_error:
+  msg_debug("ClickHouse server responded with a temporary error status code, retrying after time-reopen() seconds",
+            evt_tag_int("error_code", status.error_code()),
+            evt_tag_str("error_message", status.error_message().c_str()),
+            evt_tag_str("error_details", status.error_details().c_str()));
+>>>>>>> Stashed changes
```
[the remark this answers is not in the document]
alltilla: thanks, fixed!

OverOrion (approved):
LGTM, nice! :rocket:

modules/grpc/clickhouse/clickhouse-dest-worker.cpp
OverOrion: We agreed on merging as is, so this can be closed.

PR #360, opened 2024-10-31
Filterx: scope-sync followups for cloning

OverOrion:
- **lib/filterx: fix scope cloning**
- **lib/filterx: fix filterx_scope_foreach_variable**

### DIFF
--- a/lib/filterx/filterx-scope.c
+++ b/lib/filterx/filterx-scope.c
@@ -316,7 +316,7 @@ filterx_scope_clone(FilterXScope *other)
   if (other->variables->len > 0)
     self->dirty = other->dirty;
   self->syncable = other->syncable;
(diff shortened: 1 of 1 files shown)


MrAnno:
We've discussed this interactively: keeping the old value of the generation counter during clones may reach the point of overflow.

Alternatives:
- use `0` to mark the variable stale (`msg has newer value`), and start the actual generation counter from `1`.
- just remove the variable as @OverOrion originally did it before I started to think about potentially unnecessary optimizations 😄

OverOrion:
First I tried going with the first alternative, but it didn't work and I couldn't find where it was failing, so I went with the second alternative in the end.
Should a performance problem arise we can revisit this.

MrAnno (approved):

PR #361, opened 2024-11-02
[HAPROXY] use $SOURCEIP instead of $PROXIED_SRCIP

bazsi:
This is a large refactor of the HAProxy support, and preparations for protocol auto detection.  It also changes the HAProxy support to use the standard source/destination addresses in LogMessage, instead of a proxy specific values e.g. $SOURCEIP instead of $PROXIED_SRCIP.

Short summary of the patches:
* small refactors and cleanups (first 5 patches)
* simplify TLS compression setup code, previously this crossed a lot of layers, whereas it can be a lot simpler
* LogTransportTLS is becoming an adapter, e.g. instead of going directly to file descriptors, go through another layer of the LogTransport interface. This prepares for protocol auto detection.
* introduces LogTransportStack level aux data, making it possible to enrich messages from the any of the Transports 
* adds the $SOURCEPORT macro
* LogTransportHAProxy is changed so it invokes a log_transport_stack_switch() instead of doing this internally
* has a big refactor of TransportMapperInet, as its various settings and options were becoming too difficult to read. This patch also adds a large light testcase to cover all possible transport() cases
* contains an alternative fix for #482 (although pretty similar)
* and few smaller patches


bazsi:
This is now ready for review and I think would be a better solution than #482 with the same fix.

bazsi:
@OverOrion @MrAnno could you re-review this, so we can merge this instead of #482?

MrAnno:
I'm reviewing it.

modules/afsocket/transport-mapper-inet.c
```diff
@@ -544,6 +559,22 @@ transport_mapper_syslog_apply_transport(TransportMapper *s, GlobalConfig *cfg)
       self->delegate_tls_start_to_logproto = TRUE;
       self->super.transport_name = g_strdup("rfc5425+proxied-tls-passthrough");
     }
+  else if (strcasecmp(transport, "auto") == 0)
+    {
+      /* NOTE: this is temporary until TLS auto detection does not exist.
+       * In this case transport(auto) + tls() configuration means we want to
+       * start TLS first and then do the framing auto-detection.  If we have
+       * TLS auto-detection this entire block can be removed, as the "auto"
+       * proto would detect TLS and start it if needed.
```
MrAnno, line 569: TLS auto-detection sounds like a security issue for me. TLS provides encryption and authentication.
If we want to allow non-TLS and TLS traffic on the same port, we should notify the user with a big warning that neither encryption nor authentication is guaranteed. We could probably do this when `peer-verify(off)` is set, but even then we should warn the user that encryption is also optional so vulnerable to "downgrade attack".

MrAnno:
I'm reviewing it.

lib/transport/transport-aux-data.h
```diff
@@ -81,6 +81,16 @@ log_transport_aux_data_copy(LogTransportAuxData *dst, LogTransportAuxData *src)
     }
 }
 
+static inline void
+log_transport_aux_data_move(LogTransportAuxData *dst, LogTransportAuxData *src)
+{
+  gsize data_to_copy = sizeof(*src) - sizeof(src->data) + src->end_ptr;
+
+  if (dst)
+    memcpy(dst, src, data_to_copy);
```
OverOrion, line 90: This feels weird here as `LogTransportAuxData` is owned by the `LogTransportStack`. Am I missing something here?

modules/afsocket/transport-mapper-inet.h
```diff
@@ -32,26 +32,21 @@ typedef struct _TransportMapperInet
...
   const gchar *server_port_change_warning;
-  guint32 flags;
-  gboolean require_tls;
-  gboolean allow_tls;
-  gboolean require_tls_when_has_tls_context;
+  /* tls() options are required */
+  gboolean require_tls_configuration;
+  /* tls() options are optional, but are permitted */
+  gboolean allow_tls_configuration;
+  gboolean delegate_tls_start_to_logproto;
```
OverOrion, line 39: Could you please move the comments `the proto can start TLS but we won't do that ourselves` and `it is up to the transport plugin to start TLS` here?

modules/afsocket/transport-mapper-inet.c
```diff
@@ -511,32 +538,54 @@ transport_mapper_syslog_apply_transport(TransportMapper *s, GlobalConfig *cfg)
     }
   else if (strcasecmp(transport, "proxied-tls") == 0)
     {
+      /* SSL HAProxy -> SSL syslog */
```
OverOrion, line 541: Thanks for these explanations, they are handy! :rocket:

lib/transport/transport-tls.c
```diff
@@ -24,23 +24,121 @@
...
+      *written_bytes = 0;
+      if (errno == EAGAIN)
+        BIO_set_retry_write(bio);
+      return -1;
+    }
+  *written_bytes = ret;
+  return 1;
+}
+
+int
```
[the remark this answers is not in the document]
bazsi: done

lib/transport/transport-aux-data.h
bazsi: This function moves the responsibility from one LogTransportAuxData instance to another, e.g. it copies all
data & references while dropping it from the old one.

It is similar to log_transport_aux_copy() but also drops references from the old one. This is called when we are moving all of our state from one LogTransportStack to another. This happens when we replace the LogProto implementation on the fly.

Any why we move, instead of just having a reference, that's easy too: I did not want to add an extra dereference a pointer to LogTransportStack every time we need to do a read or write in a LogProto instance.

modules/afsocket/transport-mapper-inet.h
bazsi: I added a more elaborate comment but left the originals there as well.

lib/transport/transport-aux-data.h
bazsi: I hope this clears this up. I am resolving this conversation now, please reopen if you disagree.

lib/transport/transport-tls.c
```diff
@@ -24,23 +24,120 @@
...
+
+BIO_METHOD *
+BIO_s_transport(void)
+{
+  static BIO_METHOD *meth = NULL;
+
+  if (meth)
+    return meth;
+
+  meth = BIO_meth_new(BIO_TYPE_NONE, "LogTransportBIO");
```
MrAnno, line 123: This seems to be leaked. Shouldn't we call BIO_meth_free() through our apphooks before shutdown?

lib/transport/transport-tls.c
```diff
@@ -24,23 +24,120 @@
...
+  return meth;
+}
+
+BIO *
+BIO_transport_new(LogTransportTLS *transport)
+{
+  BIO *bio = BIO_new(BIO_s_transport());
+
+  g_assert(transport != NULL);
+  BIO_set_data(bio, transport);
```
MrAnno, line 137: Shouldn't we call `BIO_set_nbio()`?

MrAnno (commented):
I've started manual-testing the generic- and TLS-related transport changes (I think we need thorough testing since this touches core functionality).

MrAnno:
I've tested the TLS-related changes on a real network connection through multiple hops, it works as expected.

lib/transport/transport-tls.c
bazsi: added a transport-globals module and a global_init/deinit which in turn allocates and deallocates this through
apphook.

lib/transport/transport-tls.c
bazsi: As I grepped the openssl code, this would only call the BIO control function (which in our case does nothing).

If we were using the normal OpenSSL provided socket wrappers, BIO_set_nbio()  will set the O_NONBLOCK flag, which we already set outside of this constructor.

bazsi:
@MrAnno I've resolved the memory leak and did not add BIO_set_nbio() call and resolved your comments. Let me know if you agree. Thanks

MrAnno (approved):

PR #363, opened 2024-11-04
logsource: fix use after free in message processing debug log

alltilla:
`log_pipe_forward_msg()` frees `msg`.


MrAnno:
Do you have a reproducer for this PR?

As far as I see, `log_pipe_forward_msg()` should never free the message.
In most cases, the messages are placed in a queue, so no free happens in the forward method.

Even when the message is dropped (filtered or no destination exists), the actual free will happen later in `log_msg_refcache_stop()`.

But apart from its occurrence in reality, the fix LGTM.

alltilla:
I'm not at the computer, but an example msg generator source, and an always failing filter is enough to reproduce this, no destination needed.

MrAnno:
Oh, okay. The example-message-generator does not use the refcache.

MrAnno (approved):

alltilla:
Oh, that explains it :)

PR #364, opened 2024-11-05
Centralized internal logging

OverOrion:


lib/filter/filter-call.c
```diff
@@ -50,9 +50,10 @@ filter_call_eval(FilterExprNode *s, LogMessage **msgs, gint num_msg, LogTemplate
   else
     stats_counter_inc(self->super.not_matched);
 
+  msg_set_context(msgs[num_msg - 1]);
   msg_trace("filter() evaluation started",
-            evt_tag_str("called-rule", self->rule),
-            evt_tag_msg_reference(msgs[num_msg - 1]));
+            evt_tag_str("called-rule", self->rule));
+  msg_set_context(NULL);
```
bazsi, line 56: This will remove the reference to the current message, is that really a good idea? the next filter would stop printing the rcptid this way.

I don't think either msg_set_context( ) calls are needed here.

bazsi: I don't think we need to set the context this many times. Every case, the current rcptid will be the same as there's only one message being handled at any given time.

Since we are not printing the pointer anymore, I don't think any of the msg_set_context() calls are needed in this patch. Let's discuss.

bazsi (changes requested):
discuss msg_set_context() calls before merging

lib/filter/filter-call.c
OverOrion: Attila and I discussed this IRL, here is the summarized version:

1. `msg_set_context()` calls should be placed at places where new `LogMessage` instances are encountered. It's constructor seemed a viable option at first, however that would introduce a layering violation so we opted for moving it elsewhere.
2. During our internal queuing where `pop()` calls happen is the moment in time where logs are encountered before being sent to destinations. This combined with setting context at insertions (reading from source) seemed like the best place for them.
3. `LogScheduler` has it's very own (and unique) way of fetching logs, that's why it required a separate patch (easy to miss it) .
4. We could not move it inside `log_source_post` as there are already logs printed before the call actually happens.

modules/grpc/otel/otel-source-services.hpp
```diff
@@ -123,6 +123,7 @@ syslogng::grpc::otel::TraceServiceCall::Proceed(bool ok)
                 }
 
               LogMessage *msg = log_msg_new_empty();
+              msg_set_context(msg);
```
MrAnno, line 126: We want to track `the message we are currently working on`. This idea will always imply a certain amount of layering violation, but placing this call into LogMessage constructors wouldn't necessarily help as I'm not sure we want to track each message that pops into existence.

For example, what do we want to do with synthetic messages produced by the `grouping-by()` and PatternDB parsers?

lib/logqueue-fifo.c
```diff
@@ -489,6 +489,7 @@ log_queue_fifo_pop_head(LogQueue *s, LogPathOptions *path_options)
   if (!node->flow_control_requested)
     self->backlog_queue.non_flow_controlled_len++;
 
+  msg_set_context(msg);
```
MrAnno, line 492: Why don't you place this in the log queue abstract pop method? That way, the memory queues and disk buffers won't have to be careful about this.

modules/examples/sources/threaded-diskq-source/threaded-diskq-source.c
```diff
@@ -151,6 +151,7 @@ _fetch(LogThreadedFetcherDriver *s)
 
   gint64 remaining_messages = log_queue_get_length(self->queue);
   LogMessage *msg = log_queue_pop_head(self->queue, &local_options);
+  msg_set_context(msg);
```
MrAnno, line 154: This can be removed once the log queue abstract class does its job.

modules/grpc/otel/otel-source-services.hpp
MrAnno: Once this question is answered, I think we can introduce a function that can help reduce these set_context calls at least on the source side.

OverOrion:
We've discussed this with @bazsi IRL and came to the conclusion that this might not worth the effort (considering that this does not cause any problems), so closing this as is, but can be revisited if needed.

PR #365, opened 2024-11-06
filterx: `update_metrics()` and `flatten()` perf improvements

alltilla:

### DIFF
--- a/lib/filterx/object-metrics-labels.c
+++ b/lib/filterx/object-metrics-labels.c
@@ -21,125 +21,255 @@
  * COPYING for details.
  *
  */
+
 #include "object-metrics-labels.h"
+#include "object-dict-interface.h"
+#include "object-string.h"
+#include "object-extractor.h"
+#include "filterx-object-istype.h"
+#include "filterx-ref.h"
+#include "stats/stats-prometheus.h"
+#include "stats/stats-cluster.h"
+#include "scratch-buffers.h"
+
+#include <string.h>
+
+typedef struct _FilterXObjectMetricsLabels
+{
+  FilterXDict super;
+  GArray *labels;
+  gboolean sorted;
+} FilterXObjectMetricsLabels;
 
 static gboolean
 _truthy(FilterXObject *s)
 {
-  // TODO: implement
-  return FALSE;
+  return TRUE;
 }
 
 static gboolean
 _marshal(FilterXObject *s, GString *repr, LogMessageValueType *t)
 {
-  // TODO: implement
-  return FALSE;
-}
+  FilterXObjectMetricsLabels *self = (FilterXObjectMetricsLabels *) s;
 
-static gboolean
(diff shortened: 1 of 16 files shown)


lib/filterx/object-metrics-labels.c
```diff
@@ -216,6 +222,39 @@ _free(FilterXObject *s)
...
+
+  gpointer k, v;
+  while (g_hash_table_iter_next(&iter, &k, &v))
+    {
+      StatsClusterLabel *label = v;
+      g_array_append_val(new_labels, *label);
+    }
+
+  g_array_free(typed_self->labels, TRUE);
+  typed_self->labels = new_labels;
```
MrAnno, line 251: This looks like a `rw` operation here.

lib/filterx/object-metrics-labels.c
```diff
@@ -0,0 +1,352 @@
...
+  ScratchBuffersMarker marker;
+  scratch_buffers_mark(&marker);
+
+  gsize name_len, value_len;
+  const gchar *name_str = _obj_to_string(key, &name_len);
+  const gchar *value_str = _obj_to_string(*new_value, &value_len);
+  if (!name_str || !value_str)
+    return FALSE;
+
+  FilterXObject *typed_new_value = filterx_ref_unwrap_ro(*new_value);
```
MrAnno, line 131: unwrap is not necessary here (string is immutable).

alltilla: This might not be a string, this is why I introduced the check. Dicts can be passed here as well, after a `repr()` call we will store it, and return the newly allocated repr version of the dict.

MrAnno: As long as your the type you compare against is immutable, you don't need the unwrap call.

lib/stats/stats-prometheus.c
```diff
@@ -301,6 +301,23 @@ stats_format_prometheus(StatsCluster *sc, gint type, StatsCounterItem *counter,
...
+  gsize orig_size = buf->len;
+
+  g_string_append_c(buf, '{');
+
+  gboolean comma_needed = FALSE;
+  for (gsize i = 0; i < labels_len; ++i)
+    _append_formatted_label(buf, &labels[i], &comma_needed);
+
+  if (buf->len == orig_size + 1)
+    g_string_truncate(buf, orig_size);
```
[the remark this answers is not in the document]
alltilla: Good idea, fixed.

lib/filterx/object-metrics-labels.c
alltilla: Nice catch, fixed!

lib/filterx/object-metrics-labels.c
alltilla: You are right, I overlooked that. Fixed :)

MrAnno (approved):

PR #366, opened 2024-11-06
CI: Use AWS ECR location for trivy DBs

OverOrion:


MrAnno (approved):
https://github.com/aquasecurity/trivy-db/pull/440

PR #367, opened 2024-11-07
add date heuristics of wallclocktime to strptime in filterx

bshifter:
closes #2460

### DIFF
--- a/lib/filterx/object-datetime.c
+++ b/lib/filterx/object-datetime.c
@@ -282,6 +282,8 @@ _strptime_eval(FilterXExpr *s)
         break;
     }
 
+  wall_clock_time_guess_missing_fields(&wct);
+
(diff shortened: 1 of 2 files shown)


bazsi (approved):

lib/filterx/object-datetime.c
```diff
@@ -282,6 +282,8 @@ _strptime_eval(FilterXExpr *s)
         break;
     }
 
+  wall_clock_time_guess_missing_fields(&wct);
+
```
alltilla, line 286: It's unnecessary to guess the missing fields if we will return null because of a parsing error. This could go to the `if (end) { ... }` block below.

alltilla (commented):
One small comment, otherwise approve

PR #371, opened 2024-11-08
news: add missing 4.9.0 news entries

alltilla:

### DIFF
--- a/news/create-newsfile.py
+++ b/news/create-newsfile.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python3
 #############################################################################
+# Copyright (c) 2024 Axoflow
+# Copyright (c) 2024 Attila Szakacs <attila.szakacs@axoflow.com>
 # Copyright (c) 2020 Balabit
 #
 # This program is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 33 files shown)


news/create-newsfile.py
```diff
@@ -41,6 +43,7 @@
 blocks = [
     ('Features', 'feature-*.md'),
     ('Bugfixes', 'bugfix-*.md'),
+    ('FilterX', 'filterx-*.md'),
```
MrAnno, line 46: I agree that FilterX may require a separate paragraph in the NEWS file (at least until a lot of changes keep coming into it), but introducing it as a block next to "features, bugfixes, packaging, others" can be problematic, because those categories apply to FilterX too: we add features there, we fix bugs, etc.

Allowing the `fx-` prefix, like `fx-feature`, `fx-bugfix` might be a resolution for this concern, or we could just keep writing the entries as usual, but with the `filterx:` prefix in the title.

Or if we just want to separate new FilterX features, but not the bugfixes and other minor changes, the above could work as well.

MrAnno (commented):
Thanks for collecting all the FilterX changes, it must have been a huge work.

news/create-newsfile.py
alltilla: Good point, I agree.

I have added two sections: `FilterX features` and `FilterX bugfixes`. `FilterX other changes` might be another candidate for some entries, but we might be okay with putting those to `FilterX features`.

PR #375, opened 2024-11-11
Version: 4.9.0

MrAnno:
Tarball is available at: https://github.com/axoflow/axosyslog/actions/runs/11777580075


alltilla:
@MrAnno I cannot add you as a reviewer, as you are the one who opened the PR.

I updated the news file, could you kindly check it? Thanks!

MrAnno:
Sure.

Gabo said GitHub Enterprise permissions should be enough to use the github-actions bot for both the release draft and for this PR. If that's not enough, we'll need a service account in the long term.

NEWS.md
```diff
@@ -1,62 +1,399 @@
...
+    # takes the stdio fds for 10 seconds and displays syslog-ng output in that time period
+    $ syslog-ng-ctl attach stdio --seconds 10
+    ```
+    ```
+    # steal trace level log messages for 10 seconds
+    $ syslog-ng-ctl attach logs --seconds 10 --log-level trace
+    ```
+    ([#326](https://github.com/axoflow/axosyslog/pull/326))
 
-Check out the [AxoSyslog documentation](https://axoflow.com/docs/axosyslog-core/) for all the details.
```
MrAnno, line 15: I wouldn't remove this intro section.

MrAnno (commented):
I think we should mention in the intro section that FilterX features/bugfixes can be found in the news file and introduce the feature, for example, by linking this blog post:
https://axoflow.com/filterx-introduction/

It may be worth mentioning that the entire FilterX language is under heavy development, and we may break the API and potentially the syntax as well in the next few releases (for example, how we handle dicts/array and JSON/OTel records are about to change).

LGTM otherwise.

NEWS.md
alltilla: Thanks, fixed both your comments.

MrAnno:
✅ 😄 

Thanks.

alltilla (approved):
✅ [alltilla](https://github.com/alltilla) approved these changes 😄

PR #379, opened 2024-11-13
`app-transform()`: add `include-transforms()` and `exclude-transforms()` options

alltilla:
Example usage:

`syslog-ng.conf`:
```
transformation "app-1"[default] {
  transform[transform-1] {
    step["transform-1-step-1"] {
      # transform-1-step-1 implementation
    };
    step["transform-1-step-2"] {
      # transform-1-step-2 implementation
    };
  };

  transform[transform-2] {
    step["transform-2-step-1"] {
      # transform-2-step-1 implementation
    };
    step["transform-2-step-2"] {
      # transform-2-step-2 implementation
    };
  };

  transform[transform-3] {
    step["transform-3-step-1"] {
      # transform-3-step-1 implementation
    };
    step["transform-3-step-2"] {
      # transform-3-step-2 implementation
    };
  };
};

log {
  source { example-msg-generator(); };
  parser {
    app-transform(filterx-app-variable(meta.app) include-transforms(transform_1, transform_2) exclude-transforms(transform_1));
  };
};
```

```
$ syslog-ng -Fe --preprocess-into=-
Setting current version as config version; version='4.9'
transformation "app-1"[default] {
  transform[transform-1] {
    step["transform-1-step-1"] {
      #
    };
    step["transform-1-step-2"] {
      #
    };
  };

  transform[transform-2] {
    step["transform-2-step-1"] {
      #
    };
    step["transform-2-step-2"] {
      #
    };
  };

  transform[transform-3] {
    step["transform-3-step-1"] {
      #
    };
    step["transform-3-step-2"] {
      #
    };
  };
};

log {
  source { example-msg-generator(); };
  parser {
     , 
#Start Block parser generator app-transform
## exclude_transforms=transform_1
## include_transforms=transform_1, transform_2
## filterx_app_variable=meta.app
## app-transform(flavour(default))
channel {
    filterx {

#Start Application app-1
    if (meta.app == 'app-1') {
        # step: transform-2-step-1        
      # transform-2-step-1 implementation
    
        # step: transform-2-step-2        
      # transform-2-step-2 implementation
    
    };

#End Application app-1
    };
}
#End Block parser generator app-transform
;
  };
};
```

### DIFF
--- a/lib/cfg-parser.c
+++ b/lib/cfg-parser.c
@@ -378,3 +378,43 @@ cfg_process_yesno(const gchar *yesno)
     return TRUE;
   return FALSE;
 }
+
+gboolean
+cfg_process_list_of_literals(const gchar *input, GList **result)
+{
+  if (strchr(input, '"') || strchr(input, '\''))
+    {
+      fprintf(stderr, "Error: list of literals must not contain single or double quotes: %s\n", input);
+      return FALSE;
+    }
+
+  *result = NULL;
+
+  gchar *input_copy = g_strdup(input);
+  const gchar delimiters[] = ", \t\n";
+  const gchar *elem = strtok(input_copy, delimiters);
+  while (elem)
+    {
+      if (strlen(elem) > 0)
+        *result = g_list_append(*result, __normalize_key(elem));
+      elem = strtok(NULL, delimiters);
+    }
+
+  g_free(input_copy);
+  return TRUE;
+}
+
+static gint
+_compare_literals_normalized(gconstpointer a, gconstpointer b)
+{
+  gchar *normalized_b = __normalize_key(b);
+  gint result = strcmp(a, normalized_b);
+  g_free(normalized_b);
+  return result;
+}
+
+gboolean
+cfg_is_literal_in_list_of_literals(GList *list, const gchar *literal)
+{
+  return !!g_list_find_custom(list, literal, _compare_literals_normalized);
+}
--- a/modules/appmodel/app-transform-generator.c
+++ b/modules/appmodel/app-transform-generator.c
@@ -101,11 +101,11 @@ _generate_app_transform(Transformation *transformation, gpointer user_data)
   g_string_append_printf(self->block, "\n#Start Application %s\n", transformation->super.name);
   g_string_append_printf(self->block,     "    if (%s == '%s') {\n", self->filterx_app_variable,
                          transformation->super.name);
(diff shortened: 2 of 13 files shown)


modules/appmodel/app-transform-generator.c
```diff
@@ -133,7 +133,7 @@ app_transform_generator_new(gint context, const gchar *name)
   AppTransformGenerator *self = g_new0(AppTransformGenerator, 1);
 
   app_object_generator_init_instance(&self->super, context, name);
-  self->super.parse_arguments = app_transform_generator_parse_arguments;
-  self->super.generate_config = app_transform_generate_config;
+  self->super.parse_arguments = _parse_arguments;
+  self->super.generate_config = _generate_config;
```
MrAnno, line 192: Optional: I think if one chooses to use long private method names, we should leave them as is.
The short names in our codebase have already started to collide with each other, which makes debugging a bit less pleasant.

lib/cfg-parser.c
```diff
@@ -378,3 +378,43 @@ cfg_process_yesno(const gchar *yesno)
...
+  const gchar delimiters[] = ", \t\n";
+  const gchar *elem = strtok(input_copy, delimiters);
+  while (elem)
+    {
+      if (strlen(elem) > 0)
+        *result = g_list_append(*result, __normalize_key(elem));
+      elem = strtok(NULL, delimiters);
+    }
+
+  g_free(input_copy);
```
MrAnno, line 403: Wouldn't `g_strsplit()` help here (both the usage and its return value)?

alltilla: The problem is we need to normalize the substring. If we used g_strsplit, we would have to reallocate every element.

MrAnno: I overlooked the normalization part.

modules/appmodel/app-transform-generator.c
alltilla: I'm okay with either. We can keep already named functions. I had to add a free method, and it seemed strange to see long and short function names close to each other.

alltilla: Dropped the commit.

MrAnno: I've decided that I personally will use long names when the chance of collision is higher (for example, `_init()`).

Mixing short and long names may look strange, but those common short names cause some minor headache in gdb.

MrAnno (approved):

PR #380, opened 2024-11-14
check program name for rfc3164

bshifter:
The check-program flag was added to validate the program field in syslog format. It verifies that the program field contains only the allowed characters '[a-zA-Z0-9].-_/()' and requires at least one alphabetical character in the program name. If these conditions are not met, it treats the program field and the remaining fields as part of the log message. This approach avoids using regular expressions to improve performance.

examples:
```
log {
    source { tcp(port(2000) flags(check-hostname, check-program, dont-store-legacy-msghdr)); };
...
```

### DIFF
--- a/news/feature-380.md
+++ b/news/feature-380.md
@@ -1,0 +1,14 @@
+`check-program`: Introduced as a flag for global or source options.
+
+By default, this flag is set to false. Enabling the check-program flag triggers `program` name validation for `RFC3164` messages. Valid `program` names must adhere to the following criteria:
+
+Contain only these characters: `[a-zA-Z0-9-_/().]`
+Include at least one alphabetical character.
+If a `program` name fails validation, it will be considered part of the log message.
+
+
+Example:
+
+```
+source { network(flags(check-hostname, check-program)); };
+```


lib/cfg-grammar.y
```diff
@@ -1485,6 +1487,7 @@ source_reader_option
         /* NOTE: plugins need to set "last_reader_options" in order to incorporate this rule in their grammar */
 
 	: KW_CHECK_HOSTNAME '(' yesno ')'	{ last_reader_options->check_hostname = $3; }
+  | KW_CHECK_PROGRAM '(' yesno ')' { last_reader_options->check_program = $3; }
```
MrAnno, line 1490: The corresponding setting is missing from `log_reader_options_init()`.

(Setting `check-program(yes)` has no effect.)

lib/logmsg/logmsg.h
```diff
@@ -135,6 +135,7 @@ enum
   /* message field missing */
   LM_T_SYSLOG_MISSING_MESSAGE,
   LM_T_PREDEFINED_MAX,
+  LM_T_SYSLOG_RFC_3164_INVALID_PROGRAM,
```
MrAnno, line 138: This should be moved one line above.

MrAnno: Can we also insert a new line before `LM_T_PREDEFINED_MAX` macro so that we won't miss this next time?

Like it is done with `LM_V_PREDEFINED_MAX`.

modules/syslogformat/syslog-format.c
```diff
@@ -326,6 +326,29 @@ _syslog_format_parse_version(LogMessage *msg, const guchar **data, gint *length)
   return TRUE;
 }
 
+#define PROGRAM_NAME_ALLOWED_SPECIALS ".-_()/"
+#define PROGRAM_NAME_ALLOWED_SPECIALS_LEN 6
```
MrAnno, line 330: Another idea without macros and magic numbers:

```
static const gchar program_name_allowed_specials[] = ".-_()/";
static gsize program_name_allowed_spacial_chars_len = G_N_ELEMENTS(program_name_allowed_specials);
```

modules/syslogformat/syslog-format.c
```diff
@@ -340,6 +363,16 @@ _syslog_format_parse_legacy_program_name(LogMessage *msg, const guchar **data, g
...
     }
+
+  if ((flags & LP_CHECK_PROGRAM))
+    {
+      if (!_validate_program(prog_start, (src - prog_start)))
+        {
+          log_msg_set_tag_by_id(msg, LM_T_SYSLOG_RFC_3164_INVALID_PROGRAM);
+          return;
+        }
+    }
```
MrAnno, line 374: This is one of the most performance-critical paths of AxoSyslog.
For this reason, can we reimplement `LP_CHECK_PROGRAM` similarly to how `LP_CHECK_HOSTNAME` is done?
(Avoiding the double iteration by moving the character validation inside the while loop above.)

MrAnno (commented):
LGTM otherwise. 

(We may also need a NEWS file entry).

modules/syslogformat/syslog-format.c
```diff
@@ -336,10 +354,24 @@ _syslog_format_parse_legacy_program_name(LogMessage *msg, const guchar **data, g
...
+          log_msg_set_tag_by_id(msg, LM_T_SYSLOG_RFC_3164_INVALID_PROGRAM);
+          return;
+        }
       _skip_char(&src, &left);
     }
+
+  if (G_UNLIKELY(flags & LP_CHECK_PROGRAM) && !has_alpha_char)
+    {
+      log_msg_set_tag_by_id(msg, LM_T_SYSLOG_RFC_3164_INVALID_PROGRAM);
+      return;
```
MrAnno, line 372: We can't early return from here, as we need to optionally set the legacy message header:

```
if ((flags & LP_STORE_LEGACY_MSGHDR))
...
```

(In case of an invalid program name, the header should not contain the program name)

MrAnno:
LGTM. We can merge it after adding a news entry.

MrAnno (approved):

MrAnno (approved):

PR #385, opened 2024-11-19
filterx: native dict/list types

MrAnno:


bazsi:
rebased against main.

bazsi:
I rebased this to current main and followed up the change of prototype of built-in simple functions.

bazsi:
This branch helps a lot in malloc() efficiency and json cloning, but the lookups take 2x as much than the JSON lookups.

With the JSON based build, get_subscript() takes 7.5% of the same run where with this branch (and the changing of the configuration from json() to dict()), it takes 14.4%

I was running a lot of tests, so this might not be apples-to-apples completely, but that would be the reason for not improving performance _that much_.

So I would definitely merge this, it is so much better on all other aspects than the JSON based build, than I am inclined to ditch that (and protobuf) completely and use just one internal/optimized representation and convert on input/output.

MrAnno:
I'm moving out this from draft as we discussed that this will be merged and reworked.

bazsi (approved):

PR #387, opened 2024-11-22
ci: fix macOS dep conflict

MrAnno:
pkgconf is a newer, actively maintained implementation of pkg-config.

This is to fix CI on certain macOS runner instances (not all of them are affected). pkg-config is now migrated to pkgconf in brew, but a few runners have old pkg-config installed in a way they can't be overwritten by just a simple upgrade.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Install dependencies
         run: |
           brew update --preinstall
-          brew bundle --file=contrib/Brewfile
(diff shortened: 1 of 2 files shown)


OverOrion (approved):
LGTM, thank you!

FTR: They did the same switch here as well [Homebrew/homebrew-core@`eacd290` (#197590)](https://github.com/Homebrew/homebrew-core/pull/197590/commits/eacd290a0f18d804e13bd440d551fdca6d1e19e9)

PR #394, opened 2024-11-27
Filterx regexp_subst supports match groups

bshifter:
Enhanced regexp_subst to support match group identifiers in patterns.

Since this feature could have a significant impact on performance, it is
controlled by a flag called 'groups', which is set to false by default.
To enable this feature, add the optional groups=true argument to your regexp_subst call.

example:
```
filterx{
  $MSG.on = regexp_subst("25-02-2022", /(\d{2})-(\d{2})-(\d{4})/, "\\3/\\2/\\1", groups=true);
  # returns "2022\/02\/25"
}
```
This feature could have been implemented using the pcre2_substitute() function provided by the PCRE2 library. However, this approach was not viable because pcre2_substitute() does not support zero-length matches, which are a critical aspect of syslog-ng's legacy regex matcher.

As a result, the implementation introduced unwanted/unnecessary code and increased complexity. Given this, I would recommend discontinuing support for zero-length matches in filterx regexps. This change would improve both performance and code maintainability. While zero-length matches may seem like a useful feature in regular expressions, their practical use cases are likely to be rare and non-critical.


alltilla (approved):
Very nice!

This could have a news entry, but it is enough to do in another PR.

PR #397, opened 2024-11-28
filterx: `set_fields()`

alltilla:
A recurring pattern in FilterX is to take a dict and set multiple fields in it with overrides or defaults.

`set_fields()` takes a dict as the first argument and `overrides` and `defaults` as optional parameters.

`overrides` and `defaults` are also dicts, where the key is the field's name, and the value is either an expression, or
a list of expressions. If a list is provided, each expression will be evaluated, and the first successful, non-`null` one will
be used to set the respective field's value. This is similar to chaining null-coalescing (`??`) operators, but is more performant.

`overrides` are always processed for each field. The `defaults` for a field are only processed, if the field does not already have a value set.

Example usage:
```
set_fields(
  my_dict,
  overrides={
    "foo": [invalid_expr, "foo_override"],
    "baz": "baz_override",
    "almafa": [invalid_expr_1, null],  # No effect
  },
  defaults={
    "foo": [invalid_expr, "foo_default"],
    "bar": "bar_default",
    "almafa": "almafa_default",
    "kortefa": [invalid_expr_1, null],  # No effect
    }
);
```

### DIFF
--- a/lib/filterx/func-set-fields.c
+++ b/lib/filterx/func-set-fields.c
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2024 Axoflow
+ * Copyright (c) 2024 Attila Szakacs <attila.szakacs@axoflow.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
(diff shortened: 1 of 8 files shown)


lib/filterx/func-set-fields.c
```diff
@@ -127,11 +129,89 @@ typedef struct FilterXFunctionSetFields_
...
+      if (filterx_object_extract_null(value_obj))
+        {
+          filterx_object_unref(value_obj);
+          continue;
+        }
+
+      FilterXObject *value_obj_cloned = filterx_object_clone(value_obj);
+      filterx_object_unref(value_obj);
+
+      if (!filterx_object_set_subscript(dict_unwrapped, key, &value_obj_cloned))
```
MrAnno, line 159: You are using the FilterXObject interface on `dict_unwrapped` everywhere, so you don't actually need `filterx_ref_unwrap_rw()`.

A readonly local unwrap is enough to check the type and report errors early, everything else will work transparently (and we would avoid a copy as well in case someone somehow manages to produce a set_fields() call that wouldn't modify the dict :D)

alltilla: Ah, I see! Thanks, updated.

MrAnno (approved):

PR #398, opened 2024-11-28
FilterX expression eval metrics

MrAnno:
Depends on #395


alltilla (approved):

MrAnno:
That was the original idea, but I was really missing the expression type from the metric names (operator, function, etc.), but I didn't want to store that as a string in the expression, so I ended up with this boilerplate.
Maybe next time.

alltilla (approved):

Issue #392, opened 2024-11-25
FR: improvements on `regexp_search`


furiel:
Relates to https://github.com/axoflow/axoflow/issues/3016

I would like to use `regexp_search` to parse out fields of a message similarly to:

`
log.body = json(regexp_search($MESSAGE, '^(?<remote>\S*) (?<host>\S*) (?<user>\S*) \[(?<time>[^\]]*)\] "(?<method>\S+)(?: +(?<path>[^\"]*?)(?:\s+\S*)?)?" (?<code>\S*) (?<size>\S*)(?: "(?<referer>[^\"]*)" "(?<agent>[^\"]*)"(?:\s+(?<http_x_forwarded_for>\S+))?)?'));
`

In my case, the pattern may be a user input, so it can contain unnamed capturing group too.

---

### Forced return type

If none of the capturing groups have name, the type of the return value changes from object to array, which makes handling the match with a generic code a bit tricky: instead of `match["1"]`, I need to use `match[1]`. It would make implementation easier if I can force the return value to be an object independently from the pattern.

### Drop `match[0]` automatically

 A frequent use case is to match the entire log line. However, I already have that in another variable, I do not want to store it in the message, so a typical pattern looks like match + drop first element of the return value. It would be nice if these were done automatically, considering I may need to drop 0 or "0" depending on the return value.

We may need another way to check if a regexp_search has matched after that: as checking the length may not be enough.

furiel:
Fixed with #399

PR #399, opened 2024-11-29
Filterx regexp_search rework

bshifter:
continuation of #394 
based on https://github.com/axoflow/axosyslog/issues/392

Introduced two new optional flags: keep_zero and list_mode.
The result type no longer switches between dict and list based on the presence of named groups.
The default result type is now dict.
list_mode can force a list result, either via the flag or the type of the fillable (the fillable's type takes precedence).
Match group zero is now excluded by default unless no other groups are present. This behavior can be overridden using the keep_zero flag.

Additionally, the FilterXRematchState structure, shared across other regexp functions, has been updated with a generic flags field. This allows functions to pass custom options to the state without incurring additional overhead.


alltilla (dismissed):
Nice feature!

Please rebase so we can merge it.
This also needs a news entry, but can be in another PR

PR #407, opened 2024-12-03
Fix handshake regression

bazsi:
This is an alternative solution to #388 

* this fixes a regression caused by #155 , reported by @HofiOne  
* This simplifies the I/O tracking for files, LogProto/LogTransport should get full authority on whether data is already available before poll and also how we need to poll it
* Instead of the ugly "immediate_check" mechanism, we wake up LogReader if there's any event available

The regression happens in the LogReader: #155 causes us to skip a poll iteration because of the handshake (even if it does not exist). This causes a problem as at the very first iteration it is only "immediate_check" that causes us to start fetching messages. And that only fires once. By skipping that first iteration, we stall and stop fetching messages.

Instead of reverting the handshake related change  (which should be correct in any async environment), I dropped the immediate_check mechanism (instead of a single internal user that could probably be eliminated) and implemented checking for I/O state in the usual polling paths. Which means: we want to know if there are unprocessed bytes in the file (e.g. poll-file-events), and if there are no bytes there, we want to know if our buffer contains an entire line.

I tried to limit the scope of this change as much as possible.

### DIFF
--- a/lib/logproto/logproto-text-server.c
+++ b/lib/logproto/logproto-text-server.c
@@ -291,6 +291,33 @@ log_proto_text_server_set_multi_line(LogProtoServer *s, MultiLineLogic *multi_li
   self->multi_line = multi_line;
 }
 
+static gboolean
+log_proto_text_server_restart_with_state(LogProtoServer *s, PersistState *persist_state, const gchar *persist_name)
+{
+  LogProtoTextServer *self = (LogProtoTextServer *) s;
+
+  gboolean res = log_proto_buffered_server_restart_with_state(s, persist_state, persist_name);
+  if (!res)
+    return FALSE;
+
+  if (!self->super.buffer)
+    return FALSE;
+
+  LogProtoBufferedServerState *state = log_proto_buffered_server_get_state(&self->super);
+  const guchar *buffer_start = self->super.buffer + state->pending_buffer_pos;
+  gsize buffer_bytes = state->pending_buffer_end - state->pending_buffer_pos;
+
+  if (buffer_bytes > 0)
+    {
+      const guchar *eom = self->find_eom(buffer_start, buffer_bytes);
+      if (eom)
+        self->cached_eol_pos = eom - self->super.buffer;
+    }
+  log_proto_buffered_server_put_state(&self->super);
+
+  return TRUE;
+}
+
 void
 log_proto_text_server_free(LogProtoServer *s)
 {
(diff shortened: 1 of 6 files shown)


lib/logproto/logproto-text-server.c
```diff
@@ -291,6 +291,30 @@ log_proto_text_server_set_multi_line(LogProtoServer *s, MultiLineLogic *multi_li
...
+
+  LogProtoBufferedServerState *state = log_proto_buffered_server_get_state(&self->super);
+  const guchar *buffer_start = self->super.buffer + state->pending_buffer_pos;
+  gsize buffer_bytes = state->pending_buffer_end - state->pending_buffer_pos;
+
+  if (buffer_bytes > 0)
+    {
+      const guchar *eom = self->find_eom(buffer_start, buffer_bytes);
+      if (eom)
+        self->cached_eol_pos = eom - self->super.buffer;
```
MrAnno, line 311: Don't we need to adjust `cached_eol_pos` based on what `log_proto_text_server_accumulate_line()` does?

bazsi: I think we don't but maybe it would be easier to review this IRL or in a call.

My understanding is that cached_eol_pos contains the next "unprocessed" NL character that needs to be accumulated into a multi-line buffer.

Also, cached_eol_pos != 0 means that we don't need to poll, we can immediately try to extract a line from the buffer (see the prepare() method).

If restart_with_state() find the next EOL in our buffer, then:
1) prepare() would return TRUE, indicating that we can go ahead and start extracting the line text
2) fetch_from_buffer() will use this as the first line and continue with accumulating the next line
3) we would loop as long as we have newline characters in the buffer (or if the line is complete).

With that I think this works correctly. But I'll do some more testing.

MrAnno:
@HofiOne This fixes the restart mechanism without reverting the handshake simplification.

I reviewed the changes and everything seems to be fine, but the text-server/buffered-server implementation is complex and it's relatively easy to miss small details during review.

Would it be possible to run your internal tests on this PR to double-check for regressions (if you have plans to backport this change, of course).

HofiOne:
> @HofiOne This fixes the restart mechanism without reverting the handshake simplification.
> 
> I reviewed the changes and everything seems to be fine, but the text-server/buffered-server implementation is complex and it's relatively easy to miss small details during review.
> 
> Would it be possible to run your internal tests on this PR to double-check for regressions (if you have plans to backport this change, of course)?

The issue came up during the backport attempt of the simplification, and I'd like to continue it, but
- currently, I cannot work on these unfortunately
- this is the first really problematic point of the now divergent code bases
- e.g. the text-server is completely removed as I can remember during the proxy refactoring, which got a different implementation in AxoSyslog as well, meanwhile, our wildcard-file fixes also affected these code parts

I cannot see now how to resolve this best. I'll surely get back to this, but as far as I can see, it will not be in the near future.

Thanks for the signal!

MrAnno (approved):

MrAnno:
Okay, we move forward then. Thanks.

PR #409, opened 2024-12-05
Filterx regexp_subst match group changes

bshifter:
continuation of #405
- substitution match groups feature is enabled by default
- the match group control flag argument `groups` inverted and renamed to `nogroups`
- match group indexes are now supported up to 999 instead of 9 (multi digits)
- checking replacement pattern at construct time to disable match group feature if it's unnecessary


alltilla:
Really cool, thanks for this!

> the match group control flag argument groups inverted and renamed to nogroups

I might have miscommunicated this, sorry if so.
Can we keep it `groups`, but make it default to `true`? The fact that we set it to false by default if the replacement does not have a group, is not relevant for the user, it is just a performance optimization, and does not have any functional impact.
I think introducing negation to the option names usually causes confusion.

PR #416, opened 2024-12-10
http: add HTTP response text to the diagnostic messages

bazsi:
This patch adds the response text returned by the HTTP server to various diagnostic messages, so issues with the server can be diagnosed more easily.

```
@version: current

options { keep-hostname(yes); };

log {
	source { tcp(port(2000)); };

	destination {
		http(
			#url("https://httpbin.org/status/502")
			url("https://httpbin.org/html")
			content-compression(gzip)
			workers(1)
			tls(
				ssl_version(tlsv1_2)
				peer-verify(no)
			)
			time-reopen(1)
			body("$DATE $HOST $MSGHDR$MSG")
			batch-bytes(10MiB)
			batch-timeout(1000)
			retries(3)
		); 
	};
};
```

### DIFF
--- a/modules/http/http-worker.c
+++ b/modules/http/http-worker.c
@@ -91,12 +91,21 @@ _curl_debug_function(CURL *handle, curl_infotype type,
   return 0;
 }
 
+#define HTTP_RESPONSE_MAX_LENGTH 1024
 
 static size_t
 _curl_write_function(char *ptr, size_t size, size_t nmemb, void *userdata)
 {
-  // Discard response content
-  return nmemb * size;
+  HTTPDestinationWorker *self = (HTTPDestinationWorker *) userdata;
+  gsize count = nmemb * size;
+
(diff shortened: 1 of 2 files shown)


modules/http/http-worker.c
```diff
@@ -509,6 +525,7 @@ _custom_map_http_result(HTTPDestinationWorker *self, const gchar *url, HttpRespo
                  evt_tag_str("action", "disconnect"),
                  evt_tag_str("url", url),
                  evt_tag_int("status_code", http_code),
+                 evt_tag_mem("response", self->response_buffer->str, self->response_buffer->len),
```
MrAnno, line 528: A few of these messages are on the "notice" level, not on debug. 

Nevertheless, they seem useful and hopefully they are triggered rarely under normal circumstances.

MrAnno (approved):
A few of these

PR #418, opened 2024-12-14
`network(), syslog()`: Fixed a potential crash for TLS destinations during reload

sodomelle:
Fixes syslog-ng/syslog-ng#5018

It is possible to keep TLS connections alive during reload.
In that case the LogWriter instance is persisted in cfg persist.
This LogWriter's signal slot connector wasn't updated based on the new configuration, which could cause a crash.
The signal slot connector is updated, so the newly configured verifier is used, instead of the old one.

Note that the fix in syslog-ng/syslog-ng#5087 has a security issue, as in that PR, the connector's lifetime is extended, but the verifier plugins are deregistered during reload, which silently disables all TLS verifiers without the user knowing.


tests/light/functional_tests/destination_drivers/network_destination/test_tls_verifier_reload_crash.py
```diff
@@ -0,0 +1,47 @@
...
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+from src.common.file import copy_shared_file
+
+
+def test_tls_verifier_reload_crash(config, syslog_ng, port_allocator, testcase_parameters):
```
mitzkia, line 26: I think the testcase name will not relevant after this fix. So I would suggest other name (or any other similar to this): 
`test_send_log_for_tls_destination_after_a_reload()`

tests/light/functional_tests/destination_drivers/network_destination/test_tls_verifier_reload_crash.py
```diff
@@ -0,0 +1,47 @@
...
+    ca = copy_shared_file(testcase_parameters, "valid-ca.crt")
+
+    network_source = config.create_network_source(port=port_allocator())
+    network_destination = config.create_network_destination(
```
mitzkia, line 30: Maybe in the variable name we should highlight that this is a `tls` destination

tests/light/shared_files/valid-ca.crt
```diff
@@ -0,0 +1,12 @@
+-----BEGIN CERTIFICATE-----
```
mitzkia, line 1: We already have certificates for server and ca in our codebase: `axosyslog/tests/light/shared_files/`.
Why do we need new certificates, are not they good (the old one)?

tests/light/src/common/network.py
```diff
@@ -48,7 +48,10 @@ async def stop(self):
 
     async def close_client(self):
         """After a client connection is accepted, new connections will be rejected until this method is called"""
-        await self._client.close()
+        try:
+            await self._client.close()
+        except ConnectionResetError:  # remove this after fixing two way shutdown (SSL_shutdown) in AxoSyslog
+            pass
```
mitzkia, line 54: This part will silently pass for the testcase, can not we add a `logger.warning("some info about the exception")` until it is not fixed in Axosyslog?

mitzkia (commented):
Thanks a lot for the Light feature update and a new testcase. 
I have added some review notes.

tests/light/functional_tests/destination_drivers/network_destination/test_tls_verifier_reload_crash.py
MrAnno: I think tests that were created for specific bugs should capture the fact in their name, otherwise, one could refactor this test without knowing its purpose (for example, sending more messages before reloading will hide the complex bug we fixed)

tests/light/shared_files/valid-ca.crt
MrAnno: The old certs are not valid and they are expired. Peer verification is key in this test.

tests/light/src/common/network.py
MrAnno: For example,
`client closed connection prematurely`

MrAnno: #420 will fix the issue and we can remove the `try-except`. Let's see which PR goes in first.

tests/light/functional_tests/destination_drivers/network_destination/test_tls_verifier_reload_crash.py
sodomelle: Renamed the function and the file to `test_kept_alive_tls_connection_doing_handshake_after_reload`.

tests/light/src/common/network.py
sodomelle: Added warning message.

tests/light/functional_tests/destination_drivers/network_destination/test_tls_verifier_reload_crash.py
mitzkia: Thanks, I like this new testcase name.

tests/light/src/common/network.py
mitzkia: Thanks. After `logger.warning()` the `pass` expression does not need anymore

sodomelle: Removed.

tests/light/shared_files/valid-ca.crt
mitzkia: Thanks for the info, I did not know about it (the server cert already expired). 
In this case it is ok. Maybe in a separate PR I would remove expired cert and keep only the new

tests/light/src/driver_io/network/network_io.py
```diff
@@ -84,8 +86,12 @@ def construct(self, port, host=None, ip_proto_version=None):
             def _construct(server, reader_class):
                 return reader_class(server), server
 
+            tls = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
+            tls.load_cert_chain(get_shared_file("valid-localhost.crt"), get_shared_file("valid-localhost.key"))
```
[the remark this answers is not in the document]
mitzkia: We have discussed this above. Let keep this new certs now. In a separate PR I will remove old (expired) cert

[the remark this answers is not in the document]
sodomelle: I could not make the TC run correctly with the existing certificate files.

tests/light/functional_tests/destination_drivers/network_destination/test_tls_verifier_reload_crash.py
sodomelle: Added `tls` to the variable name.

lib/transport/transport-tls.c
```diff
@@ -237,6 +237,14 @@ log_transport_tls_write_method(LogTransport *s, const gpointer buf, gsize buflen
   return -1;
 }
 
+TLSSession *
```
bazsi, line 240: We should add the very least add an assertion to check if this is indeed a logtransporttls instance

modules/afsocket/afsocket-dest.c
```diff
@@ -37,13 +37,6 @@
 #include <sys/types.h>
 #include <sys/socket.h>
 
-typedef struct _ReloadStoreItem
```
bazsi, line 40: This is an ugly internal type, I'd prefer to meet this private.

modules/afsocket/afinet-dest.c
```diff
@@ -697,6 +705,31 @@ afinet_dd_free(LogPipe *s)
...
+  if (!proto)
+    return;
+
+  LogTransport *transport = log_transport_stack_get_transport(&proto->transport_stack, LOG_TRANSPORT_TLS);
+
+  if (transport)
+    {
+      TLSSession *session = log_tansport_tls_get_session(transport);
+      AFInetDestDriverTLSVerifyData *verify_data = _get_tls_verify_data (session->verifier);
+      verify_data->signal_connector = self->super.super.super.super.signal_slot_connector;
```
bazsi, line 729: I was just wondering whether we could just drop the connection if the handshake wasn't completed yet instead of trying to keep the connection alive.

This is more like a discussion rather than a request to change.

modules/afsocket/afsocket-dest.h
```diff
@@ -69,6 +69,7 @@ struct _AFSocketDestDriver
   LogWriter *(*construct_writer)(AFSocketDestDriver *self);
   gboolean (*setup_addresses)(AFSocketDestDriver *s);
   const gchar *(*get_dest_name)(const AFSocketDestDriver *s);
+  void (*on_connection_restore)(AFSocketDestDriver *s, ReloadStoreItem *ri);
```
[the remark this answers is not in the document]
MrAnno: We chose `ReloadStoreItem`  instead of `LogWriter` because it was not too clear which LogWriter we pass to this function.
This way, it is possible to manipulate with the ReloadStoreItem fields in child classes, which may come in handy in the future (we almost needed something similar to fix this bug).

We can rename it to `restore_connection()`, but the name would be a little false as this function won't restore the connection, it just opens up the the restoration process for extension.

modules/afsocket/afinet-dest.c
MrAnno: I'm guessing you are suggesting this because the implementation goes through a lot of layers, which looks ugly.

I think querying the transport handshake progress belongs to the `LogTransport` class, so querying such information would also go through the writer->proto->transport stack->transport chain. Because of this, I would rather keep the connection alive if `keep-alive(yes)` is set.

MrAnno: We discussed everything IRL, we keep the current behavior and improve the code around ReloadStoreItem a little more.

modules/afsocket/afsocket-dest.h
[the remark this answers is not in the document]
MrAnno: We discussed everything IRL, we keep the current behavior and improve the code around ReloadStoreItem a little more.

MrAnno:
The `plugin_name` field of `LogPipe` appeared at few useful places, so we didn't remove or refactor it in this PR in the end.
Let's do it in a separate pull request.

lib/transport/transport-tls.c
```diff
@@ -239,6 +241,15 @@ log_transport_tls_write_method(LogTransport *s, const gpointer buf, gsize buflen
...
+log_tansport_tls_get_session(LogTransport *s)
+{
+  g_assert(s->name == TLS_TRANSPORT_NAME);
```
bazsi, line 247: nitpick: I don't think C compilers will always ensure that the pointer value of a literal string expression is always the same. I would have used the address of one of the private methods instead.

bazsi (approved):
I only have one nitpick but this can go in without fixing that.

lib/transport/transport-tls.c
MrAnno: We wanted something that will less likely change if we happen to inherit from the TLS clss.

Using the address of `const gchar *a = "static"` and `const gchar a[] = "static"` should be safe and consistent in C99.

mitzkia (approved):
Thanks for the Light part of the update, it is approved from my side.

MrAnno (approved):

PR #419, opened 2024-12-16
Fix maxOS CI

MrAnno:


sodomelle (approved):
As far as I can tell, it looks good :D and the CI is all green.

PR #424, opened 2024-12-28
dbld: change devshell to use ubuntu-noble as base

bazsi:
Ubuntu has -fno-omit-frame-pointer in CFLAGS which Debian testing does not have, making profiling a lot more difficult on Debian.

At least as long as https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=767756 is not resolved, we are better off on Ubuntu as a devshell.

Note that this patch separates the tarball image (which continues to remain Debian testing) and devshell.

### DIFF
--- a/dbld/packages.manifest
+++ b/dbld/packages.manifest
@@ -42,16 +42,19 @@ which				[fedora]
 #############################################################################
 # eBPF related tools
 #############################################################################
-clang				[tarball]
-libbpf-dev			[tarball]
+clang				[tarball, devshell]
+libbpf-dev			[tarball, devshell]
 bpftool				[tarball]
+# as long as devshell is based on ubuntu and not tarball
+linux-tools-common		[devshell]
+linux-tools-generic		[devshell]
 
 #############################################################################
 # Tarball related tools
 #############################################################################
 
 # docbook to generate man pages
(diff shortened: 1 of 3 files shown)


dbld/packages.manifest
```diff
@@ -42,16 +42,19 @@ which				[fedora]
 #############################################################################
 # eBPF related tools
 #############################################################################
-clang				[tarball]
-libbpf-dev			[tarball]
+clang				[tarball, devshell]
+libbpf-dev			[tarball, devshell]
 bpftool				[tarball]
```
OverOrion, line 47: I think `bpftool` should be installed in `devshell` as well, what do you think?

OverOrion (approved):
Thank you! :rocket:

dbld/packages.manifest
bazsi: It is installed, but on ubuntu it comes from a different package (linux-tools-common and the versioned linux-tools builds)

bazsi:
This broke the CI unfortunately, the host running our CI job may have  a kernel for which linux-tools may not have been installed. Unfortunately all the bpftools related binaries are resolved through those linux-tools packages. For now I would just back this out.

PR #431, opened 2024-12-30
Filterx: check if value is a macro during compilation

bazsi:
This PR ended a bit more than what the title says:
* the original goal was to check whether a variable is a hard macro and do this in the compilation phase, instead of during evaluation
* then I removed the macro wrapper for $MSG, which was not really doing much anymore
* this broke value-pairs
* also broke light tests, which relied on $MSG being a string, even though it was set to a protobuf/bytes value
* and I also noticed an read-side off-by-one for macro based variables
* I ended up also adding a warning when using bytes or protobuf values in a template which does not render without a type hint.

With all that fixed is what we have in this branch.


OverOrion (approved):
Thank you!

PR #435, opened 2024-12-30
filterx keys function

bshifter:
example:
```
            d = {"foo":1,"bar":[3,2,1],"baz":{"tik":"tak"}};
            $MSG.keys = keys(d); #["foo","bar","baz"]
            $MSG.elt = keys(d)[0]; #"foo"
```

### DIFF
--- a/lib/filterx/func-keys.c
+++ b/lib/filterx/func-keys.c
@@ -1,0 +1,139 @@
+/*
+ * Copyright (c) 2024 Axoflow
+ * Copyright (c) 2024 shifter
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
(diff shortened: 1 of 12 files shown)


lib/filterx/func-keys.c
```diff
@@ -0,0 +1,139 @@
...
+
+static gboolean
+_extract_args(FilterXFunctionKeys *self, FilterXFunctionArgs *args, GError **error)
+{
+  if (filterx_function_args_len(args) != 1)
+    {
+      g_set_error(error, FILTERX_FUNCTION_ERROR, FILTERX_FUNCTION_ERROR_CTOR_FAIL,
+                  FILTERX_FUNC_KEYS_ERR_NUM_ARGS FILTERX_FUNC_KEYS_USAGE);
+      return FALSE;
+    }
```
bazsi, line 109: I think filterx_function_args_check() would validate that we indeed fetched all arguments and would return a failure if not.

bazsi (approved):
Just one minor comment, also I feel increasingly that the function support code needs to be improved, there's still to much boiler plate around creating a function and writing its argument parsing.

In a similar vein, I'd want to turn FilterXFunction as an expr into a FilterXCall, which operates on various function types, e.g. simple and generator and generic, each with more supporting framework to help facilitate the writing of those functions.

But that's just a side note, this one can go in even with the extra argument check. I'd appreciate the removal of all similar checks as a separate PR though.

lib/filterx/func-keys.c
bazsi: All the other functions have a similar check, so I guess this will need to be resolved if the "function" framework is more sophisticated.
Resolving my own comment.

alltilla:
@bshifter Can you please create a newsfile entry for this in another PR, thanks!

PR #436, opened 2025-01-02
filterx: optimize followup

alltilla:
Followup for #421

### DIFF
--- a/lib/filterx/expr-null-coalesce.c
+++ b/lib/filterx/expr-null-coalesce.c
@@ -67,20 +67,18 @@ _optimize(FilterXExpr *s)
   if (filterx_binary_op_optimize_method(s))
     g_assert_not_reached();
 
-  if (filterx_expr_is_literal(self->super.lhs))
-    {
-      FilterXObject *lhs_object = filterx_expr_eval(self->super.lhs);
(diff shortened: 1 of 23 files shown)


lib/filterx/expr-null-coalesce.c
```diff
@@ -59,26 +59,36 @@ _eval(FilterXExpr *s)
...
+static FilterXExpr *
+_optimize(FilterXExpr *s)
 {
-  if (filterx_expr_is_literal(lhs))
+  FilterXNullCoalesce *self = (FilterXNullCoalesce *) s;
+
+  if (filterx_binary_op_optimize_method(s))
+    g_assert_not_reached();
+
+  if (filterx_expr_is_literal(self->super.lhs))
```
OverOrion, line 70: [Nitpick, absolutely optional]

Thanks for retaining the original structure while introducing the `optimize` method. Would you be willing to refactor this to use early returns where possible?

OverOrion (approved):
Thanks for this! :rocket:

lib/filterx/expr-null-coalesce.c
alltilla: Not a problem, thanks for the reminder! Done :)

OverOrion (approved):

PR #443, opened 2025-01-06
Filterx small changes

bazsi:
These are three basically independent patches, but I didn't want to open a separate PR for all of these. Let's try these in one go. Should be relatively straightforward to review.

### DIFF
--- a/lib/filterx/filterx-expr.c
+++ b/lib/filterx/filterx-expr.c
@@ -88,7 +88,9 @@ filterx_expr_optimize(FilterXExpr *self)
   optimized = filterx_expr_optimize(optimized);
 
   msg_trace("FilterX: expression optimized",
-            filterx_expr_format_location_tag(self));
+            filterx_expr_format_location_tag(self),
+            evt_tag_str("old_type", self->type),
+            evt_tag_str("new_type", optimized->type));
   if (self->lloc)
     {
       /* copy location information to the optimized representation */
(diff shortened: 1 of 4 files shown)


lib/filterx/filterx-expr.c
```diff
@@ -88,7 +88,9 @@ filterx_expr_optimize(FilterXExpr *self)
   optimized = filterx_expr_optimize(optimized);
 
   msg_trace("FilterX: expression optimized",
-            filterx_expr_format_location_tag(self));
+            filterx_expr_format_location_tag(self),
+            evt_tag_str("old_type", self->type),
+            evt_tag_str("new_type", optimized->type));
```
alltilla, line 93: I don't see `type` ever get a value, in fact commenting it out from the struct builds just fine. I think we should fill the `type` field for every `Expr` before using it in logs.

bazsi: I'm extracting that commit into a separate PR, as I have a number of subsequent commits that builds on top of that.

bazsi: I have now rebased this against #444, which means this contains that patch in addition to master. With that this should compile and once that patch goes in, the rest of my patches in #434 can move up to.

bazsi:
This is now based on top of #444 and with that this is green now.

alltilla (dismissed):

alltilla (approved):

alltilla:
Thanks for this! After a rebase we can merge.

alltilla (approved):

PR #445, opened 2025-01-07
Filterx expr constructor improvements

bazsi:
Next out of #434, this time it is based on #444 that needs to go in first.

There are two patches here (on top of #444):
* the nullv style constructors are reimplemented as a "delta" to the normal constructor to avoid code duplication
* _eval() methods are renamed everywhere so it's easier to read perf stackdumps


bazsi:
rebased against master to resolve conflicts.

alltilla (approved):

PR #447, opened 2025-01-07
logmsg: add generation counter

bazsi:
This adds a counter to LogMessage to track changes made to the NVTable. Any set_value() would increment that counter. Any other fields (timestamps, etc) are not yet tracked, but probably should....


OverOrion (approved):
Thank you!

For the curious, the previous struct size was 160 bytes with holes, now only 152 (no holes!), so yay as well for this! :rocket:

PR #448, opened 2025-01-07
filterx/filterx-scope: keep track of the maximum number of variables

bazsi:
Preallocate the right number of variables so that we never have to extend the variables array during runtime.

### DIFF
--- a/lib/filterx/filterx-scope.c
+++ b/lib/filterx/filterx-scope.c
@@ -30,6 +30,8 @@ struct _FilterXScope
   guint32 generation:20, write_protected, dirty, syncable, log_msg_has_changes;
 };
 
+volatile gint filterx_scope_variables_max = 16;
+
 static gboolean
 _lookup_variable(FilterXScope *self, FilterXVariableHandle handle, FilterXVariable **v_slot)
 {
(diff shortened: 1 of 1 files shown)


lib/filterx/filterx-scope.c
```diff
@@ -353,6 +355,11 @@ filterx_scope_make_writable(FilterXScope **pself)
 static void
 _free(FilterXScope *self)
 {
+  /* NOTE: update the number of inlined variable allocations */
+  gint variables_max = filterx_scope_variables_max;
+  if (variables_max < self->variables->len)
+    filterx_scope_variables_max = self->variables->len;
```
MrAnno, line 361: Can we use `g_atomic_int` here?

I don't know any architectures where a `volatile` wouldn't be enough for this purpose, so I'm just asking.

bazsi: I would try to avoid atomics if possible, especially on the read side. It does not matter if we race on  the read itself, but atomics have additional overhead and we are allocating a scope 30-40 times per message.

By itself, one atomic is not visible in the profile, but the number of total atomic operations per second matters,
just as the reduction of malloc() calls improved scalability a lot.

So I would keep this as is, especially until our scalability starts to decline at around 5-6 cores.

With the current per core performance, we need to scale almost linearly on typical server class CPUs (at least 8
cores, but would be better to 16).

MrAnno (approved):

PR #452, opened 2025-01-09
Filterx: unset-empties() performance changes

bazsi:
UTF8 aware case folding are expensive, remove that feature for now. The branch also changes the default for the ignorecase parameter.


bshifter (approved):
lgtm

PR #453, opened 2025-01-09
Filterx string performance opt

bazsi:
This removes another set of allocations in upper/lower/includes/startswith/endswith

these are often used in initial qualification of the data, so their performance is important.


OverOrion (approved):
Thank you!

PR #456, opened 2025-01-15
filterx pubsub message 

bshifter:
Add pubsub_message() function to FilterX and extend Pub/Sub destination with protovar

This PR introduces new functionality for managing Google Pub/Sub messages in syslog-ng:

1. pubsub_message() Function:

* Adds a FilterX representation of the Google Pub/Sub message Protobuf descriptor.
* Enables users to define and manipulate Pub/Sub message content directly within the FilterX scripting module (source-side).

2. protovar Option in Pub/Sub Destination:

* Extends the Pub/Sub destination module with the protovar option.
* Allows referencing the FilterX-defined Pub/Sub Protobuf variable for setting the outgoing message payload.
Together, these changes provide greater flexibility and control over Pub/Sub destination messages, supporting advanced customization directly from FilterX.

example:
```
filterx {
    # Adding data and attributes 
    $x = pubsub_message($MESSAGE, {"foo":{"bar":"baz"}});
};

log {
    destination {
        google-pubsub-grpc(
            ...
            proto_var($x)
        );
    };
};


```


modules/grpc/pubsub/filterx/object-pubsub-message.cpp
```diff
@@ -0,0 +1,252 @@
...
+{
+  auto *attr = static_cast<std::map<std::string, std::string>*>(user_data);
+
+  GString *key_str = scratch_buffers_alloc();
+  GString *val_str = scratch_buffers_alloc();
+
+  if (!filterx_object_repr(key, key_str))
+    return FALSE;
+  if (!filterx_object_repr(val, val_str))
+    return FALSE;
```
alltilla, line 183: In the majority of the time these will already be strings, and `repr()` copies strings even. For strings we could call `get_value()`, to optimize out the copy, for the rest `repr()` is perfect.

modules/grpc/pubsub/filterx/object-pubsub.h
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright (c) 2024 Axoflow
+ * Copyright (c) 2023 shifter
```
alltilla, line 3: Please use recent dates.

modules/grpc/pubsub/filterx/object-pubsub.h
```diff
@@ -0,0 +1,59 @@
...
+static inline FilterXObject *
+filterx_pubsub_message_new(void)
+{
+  return filterx_pubsub_message_new_from_args(NULL, NULL, 0);
+}
+
+FILTERX_DECLARE_TYPE(pubsub_message);
+
+static inline void
+pubsub_filterx_objects_global_init(void)
```
alltilla, line 46: Shouldn't this be called in `google_pubsub_grpc_module_init()`?

modules/grpc/pubsub/pubsub-dest-worker.cpp
```diff
@@ -99,30 +99,70 @@ DestWorker::insert(LogMessage *msg)
...
 
   ::google::pubsub::v1::PubsubMessage *message = this->request.add_messages();
 
-  buf_slice = this->format_template(owner_->data, msg, buf, NULL, this->super->super.seq_num);
-  message->set_data(buf_slice.str, buf_slice.len);
-  message_bytes += buf_slice.len;
-
-  auto attributes = message->mutable_attributes();
-  for (const auto &attribute : owner_->attributes)
+  if (owner_->protovar)
```
alltilla, line 106: [optional]

I think this is a good candidate for factoring out to some smaller functions.

modules/grpc/pubsub/filterx/object-pubsub-message.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+  *assoc_object =  filterx_json_new_from_repr(serialized.c_str(), serialized.length());
+  // *object = filterx_json_object_get_value(filterx_object_ref(*assoc_object));
+  *object = json_object_new_object();
+  json_object_object_foreach(filterx_json_object_get_value(*assoc_object), key, val)
+  {
+    json_object_object_add(*object, key, json_object_get(val));
+  }
+  return TRUE;
+}
```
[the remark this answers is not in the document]
bshifter: This implementation is intended solely for debugging purposes. As a result, I prioritized simplicity over performance and focused on providing the most straightforward representation of the object in JSON. However, I encountered some unexpected behavior, which made the implementation far more complex than it should have been. Thank you for the suggestion to simplify the code.

modules/grpc/pubsub/filterx/object-pubsub-message.cpp
```diff
@@ -0,0 +1,352 @@
...
+Message::set_attribute(const std::string &key, const std::string &value)
+{
+  if (key.empty())
+    {
+      throw std::invalid_argument("Key cannot be empty");
+    }
+  if (value.empty())
+    {
+      throw std::invalid_argument("Value cannot be empty");
+    }
```
alltilla, line 105: These exceptions are not caught, which can cause an abort.

modules/grpc/pubsub/filterx/object-pubsub-message.cpp
```diff
@@ -0,0 +1,352 @@
...
+          (*attributes)[key] = value;
+        }
+      else
+        {
+          throw std::runtime_error("Failed to access mutable attributes");
+        }
+    }
+  catch (const std::exception &e)
+    {
+      return false;
```
alltilla, line 121: Shouldn't we log the error?

modules/grpc/pubsub/filterx/object-pubsub-message.cpp
```diff
@@ -0,0 +1,352 @@
...
+              else if (filterx_object_is_type(data_arg, &FILTERX_TYPE_NAME(bytes)))
+                {
+                  const gchar *data_str = filterx_bytes_get_value_ref(data_arg, &len);
+                  data_cpp = std::string(data_str, len);
+                }
+              else if (filterx_object_is_type(data_arg, &FILTERX_TYPE_NAME(protobuf)))
+                {
+                  const gchar *data_str = filterx_protobuf_get_value_ref(data_arg, &len);
+                  data_cpp = std::string(data_str, len);
+                }
```
alltilla, line 298: Sorry, here also I forgot about `message_value`s. I think we should use the extractor here, but the good news is that it will make the code a lot more simpler, as you can call each `extract()` in one `if` condition.

alltilla:
Thanks for the updates on the PR. I found a couple more things that I unfortunately missed the last time.

alltilla (approved):

PR #458, opened 2025-01-15
dbld: install sudo package into RHEL based docker images

OverOrion:
This is a backport of https://github.com/syslog-ng/syslog-ng/pull/5210


MrAnno:
#460 disables privileged containers, which will fix the issue.
The issue will be still reproducible in `dbld shell-*`, let's investigate it separately.

PR #460, opened 2025-01-16
Fix rpm build

MrAnno:
This is only needed for gdb sessions, but it interferes with auth modules
in sudo, for example.

The issue will be still reproducible in `dbld shell-*`, let's investigate it separately.


OverOrion (approved):
Thank you for getting to the bottom of it! :rocket:

PR #464, opened 2025-01-23
http: enable flush-on-worker-key-change for 1 worker

alltilla:
When the URL is templated and batching is enabled it does not matter how many workers we have, we
need to close the batch when the partition key changes.

The condition for the if statement was wrong for a long time, however since 70c704c037 it not only resulted in a missing error log and missing init fail, but we ran with `flush-on-worker-key-change(no)`, messing up the batching.


MrAnno (approved):
Oh :(

bazsi (approved):

PR #465, opened 2025-01-23
Filterx expr type macros

bazsi:
The current convention is to decide whether we are looking at a specific FilterXExpr instance based on the value of the
"eval" function pointer.

Issue is that #433 will change that pointer at runtime to incorporate expr level visibility into perf backtraces.

This could also be the beginning of a real FilterXExprType struct (to avoid repeating the function pointers into all expr instances), but this PR does not go that far.

NOTE: this does not create FILTERX_EXPR_TYPE_DECLARE() for all exprs, only those which had a filterx_expr_is_XXX() function. As long as noone is checking if an expr is of a specific type, this is optional.

### DIFF
--- a/lib/filterx/filterx-expr.h
+++ b/lib/filterx/filterx-expr.h
@@ -68,6 +68,20 @@ struct _FilterXExpr
   gchar *expr_text;
 };
 
+#define FILTERX_EXPR_TYPE_NAME(_type) filterx_expr_type_ ## _type
+
+#define FILTERX_EXPR_DECLARE_TYPE(_type) \
+  extern const gchar *FILTERX_EXPR_TYPE_NAME(_type); \
+  \
+  static inline gboolean \
+  filterx_expr_is_ ## _type(FilterXExpr *expr) \
+  { \
+    return expr && expr->type == FILTERX_EXPR_TYPE_NAME(_type); \
+  }
+
+#define FILTERX_EXPR_DEFINE_TYPE(_type) \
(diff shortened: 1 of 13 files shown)


lib/filterx/filterx-expr.h
```diff
@@ -68,6 +68,20 @@ struct _FilterXExpr
...
+#define FILTERX_EXPR_TYPE_NAME(_type) filterx_expr_type_ ## _type
+
+#define FILTERX_EXPR_DECLARE_TYPE(_type) \
+  extern const gchar *FILTERX_EXPR_TYPE_NAME(_type); \
+  \
+  static inline gboolean \
+  filterx_expr_is_ ## _type(FilterXExpr *expr) \
+  { \
+    return expr && expr->type == FILTERX_EXPR_TYPE_NAME(_type); \
+  }
```
MrAnno, line 80: It's not easy to find the existence of this function when it's hidden under a macro with such a generic name.

If we really want to use a macro to generate this one-liner function, can we separate it and give it a name related to is_*_type?

I would prefer using explicit functions or just the type comparison expression that is inside this function to avoid more macro-generated code that is difficult to find.

bazsi: I removed the declaration from the macro and defined it for all types.

MrAnno (approved):

lib/filterx/filterx-expr.h
MrAnno: Thank you

PR #466, opened 2025-01-23
Time releated fixes

bshifter:
A bunch of datetime releated fixes:
- dump functions for wallclock time and unixtime (they were extermely useful, so i kept them)
- fix: filterx datetime string formatting error: unnecessary dash removed
- prepartion for nano second precision support: timeutils/misc extended, typo fixed
- fix:  unix_time_to_unix_epoch calculation
- fix: otel nanosecond-precision fields setter/getter
- unit test for otel nanosec precision and timezone information

### DIFF
--- a/lib/filterx/object-datetime.c
+++ b/lib/filterx/object-datetime.c
@@ -134,14 +134,14 @@ filterx_typecast_datetime(FilterXExpr *s, FilterXObject *args[], gsize args_len)
   gint64 i;
   if (filterx_object_extract_integer(object, &i))
     {
-      UnixTime ut = unix_time_from_unix_epoch((guint64) MAX(i, 0));
+      UnixTime ut = unix_time_from_unix_epoch_usec((guint64) MAX(i, 0));
       return filterx_datetime_new(&ut);
     }
 
   gdouble d;
   if (filterx_object_extract_double(object, &d))
     {
-      UnixTime ut = unix_time_from_unix_epoch((gint64)(d * USEC_PER_SEC));
+      UnixTime ut = unix_time_from_unix_epoch_usec((gint64)(d * USEC_PER_SEC));
       return filterx_datetime_new(&ut);
     }
 
--- a/lib/filterx/object-datetime.c
+++ b/lib/filterx/object-datetime.c
@@ -69,8 +69,6 @@ _convert_unix_time_to_string(const UnixTime *ut, GString *result, gboolean inclu
     {
       if (ut->ut_gmtoff >= 0)
         g_string_append_c(result, '+');
-      else
-        g_string_append_c(result, '-');
 
       format_int64_padded(result, 2, '0', 10, ut->ut_gmtoff / 3600);
       g_string_append_c(result, ':');
--- a/lib/timeutils/unixtime.c
+++ b/lib/timeutils/unixtime.c
@@ -370,8 +370,7 @@ unix_time_from_unix_epoch_usec(guint64 unix_epoch_usec)
 guint64
 unix_time_to_unix_epoch_usec(const UnixTime ut)
 {
-  gint32 gmtoff = (ut.ut_gmtoff != - 1) ? ut.ut_gmtoff : 0;
-  return (guint64)((ut.ut_sec + gmtoff) * USEC_PER_SEC + ut.ut_usec);
+  return (guint64)(ut.ut_sec * USEC_PER_SEC + ut.ut_usec);
 }
 
 UnixTime
--- a/lib/timeutils/unixtime.c
+++ b/lib/timeutils/unixtime.c
@@ -359,16 +359,16 @@ _unix_time_parts_from_unix_epoch(guint64 unix_epoch, gint64 *secs, guint32 *usec
 }
 
 UnixTime
-unix_time_from_unix_epoch(guint64 unix_epoch)
+unix_time_from_unix_epoch_usec(guint64 unix_epoch_usec)
 {
   UnixTime ut = UNIX_TIME_INIT;
   ut.ut_gmtoff = 0;
-  _unix_time_parts_from_unix_epoch(unix_epoch, &ut.ut_sec, &ut.ut_usec);
+  _unix_time_parts_from_unix_epoch(unix_epoch_usec, &ut.ut_sec, &ut.ut_usec);
   return ut;
 }
 
 guint64
-unix_time_to_unix_epoch(const UnixTime ut)
+unix_time_to_unix_epoch_usec(const UnixTime ut)
 {
   gint32 gmtoff = (ut.ut_gmtoff != - 1) ? ut.ut_gmtoff : 0;
   return (guint64)((ut.ut_sec + gmtoff) * USEC_PER_SEC + ut.ut_usec);
--- a/lib/timeutils/unixtime.c
+++ b/lib/timeutils/unixtime.c
@@ -400,3 +400,34 @@ unix_time_add_duration(UnixTime time, guint64 duration)
   return ut;
 
 }
+
+void
+dump_unix_time(const UnixTime *ut, GString *output)
+{
+  if (!ut || !output)
+    return;
+
+  g_string_append_printf(output, "UnixTime:\n");
+  g_string_append_printf(output, "  Seconds since epoch: %" G_GINT64_FORMAT "\n", ut->ut_sec);
+  g_string_append_printf(output, "  Microseconds: %u\n", ut->ut_usec);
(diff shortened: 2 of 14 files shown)


lib/timeutils/unixtime.h
```diff
@@ -69,6 +69,37 @@ unix_time_is_timezone_set(const UnixTime *self)
   return self->ut_gmtoff != -1;
 }
 
+static inline
+void dump_unix_time(const UnixTime *ut, GString *output)
```
bazsi, line 73: can you move the return value type to the previous line? that's the convention we are using, e.g.

```
static inline void
dump_unix_time(...)
```

bazsi: also, we might hide this in the .c file, I don't see why this should be inline.

lib/timeutils/wallclocktime.h
```diff
@@ -124,6 +124,33 @@ struct _WallClockTime
...
+  g_string_append_printf(output, "  Microsecond: %d\n", wct->wct_usec);
+  g_string_append_printf(output, "  DST: %d\n", wct->wct_isdst);
+
+#ifdef SYSLOG_NG_HAVE_STRUCT_TM_TM_GMTOFF
+  g_string_append_printf(output, "  GMT Offset: %ld\n", wct->wct_gmtoff);
+  g_string_append_printf(output, "  Timezone: %s\n", wct->wct_zone ? wct->wct_zone : "(null)");
+#else
+  g_string_append_printf(output, "  GMT Offset: %ld\n", wct->wct_gmtoff);
+  g_string_append_printf(output, "  Timezone: %s\n", wct->wct_zone ? wct->wct_zone : "(null)");
+#endif
```
bazsi, line 151: I don't see the difference in these two cases. The idea was to always have wct_gmtoff and wct_zone in WallClockTime so that we have those on all platforms. I don't think we need the conditional compilation.

lib/timeutils/wallclocktime.h
```diff
@@ -124,6 +124,33 @@ struct _WallClockTime
 
 #endif
 
+static inline
+void dump_wall_clock_time(const WallClockTime *wct, GString *output)
```
bazsi, line 128: again, this can go to the implementation file and please move "void" to the previous line.

lib/timeutils/unixtime.c
```diff
@@ -371,7 +371,7 @@ guint64
 unix_time_to_unix_epoch(const UnixTime ut)
```
bazsi, line 371: this function name is misleading for me. I know this patch does not introduce these, but for me "time since unix epoch" is in seconds and not in any particular resolution (msec, usecs or nsecs).

I would rename these to unix_time_from_unix_epoch_usec() and unix_time_to_unix_epoch_usec() to show the 
granularity.

Later (if needed we can add the sec/nsec variants as well).

lib/timeutils/unixtime.c
```diff
@@ -371,7 +371,7 @@ guint64
...
 {
   gint32 gmtoff = (ut.ut_gmtoff != - 1) ? ut.ut_gmtoff : 0;
-  return (guint64)((ut.ut_sec + gmtoff) * USEC_PER_SEC + ut.ut_usec);
+  return (guint64)((ut.ut_sec - gmtoff) * USEC_PER_SEC + ut.ut_usec);
```
bazsi, line 374: unit tests for this would be dearly needed.

modules/grpc/otel/filterx/otel-field.cpp
```diff
@@ -269,7 +272,7 @@ class OtelDatetimeConverter : public ProtobufField
   FilterXObject *FilterXObjectGetter(Message *message, ProtoReflectors reflectors)
   {
     uint64_t val = reflectors.reflection->GetUInt64(*message, reflectors.fieldDescriptor);
-    UnixTime utime = unix_time_from_unix_epoch(val);
+    UnixTime utime = unix_time_from_unix_epoch(NSEC_TO_USEC(val));
```
bazsi, line 275: with this, I'd probably introduce the unix_time_from_unix_epoch_nsec() function, so we can push the NSEC_TO_USEC to be within the function body and not here.

modules/grpc/otel/filterx/otel-field.cpp
```diff
@@ -279,7 +282,7 @@ class OtelDatetimeConverter : public ProtobufField
     if (filterx_object_extract_datetime(object, &utime))
       {
         uint64_t unix_epoch = unix_time_to_unix_epoch(utime);
-        reflectors.reflection->SetUInt64(message, reflectors.fieldDescriptor, unix_epoch);
+        reflectors.reflection->SetUInt64(message, reflectors.fieldDescriptor, unix_epoch * NSEC_PER_USEC);
```
bazsi, line 285: likewise, unix_time_unix_epoch_nsec() could do the multiplication

lib/timeutils/unixtime.c
```diff
@@ -359,19 +359,31 @@ _unix_time_parts_from_unix_epoch(guint64 unix_epoch, gint64 *secs, guint32 *usec
...
   return ut;
 }
 
 guint64
-unix_time_to_unix_epoch(const UnixTime ut)
+unix_time_to_unix_epoch_usec(const UnixTime ut)
 {
   gint32 gmtoff = (ut.ut_gmtoff != - 1) ? ut.ut_gmtoff : 0;
-  return (guint64)((ut.ut_sec + gmtoff) * USEC_PER_SEC + ut.ut_usec);
+  return (guint64)((ut.ut_sec - gmtoff) * USEC_PER_SEC + ut.ut_usec);
```
alltilla, line 374: I don't think we should subtract the GMT offset here. `UnixTime`'s `ut_sec` and `ut_usec` fields are storing values that correspond to UTC already, the GMT offset is a hint to the original timezone. If you change the timezone in a `UnixTime` object, you don't need to modify `ut_sec` and `ut_usec`.

Let me demonstrate this with a config:
```
log {

  filterx {
    dt = isodate("2025-01-01T04:00:00-03:00");
    $logmsg = dt;
    $js = json({"dt": dt});
    $repr = string(dt);
    $int = int(dt);
    $double = double(dt);

    log = otel_logrecord();
    log.time_unix_nano = dt;
    $otel_dt_as_int = int(log.time_unix_nano);
    $otel_dt_as_double = double(log.time_unix_nano);

    log.attributes.int = dt;
    log.attributes.double = dt;
    $otel_int = log.attributes.int;
  };

  destination {
    stdout(template("$logmsg\n$js\n$repr\n$int\n$double\n$otel_dt_as_int\n$otel_dt_as_double\n$otel_int\n"));
  };
};

```
```
$ ./install/sbin/syslog-ng -Fe
[2025-01-24T10:54:56.623207] Setting current version as config version; version='4.9'
[2025-01-24T10:54:56.623207] syslog-ng starting up; version='4.9.0.155.g8aed412.dirty'
1735714800.000000-3:00
{"dt":"1735714800.000000"}
2025-01-01T04:00:00.000-03:00
1735714800000000
1735714800
1735725600000000
1735725600
1735725600000000
```

From https://greenwichmeantime.com/articles/clocks/iso/ we can see that `1735714800` is the correct value.
image

We can deduce 2 things from this output:
1. We are not using the same logic for otel and the rest of the filterx code.
2. The otel logic is erroneous.

We should fix both. Let's fix this function and use it everywhere:
* datetime -> logmsg (`marshal()`)
* datetime -> json (`map_to_json()`)
* datetime -> string (`repr()`)
* datetime -> int (`typecast_integer()`)
* datetime -> double (`typecast_double()`)
* datetime -> otel int (`u64Field`, `i64Field`)
* datetime -> otel_datetime (`OtelDatetimeConverter`)

bazsi: Wildly agree. Sorry that I missed this.

bshifter: Ohh gosh! My bad, i've missed the concept of Unixtime, and thought this was a bug whole time. Let me clean up my mess!

lib/filterx/object-datetime.c
```diff
@@ -69,8 +69,6 @@ _convert_unix_time_to_string(const UnixTime *ut, GString *result, gboolean inclu
     {
       if (ut->ut_gmtoff >= 0)
         g_string_append_c(result, '+');
-      else
-        g_string_append_c(result, '-');
```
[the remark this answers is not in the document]
bazsi: I saw the unit test.

lib/timeutils/unixtime.c
bazsi: Thanks for annotating these functions.

bazsi (approved):
This looks good to me, all my review notes were fixed.

PR #467, opened 2025-01-23
Filterx scope to track message changes

bazsi:
This pull request replaces the old "log_msg_has_changes" mechanism with the recently merged generation counter.

This has the following goals/benefits:
* it fixes a bug in message tracking, as previously we considered a message changed when it was cloned. But a message can change without being cloned.
* we delegate the message change tracking to FilterXScope, 
  * this requires that FilterXScope now has a reference to the message being processed. 
  * this becomes similar to what we do with "undeclared" floating variables
  * the scope will provide a higher level interface to manage variables, moving the complexity from FilterXVariable to FilterXScope
* it adds slight changes and makes a couple of functions inline.
*

### DIFF
--- a/lib/filterx/expr-variable.c
+++ b/lib/filterx/expr-variable.c
@@ -41,31 +41,6 @@ typedef struct _FilterXVariableExpr
 } FilterXVariableExpr;
 
 static FilterXObject *
-_pull_variable_from_message(FilterXVariableExpr *self, FilterXEvalContext *context, LogMessage *msg)
-{
-  gssize value_len;
-  LogMessageValueType t;
-  const gchar *value = log_msg_get_value_if_set_with_type(msg, self->handle, &value_len, &t);
-  if (!value)
-    {
-      filterx_eval_push_error("No such name-value pair in the log message", &self->super, self->variable_name);
-      return NULL;
-    }
-
-  if (self->handle_is_macro)
-    return filterx_message_value_new(value, value_len, t);
-  else
-    return filterx_message_value_new_borrowed(value, value_len, t);
-}
-
-/* NOTE: unset on a variable that only exists in the LogMessage, without making the message writable */
-static void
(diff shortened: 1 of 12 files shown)


lib/filterx/expr-variable.c
```diff
@@ -75,21 +50,20 @@ _eval_variable(FilterXExpr *s)
...
-      FilterXObject *value = filterx_variable_get_value(variable);
+      FilterXObject *value = filterx_scope_get_variable(context->scope, variable);
       if (!value)
         {
           filterx_eval_push_error("Variable is unset", &self->super, self->variable_name);
         }
       return value;
     }
 
-  if (filterx_variable_handle_is_message_tied(self->handle))
```
MrAnno, line 86: `filterx_variable_handle_is_message_tied()` is no longer used. We should probably remove it to avoid confusion with `FX_VAR_MESSAGE_TIED`

lib/filterx/filterx-variable.h
```diff
@@ -71,17 +70,15 @@ typedef struct _FilterXVariable
    *
    * declared -- this variable is declared (e.g. retained for the entire input pipeline)
    */
-  guint32 assigned:1,
-          variable_type:2,
```
MrAnno, line 75: The number of values in `FilterXVariableType` is not checked against this `:2` bitfield. Can we add a static assert to make sure it breaks at compile time when one adds 2 more fields?

lib/filterx/expr-variable.c
bazsi: good point. fixed.

lib/filterx/filterx-variable.h
bazsi: fixed

bazsi:
fixed up the comments.

MrAnno (approved):
Thank you

PR #468, opened 2025-01-23
Filterx scope optimizations

bazsi:
This goes on top of #467 and brings significant performance improvements for FilterXScope

### DIFF
--- a/lib/filterx/filterx-pipe.c
+++ b/lib/filterx/filterx-pipe.c
@@ -66,9 +66,22 @@ log_filterx_pipe_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_o
   FilterXEvalContext eval_context;
   LogPathOptions local_path_options;
   FilterXEvalResult eval_res;
+  gboolean local_scope = FALSE;
 
   path_options = log_path_options_chain(&local_path_options, path_options);
-  filterx_eval_init_context(&eval_context, path_options->filterx_context, msg);
+
+  FilterXScope *scope = NULL;
+  if (path_options->filterx_context)
+    scope = filterx_scope_reuse(path_options->filterx_context->scope);
+
+  if (!scope)
+    {
+      gsize alloc_size = filterx_scope_get_alloc_size();
+      scope = g_alloca(alloc_size);
+      filterx_scope_init_instance(scope, alloc_size, path_options->filterx_context ? path_options->filterx_context->scope : NULL);
+      local_scope = TRUE;
+    }
+  filterx_eval_init_context(&eval_context, path_options->filterx_context, scope, msg);
 
   msg_trace(">>>>>> filterx rule evaluation begin",
             evt_tag_str("rule", self->name),
@@ -106,6 +119,8 @@ log_filterx_pipe_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_o
     }
 
   filterx_eval_deinit_context(&eval_context);
+  if (local_scope)
+    filterx_scope_clear(scope);
   nv_table_unref(payload);
 }
 
--- a/lib/filterx/filterx-pipe.c
+++ b/lib/filterx/filterx-pipe.c
@@ -64,63 +64,51 @@ log_filterx_pipe_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_o
 {
   LogFilterXPipe *self = (LogFilterXPipe *) s;
   FilterXEvalContext eval_context;
-  LogPathOptions local_path_options;
-  FilterXEvalResult eval_res;
-  gboolean local_scope = FALSE;
-
-  path_options = log_path_options_chain(&local_path_options, path_options);
-
-  FilterXScope *scope = NULL;
-  if (path_options->filterx_context)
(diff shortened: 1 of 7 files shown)


lib/filterx/filterx-scope.h
```diff
@@ -43,11 +43,22 @@ typedef struct _FilterXScope FilterXScope;
...
+  struct
+  {
+    /* number of elems */
+    gint len;
+    /* allocated elems */
+    gint size;
+    union
+    {
+      FilterXVariable *separate;
+      FilterXVariable coupled[0];
```
MrAnno, line 58: Does it really worth it to save an indirection and an initial malloc to inline variables into the scope?
If it's really worth the code complexity and want to keep it, can we rename `coupled` to `inlined`, `separate` to `allocated`, and `coupled_variables` to `inlined_variables`?

lib/filterx/filterx-scope.c
```diff
@@ -356,12 +420,21 @@ _free(FilterXScope *self)
 {
   /* NOTE: update the number of inlined variable allocations */
   gint variables_max = filterx_scope_variables_max;
-  if (variables_max < self->variables->len)
-    filterx_scope_variables_max = self->variables->len;
+  if (variables_max < self->variables.len)
+    filterx_scope_variables_max = self->variables.len;
```
MrAnno, line 392: I think we should put a hard-coded upper limit to this variable.

If someone loads a configuration with millions of variables, not even a reload will help reduce the memory usage.

lib/filterx/filterx-pipe.c
```diff
@@ -66,9 +66,22 @@ log_filterx_pipe_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_o
...
   path_options = log_path_options_chain(&local_path_options, path_options);
-  filterx_eval_init_context(&eval_context, path_options->filterx_context, msg);
+
+  FilterXScope *scope = NULL;
+  if (path_options->filterx_context)
+    scope = filterx_scope_reuse(path_options->filterx_context->scope);
+
+  if (!scope)
+    {
+      gsize alloc_size = filterx_scope_get_alloc_size();
```
MrAnno: Now that we are on the stack, we should definitely put an upper limit on the maximum allocated variables count.

MrAnno (commented):
Please revert ` f917c67d02d6a60883ed7da7915bdaf37338a998` in this PR.

lib/filterx/filterx-scope.h
bazsi: We have quite a few FilterXScope instances, as each and every filterx block creates one. And we can argue that all of our expressions operate on variables, which are stored in these slots. IIRC in the case of our palo alto pipeline, there were 37 FilterXScope instances which generated 37 allocations for FilterXScope and then the same amount for the GPtrArray instance and then the pointer array within that GPtrArray instance.

I don't remember how much this was worth in performance improvements, but I measured all of these changes and dropped those which didn't have a measurable impact.

I went back to some of my notes on Slack. I've found a relatively old profile where filterx_scope_unref() (with the associated cost in _freeing_ of the variables array) was responsible for 10.9% of the cycles while the one with the new layout it was 3.3%

Another comparison, all FilterXScope related operations are 7.2% on a new profile, and 33.7% on the old one.

lib/filterx/filterx-scope.c
bazsi: I am not sure how that could happen, we don't have dynamic allocation of variables, we do not have loops, etc. The only way to create 1M variables is to create a configuration with 1M lines, each defining a variable.

With that said, the usual stack size is 8M, a FilterXVariable is 16 bytes, but we need stack space for the actual function calls, and as we just discussed, FilterXScopes are embedded on the stack. What if we just limited the number of maximum variables any FilterXScope can support to 1024? Or maybe 4096? I don't think anyone would need more, or if they do, we can always bump that higher.

That should limit the stack use to 64kB

bazsi:
Rebased the branch, added the revert and reimplemented a small optimization that was part of the revert. I did not change any other things yet. I looked into adding the limit, but unfortunately most things that take care of variables basically assume we have an unlimited supply of them, errors are not handled. This is something I wouldn't start at 9pm.

lib/filterx/filterx-scope.h
bazsi: wrt naming, I kind of liked those as I worked with them (even though I do change names while working on a specific code, until I feel they are descriptive enough. I guess this is a matter of style. I don't mind renaming them to what you recommended (might be easier to understand for some), but I think these are good enough for merging, especially as this seems to fix a blocker bug and improves performance a lot.

lib/filterx/filterx-scope.c
MrAnno: Maybe I missed something, but I'm only talking about the maximum number of variables that we allocate on the stack. Heap-allocated variables could remain infinite in my opinion.

My suggested would be to rename and calculate `filterx_scope_variables_max` like this:
`filterx_scope_variables_max_inlined = MIN(self->variables.len, 1024)`

bazsi (commented):
THe new patches look good to me. Thanks @MrAnno

PR #469, opened 2025-01-24
filterx/unset-empties: avoid using a GList for keys to unset

bazsi:
Every g_list_append() is an extra allocation, so by avoiding GList altogether we can reduce the number of malloc() calls by the number of keys that we unset.


MrAnno (approved):

bshifter (approved):
looks good, unit tests are ok, light tests are ok, valgrind negative

PR #471, opened 2025-01-24
Filterx initial string cache

bazsi:
This PR implements the caching of trivial FilterXString instances (e.g. zero-length and single character numbers), which are commonly found in our log patterns as string values. Instead of allocating these all the time, use a cached instance.

This is "initial" string cache, as I have some pending patches to implement a larger table of strings that could be cached
similarly, but I got to make more testing with those to see how they impact performance.

### DIFF
--- a/lib/filterx/expr-getattr.c
+++ b/lib/filterx/expr-getattr.c
@@ -22,6 +22,7 @@
  */
 #include "filterx/expr-getattr.h"
 #include "filterx/object-string.h"
+#include "filterx/filterx-object-istype.h"
 #include "filterx/filterx-eval.h"
 #include "stats/stats-registry.h"
 #include "stats/stats-cluster-single.h"
@@ -132,7 +133,7 @@ _free(FilterXExpr *s)
 
 /* NOTE: takes the object reference */
 FilterXExpr *
-filterx_getattr_new(FilterXExpr *operand, FilterXString *attr_name)
+filterx_getattr_new(FilterXExpr *operand, FilterXObject *attr_name)
 {
   FilterXGetAttr *self = g_new0(FilterXGetAttr, 1);
 
@@ -146,7 +147,8 @@ filterx_getattr_new(FilterXExpr *operand, FilterXString *attr_name)
   self->super.free_fn = _free;
   self->operand = operand;
 
-  self->attr = (FilterXObject *) attr_name;
+  filterx_object_is_type(attr_name, &FILTERX_TYPE_NAME(string));
+  self->attr = attr_name;
   /* NOTE: name borrows the string value from the string object */
   self->super.name = filterx_string_get_value_ref(self->attr, NULL);
   return &self->super;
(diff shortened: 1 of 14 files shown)


lib/filterx/expr-getattr.c
```diff
@@ -146,7 +147,8 @@ filterx_getattr_new(FilterXExpr *operand, FilterXString *attr_name)
   self->super.free_fn = _free;
   self->operand = operand;
 
-  self->attr = (FilterXObject *) attr_name;
+  filterx_object_is_type(attr_name, &FILTERX_TYPE_NAME(string));
+  self->attr = attr_name;
```
MrAnno, line 151: Why are we reverting this?
This was needed for #252 to guarantee type safety at compile time.

The `filterx_object_is_type()` call here is a NOP.

bazsi: We started using FilterXString pointers in places where I think the argument being a string might easily be a temporary limitation.

getattr()/setattr() is better described with key having a FilterXObject type, even though today we only support string keys.

Because we started using FilterXString types, we  started duplicating functions:
* we had two constructors for FilterXString, one that returned FilterXString and one that returned FilterXObject, but the interfaces did not match otherwise. This I found confusing and I needed a string constructor that also had a length argument, which the string returning variant did not support
* sometimes we only implemented one or the other. E.g filterx_config_frozen_string() only was available as a variant returning FilterXString, but wasn't available in a format return FilterXObject. I could freeze a string in the grammar for interfaces where they expect a string instance with a FilterXObject pointer.

While I generally agree with type safety that the compiler can provide, but in this case the FilterXObject pointer means that we have a generic type that can be any object and as an interface, this better describes the intent with getattr/setattr, where we might want to provide other types of arguments as keys, like Python does.

This leaves us the interface of FilterXVariable, which can only indeed be string, but in that case a better solution
would be to just take a const gchar *, not even a FilterXString.

That was my thinking:
1) duplicate/redundant work on the string side, providing both types as interfaces, and I would have needed to change a FilterXString returning constructor
2) in the case of getattr/setattr the right interface is FilterXObject

I am trying to convert FilterXVariable to use a const gchar * instead.

bazsi: just pushed an implementation that takes out FilterXObject from FilterXVariableExpr and it indeed simplifies things a lot. It simplifies the call sites (test programs and the grammar), but it also hides an ugly implementation detail into the implementation instead of exposing it to callers. (the dollar formatting of variable names).

With this, we are losing the "freezing" of the string instance with variables, however we are basically doing the same by saving the string instance as a part of the FilterXVariableExpr instance and construction time.

We should be caching literal strings at the configuration level, that would actually help performance.

bazsi: the more I think if it, my argument around getattr/setattr needing to be generic on the attr_name seems to be bogus. I don't think an attr is ever going to be something that is not a string.

let's discuss this irl

MrAnno (approved):

PR #475, opened 2025-01-26
generic-number: make the gint64 path in the functions inline

bazsi:
We are encapsulating a GenericNumber instance in our integer/boolean/double objects, basically having two layers of encapsulation.

I attempted to remove one of these but it was a lot more difficult and would require API changes, so for now I was satisfied with inlining the integer paths, which is exercised whenever we are using an integer in filterx (e.g. iterating over a list in parse_csv()).

### DIFF
--- a/lib/generic-number.c
+++ b/lib/generic-number.c
@@ -22,26 +22,8 @@
 #include "generic-number.h"
 #include <math.h>
 
-gdouble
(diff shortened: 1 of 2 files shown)


bazsi:
Closing this in favour to #476 that directly builds on this.

PR #476, opened 2025-01-26
Filterx allow stack based integer wrappers

bazsi:
Similar to #470 but this time with integer objects. It seems that we have a parse_csv() invocation with over a 100 integer
fields, so the integer cache does not cover it and I don't want to increase that cache infinitely.

This should go in after #470


alltilla (approved):
Approve for the last 3 commits, let's merge the dependency PR, then we can merge this.

alltilla:
Rebased to main.

alltilla (approved):

PR #479, opened 2025-01-29
ci: schedule nightly only when there are new changes on main

MrAnno:
From now on, we keep the last 30 nightlies, but we only build nightly packages and images if we have new changes on main. Manually triggering the job will always allow rebuilds.

This will avoid a unnecessary storage and computation costs and also increases the nightly retention period.

### DIFF
--- a/.github/workflows/axosyslog-nightly.yml
+++ b/.github/workflows/axosyslog-nightly.yml
@@ -109,8 +109,8 @@ jobs:
         uses: actions/github-script@v7
         with:
           script: |
-            const daysToKeep = 30
-            const snapshotTagPattern = /_git[0-9]+/
+            const nightliesToKeep = 30
+            const nightlyTagPattern = /_git[0-9]+/
 
             const package_name = "axosyslog"
             const org = "axoflow"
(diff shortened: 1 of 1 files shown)


.github/workflows/axosyslog-nightly.yml
```diff
@@ -9,7 +9,24 @@ on:
 
 
 jobs:
+  precheck:
+    runs-on: ubuntu-latest
+    outputs:
+      nightly-needed: ${{ github.event_name != 'schedule' || fromJson(steps.previous-nightly.outputs.data).workflow_runs[0].head_sha != github.sha }}
```
OverOrion, line 15: I might be missing some GHA trickery here, but if I understand it correctly then for the firs run (when no previous hash is available) the failing property access (`.workflow_runs[0].head_sha`) will be converted into a `false` value?

OverOrion: For future reference: https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-repository

OverOrion (approved):

PR #480, opened 2025-01-30
str-utils: fix always true condition warning (-Wtype-limits)

MrAnno:
gchar is either signed or unsigned (127 may be the max value), guchar can be ASCII-checked, but then the 0-check is not needed.

### DIFF
--- a/lib/str-utils.h
+++ b/lib/str-utils.h
@@ -169,7 +169,7 @@ ch_isalpha(gchar c)
 static inline gboolean
 ch_isascii(gchar c)
 {
-  return c >= 0 && c <= 127;
+  return c >= 0 && (guchar ) c <= 127;
 }
 
 static inline gchar


lib/str-utils.h
```diff
@@ -167,9 +167,9 @@ ch_isalpha(gchar c)
 }
 
 static inline gboolean
-ch_isascii(gchar c)
+ch_isascii(guchar c)
 {
-  return c >= 0 && c <= 127;
+  return c <= 127;
```
bazsi, line 172: I expected the compiler to omit the superfluous condition, but I found it more reassuring to have the condition
anyway.

Changing the signature of the function I do not like that much, as C uses signed chars by default and everything is signed in  string handling.

I'd prefer an appropriate cast with this in mind.

bshifter (approved):

PR #481, opened 2025-01-30
cloud-auth: do no directly include picojson/picojson.h

alltilla:
There is a macro that modifies the number parsing behavior of picojson (PICOJSON_USE_INT64).

When jwt-cpp/jwt.h includes picojson.h it explicitly enables PICOJSON_USE_INT64. I did not check the code thoroughly, but it is possible that they explicitly build on some int64_t related functionality.

However if we directly include jwt.h without setting PICOJSON_USE_INT64, picojson will not have int64_t support. This resulted in a strange behavior of having int64_t type hinted value in a json object, but the type checker not recognizing it as a number.

### DIFF
--- a/modules/cloud-auth/azure-auth.hpp
+++ b/modules/cloud-auth/azure-auth.hpp
@@ -28,7 +28,7 @@
 #include "cloud-auth.hpp"
 
 #include <mutex>
-#include <picojson/picojson.h>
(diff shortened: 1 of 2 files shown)


sodomelle (approved):
Nice catch. Thank you for the help.

PR #483, opened 2025-01-30
Wallclocktime strftime

bazsi:
This PR pulls the strftime() implementation from musl and exposes it using the strftime() filterx function.

This adds '%f' and the %z/%Z behaviour similar to what strptime() can easily parse.

@jszigetvari

### DIFF
--- a/lib/timeutils/wallclocktime.c
+++ b/lib/timeutils/wallclocktime.c
@@ -54,7 +54,6 @@
 
 /*
  * MIT: musl http://git.musl-libc.org/cgit/musl/tree/src/time/strftime.c?id=6ad514e4e278f0c3b18eb2db1d45638c9af1c07f#n19
- * The license below only applies for the ISO8601 week calculation in this file.
  *
  */
 /*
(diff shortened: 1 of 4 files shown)


jszigetvari:
This would have been really useful for me today.

lib/timeutils/wallclocktime.c
```diff
@@ -1029,3 +1039,274 @@ dump_wall_clock_time(const WallClockTime *wct, GString *output)
...
+      f++;
+      pad = 0;
+      if (*f == '-' || *f == '_' || *f == '0') pad = *f++;
+      if ((plus = (*f == '+'))) f++;
+      width = (int) strtoul(f, &p, 10);
+      if (*p == 'C' || *p == 'F' || *p == 'G' || *p == 'Y')
+        {
+          if (!width && p!=f) width = 1;
+        }
+      else if (*p == 'f')
```
MrAnno, line 1259: Can we add the musl implementation without modifying it (marking their commit id in our commit message), and apply the modifications in a later patch so that we can review and follow upstream changes easier?

MrAnno (commented):
LGTM otherwise.

MrAnno (approved):

PR #484, opened 2025-01-31
Add repr to FilterX Otel objects

bshifter:
Axorouter encountered an error:
filterx: unable to repr; from='otel_kvlist', to='string'.

Upon investigating the issue, I found relevant hints in Axorouter’s configuration. The configuration suggests that OTel objects still require a repr method, but currently, a JSON-based workaround is used:
```
      filterx {
        # Fallback if map-to-text is not defined for the appliance.
        # TODO: OTel based types do not support string() cast, so we fallback for format_json().
        #       Remove that when OTel types start supporting string() cast.
        log.body = string(log.body) ?? format_json(log.body);
```

To address this, I implemented repr methods for all OTel-related FilterX objects:
- array
- kvlist
- logrecord
- resource
- scope

Additionally, I provided light tests to validate repr methods using forced string typecasting.

### DIFF
--- a/modules/grpc/otel/filterx/CMakeLists.txt
+++ b/modules/grpc/otel/filterx/CMakeLists.txt
@@ -5,6 +5,8 @@ endif()
 # C++ code
 
 set(OTEL_LOGRECORD_CPP_SOURCES
+  object-otel-base.hpp
+  object-otel-base.cpp
   object-otel.h
   object-otel-logrecord.cpp
   object-otel-logrecord.hpp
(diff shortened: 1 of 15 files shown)


OverOrion (commented):
`MessageToJsonString` [returns](https://protobuf.dev/reference/cpp/api-docs/google.protobuf.util.json_util/#MessageToJsonString.details) a `Status`. It might seem overly cautious, but could you please check it's value to make sure the output buffer's contents is valid?

OverOrion (approved):

PR #487, opened 2025-02-04
timeutils/wallclocktime: make %f completely optional

bazsi:
If the fraction of a second part does not start with a decimal dot or a digit, we just ignore it as if we parsed "0" usecs.

Comes with unit tests. @jszigetvari


jszigetvari (approved):
LGTM

MrAnno (approved):

PR #488, opened 2025-02-04
package-indexer: fix setting of value for integrity check

alltilla:
It seems like we need to set these on top level instead of in `client_context_params`.

This is now properly tested locally.


MrAnno (approved):
Thanks.

PR #489, opened 2025-02-05
github-actions: upload packages disable checksum validation

alltilla:
`aws` command tested locally with a private R2 instance.

I hope this is the last place we need to do this.


alltilla:
Test run: https://github.com/alltilla/axosyslog/actions/runs/13153469840

MrAnno (approved):

PR #491, opened 2025-02-05
filterx: stack allocated string followups

alltilla:
Followup for #470

### DIFF
--- a/lib/filterx/object-metrics-labels.c
+++ b/lib/filterx/object-metrics-labels.c
@@ -128,14 +128,17 @@ _set_subscript(FilterXDict *s, FilterXObject *key, FilterXObject **new_value)
   if (!name_str || !value_str)
     return FALSE;
 
-  if (!filterx_object_is_type(*new_value, &FILTERX_TYPE_NAME(string)))
-    *new_value = filterx_string_new(value_str, -1);
-
   StatsClusterLabel label = stats_cluster_label(g_strndup(name_str, name_len), g_strndup(value_str, value_len));
   g_array_append_val(self->labels, label);
 
   self->sorted = FALSE;
   self->deduped = FALSE;
+
+  if (!filterx_object_is_type(*new_value, &FILTERX_TYPE_NAME(string)))
+    {
+      filterx_object_unref(*new_value);
+      *new_value = filterx_string_new(value_str, value_len);
+    }
 
   scratch_buffers_reclaim_marked(marker);
   return TRUE;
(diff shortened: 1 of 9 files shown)


lib/filterx/object-metrics-labels.c
```diff
@@ -128,15 +128,18 @@ _set_subscript(FilterXDict *s, FilterXObject *key, FilterXObject **new_value)
...
 
   self->sorted = FALSE;
   self->deduped = FALSE;
 
+  if (!filterx_object_is_type(*new_value, &FILTERX_TYPE_NAME(string)))
+    {
+      filterx_object_unref(*new_value);
+      *new_value = filterx_string_new(value_str, value_len);
+    }
+
```
bazsi, line 142: I don't follow this a 100%, why do we need to update the value that we got via set_subscript()?

bazsi (commented):
This looks good to me, I don't understand the first patch entirely.

lib/filterx/object-metrics-labels.c
alltilla: If you have a `json_object` typed object and set one of its fields with for example an `otel_kvlist`, the `json_object` will implicitly convert the `otel_kvlist` to a `json_object` and store that.

Let's imagine the user-side without modifying the pointer to `new_value`:
```c
FilterXObject *fx_json_object = <...>;
FilterXObject *fx_some_dict = <...>;

g_assert(filterx_object_set_subscript(fx_json_object, <some_key>, fx_some_dict));

// We expect to change fx_json_object[<some_key>], but fx_some_dict is not the same as fx_json_object[<some_key>].
// If we want to modify fx_json_object[<some_key>], we need to do a new lookup.
filterx_object_set_subscript(fx_some_dict, <some_other_key>, <some_value>);
```

If we make `new_value` an in&out argument and make it point to the converted value after processing, the code above works and we save a lookup.

As far as I remember, the only place where the pattern above is used is at filterx based parsers. I cannot remember exactly which parser's which part needed this, but I'm sure this was a problem at some point.

If there was a conversion, the pre-conversion `new_value` must be released and it is the responsibility of `set_subscript()`.

This is a rather convoluted thing and we can mess it up in every `set_subscript()` implementation, this is why I spent a lot of time validating that `new_value` can be a stack allocated string, but I think it is okay.

alltilla:
Rebased to main.

alltilla:
@bazsi Can you re-review this please? All the dependency PRs are merged.

lib/filterx/object-metrics-labels.c
bazsi: In this specific case, I don't think the converted value can be used for further modification, but I understand that for the sake mimicking what the other dict-like objects do, this is ok.

PR #492, opened 2025-02-06
filterx: add assertion on generation counter change in filterx_scope_…

bazsi:
…set_message()

The message generation counter is embedded in FilterXVariable instances that are stored in the scope. We assume that those variables are in sync with the message under processing and a similar assertion exists in filterx_scope_sync().

The assertion in filterx_scope_sync() fired, but at that point it is difficult to know where the generation counter got out of sync.  One such point is in filterx_scope_set_message(), hopefully that's closer to the root cause.


bazsi:
This is to help the diagnosis of the bug reported by @mitzkia

MrAnno (approved):

PR #496, opened 2025-02-07
filterx: fix racy writing of readonly JSON object caches 

MrAnno:
This may not be the best way to fix the race, but this whole caching mess will be reimplemented/removed soon.

### DIFF
--- a/lib/filterx/object-json-object.c
+++ b/lib/filterx/object-json-object.c
@@ -140,9 +140,6 @@ _set_subscript(FilterXDict *s, FilterXObject *key, FilterXObject **new_value)
   if (!filterx_object_map_to_json(*new_value, &jso, &assoc_object))
     return FALSE;
 
-  if (!(*new_value)->readonly)
-    filterx_json_associate_cached_object(jso, assoc_object);
-
   if (json_object_object_add(self->jso, key_str, jso) != 0)
     {
       filterx_object_unref(assoc_object);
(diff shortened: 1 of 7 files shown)


lib/filterx/object-json-object.c
```diff
@@ -140,7 +140,8 @@ _set_subscript(FilterXDict *s, FilterXObject *key, FilterXObject **new_value)
   if (!filterx_object_map_to_json(*new_value, &jso, &assoc_object))
     return FALSE;
 
-  filterx_json_associate_cached_object(jso, assoc_object);
+  if (!(*new_value)->readonly)
+    filterx_json_associate_cached_object(jso, assoc_object);
```
alltilla: If I understand correctly with this change we will never cache the associated `FilterXObject`s of inner layers of the top level `json_object` typed `FilterXObject`. If this is true this kind of defeats half the purpose of the caching (parsing itself is cached, but we create `FilterXObject`s over and over again for each lookup).

Can't we build a deep `FilterXObject` cache instead in `cache_json_file()` once? I think it would make sure that we do not fill the inner cache from multiple threads at lookup time and would still make this performant.

In other words, I think the lazy `FilterXObject` cache buildup is the problem, which can be solved either with never building a cache or building it not-lazy, too, and I would prefer the latter.

alltilla (commented):
Nice catch!

lib/filterx/object-json-object.c
MrAnno: I think we already populate the cache non-lazily when deep-freezing `cache_json_file()`.

Hopefully this entire mess will be removed with the new dict implementation.

alltilla: Yes, I see now, thanks!

bazsi (approved):

PR #499, opened 2025-02-08
filterx: fix compilation error on 32-bit platforms (gsize vs. guint64)

MrAnno:

### DIFF
--- a/lib/filterx/func-unset-empties.c
+++ b/lib/filterx/func-unset-empties.c
@@ -372,14 +372,13 @@ static gboolean
 _handle_target_object(FilterXFunctionUnsetEmpties *self, FilterXObject *target, GError **error)
 {
   g_assert(target);
-  guint64 len;
-
   target = filterx_ref_unwrap_ro(target);
 
   if (filterx_object_is_type(target, &FILTERX_TYPE_NAME(null)))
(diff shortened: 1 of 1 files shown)


lib/filterx/func-unset-empties.c
```diff
@@ -402,6 +402,7 @@ _handle_target_object(FilterXFunctionUnsetEmpties *self, FilterXObject *target,
     }
   else if (filterx_object_is_type(target, &FILTERX_TYPE_NAME(string)))
     {
+      gsize len;
```
bazsi, line 405: isn't this triggering a shadowed name issue?

bazsi: it does, that's why the compilation failed.

MrAnno: Indeed. I've had too much Rust it seems.

MrAnno: We should probably unify this, gsize would be a good candidate to represents length.

bazsi: I agree, filterx_object_len() could return gsize, but I am merging this as is.

bazsi (approved):

PR #504, opened 2025-02-12
Fix `vars()`

MrAnno:
Follow-up for 145c50840708323e3f5f5bffff27f5ac72c1766a

If the GHashTable allocation appears to be costly, this can be reimplemented using a fix-sized heapqueue and a "merge" operation on the sorted variables arrays.

### DIFF
--- a/lib/filterx/filterx-scope.c
+++ b/lib/filterx/filterx-scope.c
@@ -248,7 +248,7 @@ filterx_scope_register_variable(FilterXScope *self,
 }
 
 gboolean
-filterx_scope_foreach_variable(FilterXScope *self, FilterXScopeForeachFunc func, gpointer user_data)
+filterx_scope_foreach_variable_readonly(FilterXScope *self, FilterXScopeForeachFunc func, gpointer user_data)
 {
   FilterXVariable *variables = _get_variable_array(self);
 
--- a/lib/filterx/filterx-scope.c
+++ b/lib/filterx/filterx-scope.c
@@ -250,23 +250,40 @@ filterx_scope_register_variable(FilterXScope *self,
 gboolean
 filterx_scope_foreach_variable_readonly(FilterXScope *self, FilterXScopeForeachFunc func, gpointer user_data)
 {
-  FilterXVariable *variables = _get_variable_array(self);
+  gboolean result = TRUE;
+  GHashTable *already_found = g_hash_table_new(g_direct_hash, g_direct_equal);
 
-  for (gint i = 0; i < self->variables.len; i++)
+  for (FilterXScope *scope = self; scope; scope = scope->parent_scope)
     {
-      FilterXVariable *variable = &variables[i];
+      FilterXVariable *variables = _get_variable_array(scope);
+      for (gint i = 0; i < scope->variables.len; i++)
+        {
+          FilterXVariable *variable = &variables[i];
 
-      if (!variable->value)
-        continue;
+          if (!variable->value)
+            continue;
 
-      if (!_validate_variable(self, variable))
-        continue;
+          /* NOTE: we validate against @self */
+          if (!_validate_variable(self, variable))
+            continue;
 
-      if (!func(variable, user_data))
-        return FALSE;
+          if (g_hash_table_contains(already_found, GSIZE_TO_POINTER(variable->handle)))
+            continue;
+
+          /* objects are not pulled into the current scope (not cloned), so this is a readonly view */
+          if (!func(variable, user_data))
+            {
+              result = FALSE;
+              goto exit;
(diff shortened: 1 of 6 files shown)


MrAnno:
I'm writing a test as well.

lib/filterx/filterx-scope.c
```diff
@@ -248,7 +248,7 @@ filterx_scope_register_variable(FilterXScope *self,
 }
 
 gboolean
-filterx_scope_foreach_variable(FilterXScope *self, FilterXScopeForeachFunc func, gpointer user_data)
+filterx_scope_foreach_variable_readonly(FilterXScope *self, FilterXScopeForeachFunc func, gpointer user_data)
```
bazsi, line 255: I'm fine with the rename, however for each style functions often have the restriction that the object can't be modified, so I don't understand why this is special.

lib/filterx/filterx-scope.c
```diff
@@ -250,23 +250,40 @@ filterx_scope_register_variable(FilterXScope *self,
 gboolean
 filterx_scope_foreach_variable_readonly(FilterXScope *self, FilterXScopeForeachFunc func, gpointer user_data)
 {
-  FilterXVariable *variables = _get_variable_array(self);
+  gboolean result = TRUE;
+  GHashTable *already_found = g_hash_table_new(g_direct_hash, g_direct_equal);
```
bazsi, line 258: Using a hash table here should be fine, vars() shouldn't be used for performance intensive code.

I was thinking how this could be avoided, but no good ideas.

lib/filterx/tests/test_scope.c
```diff
@@ -0,0 +1,107 @@
...
+  cr_assert_null(filterx_scope_lookup_variable(s, filterx_map_varname_to_handle("var", FX_VAR_DECLARED_FLOATING)));
+  filterx_scope_free(s);
+}
+
+Test(filterx_scope, test_scope_stacking)
+{
+  gsize alloc_size = filterx_scope_get_alloc_size();
+
+  FilterXScope *s = g_alloca(alloc_size);
+  filterx_scope_init_instance(s, filterx_scope_get_alloc_size(), NULL);
```
bazsi, line 49: This could use the same alloc_size variable we declared above

lib/filterx/tests/test_scope.c
```diff
@@ -0,0 +1,107 @@
...
+Test(filterx_scope, test_scope_sync)
+{
+  gsize alloc_size = filterx_scope_get_alloc_size();
+  FilterXScope *s = g_alloca(alloc_size);
+  filterx_scope_init_instance(s, filterx_scope_get_alloc_size(), NULL);
+  LogMessage *msg = log_msg_new_empty();
+
+  filterx_scope_set_message(s, msg);
+
+  FilterXVariableHandle var = filterx_map_varname_to_handle("$var", FX_VAR_MESSAGE_TIED);
```
bazsi, line 72: Iirc I removed the need to start these names with a dollar. Actually this works as if you used $$var in the config

lib/filterx/filterx-scope.c
```diff
@@ -180,9 +180,13 @@ filterx_scope_lookup_variable(FilterXScope *self, FilterXVariableHandle handle)
...
 
-  if (_lookup_variable(self, handle, &v, &v_index) &&
-      _validate_variable(self, v))
-    return v;
+  if (_lookup_variable(self, handle, &v, &v_index))
+    {
+      if (_validate_variable(self, v))
+        return v;
+
+      return NULL;
```
bazsi, line 188: Good catch

lib/filterx/filterx-scope.c
MrAnno: The arrays are sorted, they could be merged using a heap, but I think this looks simpler this way.

lib/filterx/tests/test_scope.c
MrAnno: `filterx_map_varname_to_handle()` removed the `$`.

bazsi: Maybe that patch is not yet merged though

bazsi (dismissed):

lib/filterx/filterx-scope.c
MrAnno: `foreach()` functions usually allow storing or modifying the returned objects. Since these come from multiple scopes, neither is allowed.

lib/filterx/tests/test_scope.c
MrAnno: Fixed, thank you.

bazsi (approved):

PR #505, opened 2025-02-12
`vars()`: add `exclude_message_values` parameter

MrAnno:
Depens on #504

Note: `vars()` lies a little, because it formats both floating and message-tied variables, but NOT ALL message variables as they might have never appeared inside FilterX.


bazsi (dismissed):
Looks good. I was thinking if we needed to exclude non declared variables how would that look?

What if the option was exclude_message_values or something. This is quite ugly as a name, my point is that we probably only need to include declared, floating values

MrAnno:
Let's go with `exclude_message_values`. We can add `exclude_local` or `exclude_non_declared` later.

bazsi:
This looks good, albeit we would need to rebase it against main as #504 was merged.

bazsi (approved):

alltilla (approved):

PR #510, opened 2025-02-14
Filterx set timestamp func

sodomelle:
Example usage:
```
set_timestamp(datetime, "stamp");
```

### DIFF
--- a/lib/filterx/func-set-timestamp.c
+++ b/lib/filterx/func-set-timestamp.c
@@ -23,10 +23,19 @@
  */
 
 #include "func-set-timestamp.h"
+#include "object-datetime.h"
+#include "filterx/object-extractor.h"
+#include "filterx/object-primitive.h"
+#include "filterx/filterx-eval.h"
+
+#define FILTERX_FUNC_SET_TIMESTAMP_USAGE "Usage: set_timestamp(datetime, stamp=[\"stamp\", \"recvd\"])"
 
 typedef struct FilterXFunctionSetTimestamp_
 {
   FilterXFunction super;
+
+  FilterXExpr *datetime_expr;
+  LogMessageTimeStamp timestamp_idx;
 } FilterXFunctionSetTimestamp;
 
 static FilterXObject *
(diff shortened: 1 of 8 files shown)


lib/filterx/func-set-timestamp.c
```diff
@@ -0,0 +1,173 @@
...
+      return FALSE;
+    }
+
+  self->datetime_expr = _extract_set_timestamp_datetime_expr(args, error);
+  if (!self->datetime_expr)
+    {
+      return FALSE;
+    }
+
+  gchar *idx_str = g_strdup(_extract_set_timestamp_idx_str(args, error));
```
alltilla, line 127: I think this is leaking. Also we don't need to dup it, we don't store it anywhere.

lib/filterx/func-set-timestamp.c
```diff
@@ -0,0 +1,173 @@
...
+  if (!idx_str)
+    {
+      return FALSE;
+    }
+
+  if (g_str_equal(idx_str, "ts_stamp"))
```
alltilla, line 133: I think `ts_` would just confuse the user, what do you think about accepting `"stamp"` or `"recvd"`?
`date-parser()` does so, and uses this function, which can be useful here, too: https://github.com/axoflow/axosyslog/blob/a0a400f24b8ff0c556a69341fd0fc75a64d48636/lib/logmsg/logmsg.c#L2103

lib/filterx/func-set-timestamp.c
```diff
@@ -0,0 +1,173 @@
...
+  if (g_str_equal(idx_str, "ts_recvd"))
+    {
+      self->timestamp_idx = LM_TS_RECVD;
+      return TRUE;
+    }
+  if (g_str_equal(idx_str, "ts_processed"))
+    {
+      self->timestamp_idx = LM_TS_PROCESSED;
+      return TRUE;
+    }
```
alltilla, line 147: I think most of the time the user wants to set the `stamp` timestamp, maybe sometimes `recvd`. Setting `processed` is never needed IMO, that is a disk-buffer related strange timestamp.

`date-parser` defaults to `stamp` and can be set to `recvd`, but not `processed`, we could follow that.
(Using `log_msg_lookup_time_stamp_name()` would implicitly do so)

Doing so would make this a named/optional parameter, if we follow `date-parser()` it could be called `time_stamp`, but I think we can challenge that if we are not fully satisfied with it.

lib/filterx/func-set-timestamp.c
```diff
@@ -0,0 +1,173 @@
...
+#include "filterx/filterx-eval.h"
+
+#define FILTERX_FUNC_SET_TIMESTAMP_USAGE "Usage: set_timestamp(datetime, [ts_stamp, ts_recvd, ts_processed])"
+
+typedef struct FilterXFunctionSetTimestamp_
+{
+  FilterXFunction super;
+
+  FilterXExpr *datetime_expr;
+  gint32 timestamp_idx;
```
alltilla, line 38: Can we store this with explicit type `LogMessageTimeStamp`?

lib/filterx/func-set-timestamp.c
sodomelle: Done.

lib/filterx/func-set-timestamp.c
sodomelle: Done.

lib/filterx/func-set-timestamp.c
sodomelle: Yep, my bad. Done.

lib/filterx/func-set-timestamp.c
```diff
@@ -40,14 +49,79 @@ _set_timestamp_free(FilterXExpr *s)
...
+      self->timestamp_idx = LM_TS_RECVD;
+      return TRUE;
+    }
+  if (g_str_equal(idx_str, "ts_processed"))
+    {
+      self->timestamp_idx = LM_TS_PROCESSED;
+      return TRUE;
+    }
+
+  return FALSE;
```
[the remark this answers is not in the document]
sodomelle: Done.

lib/filterx/func-set-timestamp.c
```diff
@@ -40,13 +49,75 @@ _set_timestamp_free(FilterXExpr *s)
...
 
+static const gchar *
+_extract_set_timestamp_idx_str(FilterXFunctionArgs *args, GError **error)
+{
+  gboolean exists;
+  const gchar *idx_str = filterx_function_args_get_named_literal_string(args, "stamp", NULL, &exists);
+
+  if (!exists)
+    {
+      return g_strdup("stamp");
```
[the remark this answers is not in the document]
sodomelle: Done.

lib/filterx/func-set-timestamp.c
```diff
@@ -23,10 +23,19 @@
  */
 
 #include "func-set-timestamp.h"
+#include "object-datetime.h"
+#include "filterx/object-extractor.h"
+#include "filterx/object-primitive.h"
+#include "filterx/filterx-eval.h"
+
+#define FILTERX_FUNC_SET_TIMESTAMP_USAGE "Usage: set_timestamp(datetime, stamp=[stamp, recvd])"
```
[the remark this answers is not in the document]
sodomelle: Done.

alltilla (approved):

PR #511, opened 2025-02-14
Partially fix FilterX heap corruption

MrAnno:
The race condition around json-c `get`/`put` is unsolved, but this could go in separately if reviewers agree.

### DIFF
--- a/modules/json/filterx-cache-json-file.c
+++ b/modules/json/filterx-cache-json-file.c
@@ -205,6 +205,8 @@ _deep_freeze(FilterXFuntionCacheJsonFile *self, FilterXObject *object)
   if (filterx_object_freeze(object))
     g_ptr_array_add(self->frozen_objects, object);
 
+  filterx_object_make_readonly(object);
+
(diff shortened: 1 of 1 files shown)


bazsi (approved):
As I understand make_readonly() happened at runtime without this patch when
getattr() encountered a readonly parent.

I did not try to reproduce the problem, but the fix looks legit.

bazsi:
not merging as this is in the draft state.

MrAnno:
I believe the runtime `make_readonly()` call didn't happen as we used an alternative approach (`json_object_object_foreachC()`) while deep freezing.

bshifter (dismissed):
i remove it from my PR then ;)

MrAnno:
Okay, we discussed IRL these will go in separately due to their complexity.

MrAnno:
The race condition around json-c `get`/`put` is unsolved, but this could go in separately if reviewers agree.

bshifter (approved):

PR #512, opened 2025-02-16
Filterx stack based string allocation fixes

bazsi:
There are two patches in this PR that fix two distinct issues:
* when moving filterx_string_get_value_ref() to be inline, I removed the check for NUL termination, as all strings are NUL terminated, when constructed normally. Unfortunately stack allocated ones may rely on the explicit length without NUL termination, such as the ones in get_sdata() function.
* sometimes "-1" as an extremal value to the stack allocation macro comes in an unsigned variable, causing the check in the macro not to fire. Avoid these cases by an explicit cast.

### DIFF
--- a/lib/filterx/object-string.h
+++ b/lib/filterx/object-string.h
@@ -95,7 +95,7 @@ void filterx_string_global_deinit(void);
   { \
     FILTERX_OBJECT_STACK_INIT(string), \
     .str = (cstr), \
-    .str_len = ((cstr_len) < 0 ? strlen(cstr) : (cstr_len)), \
+    .str_len = (((gssize) cstr_len) == -1 ? strlen(cstr) : (cstr_len)), \
   }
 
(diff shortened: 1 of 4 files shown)


lib/filterx/object-string.h
```diff
@@ -95,7 +95,7 @@ void filterx_string_global_deinit(void);
   { \
     FILTERX_OBJECT_STACK_INIT(string), \
     .str = (cstr), \
-    .str_len = ((cstr_len) < 0 ? strlen(cstr) : (cstr_len)), \
+    .str_len = (((gssize) cstr_len) < 0 ? strlen(cstr) : (cstr_len)), \
```
MrAnno, line 98: Shouldn't we check for an explicit `-1` here?

Hopefully our strings will never reach `~2^31`, because all subsequent value would be interpreted as negative on a 32-bit platform, but a `-1` check would still be more concrete.

MrAnno (approved):

PR #515, opened 2025-02-16
filterx/object-primitive: fix double repr method to generate JSON like result

bazsi:
We are using repr to render doubles in format_json() (that I am going to use a lot more in subsequent patches).

This makes sure that the rendered double actually matches  JSON expectations.

Our double repr was different to JSON a number of ways:
  * in JSON we need a ".0" even if we have an integer value to indicate that a member is floating point
  * we should not include trailing zeroes
  * we should use Infinite and NaN tokens to indicate these values


lib/filterx/object-primitive.c
```diff
@@ -139,12 +139,54 @@ _double_add(FilterXObject *s, FilterXObject *object)
...
+      snprintf(buf, sizeof(buf), "%.17g", dbl);
+
+      /* find fractions, either '.' or ',' can separate integer and fraction
+       * part */
+      dot = strchr(buf, '.');
+      if (!dot)
+        {
+          dot = strchr(buf, ',');
+          if (dot)
+            *dot = '.';
```
MrAnno, line 168: `g_ascii_dtostr()` uses `"%.17g"` and has no locale problems (comma).

lib/generic-number.h
```diff
@@ -104,6 +104,16 @@ gn_is_nan(const GenericNumber *number)
...
 }
 
+gboolean _gn_is_inf(const GenericNumber *number);
+
+static inline gboolean
+gn_is_inf(const GenericNumber *number)
+{
+  if (number->type != GN_DOUBLE)
+    return FALSE;
+  return _gn_is_nan(number);
```
MrAnno, line 114: Typo

lib/filterx/object-primitive.c
```diff
@@ -139,12 +139,46 @@ _double_add(FilterXObject *s, FilterXObject *object)
...
-  g_string_set_size(repr, init_len + strlen(repr->str + init_len));
+  if (isnan(dbl))
+    {
+      g_string_append_len(repr, "NaN", 3);
+    }
+  else if (isinf(dbl))
+    {
+      if (dbl < 0)
+        g_string_append_c(repr, '-');
+      g_string_append_len(repr, "Infinity", 8);
```
MrAnno, line 152: Both JSON RFCs says:
```
Numeric values that cannot be represented in the grammar below (such as Infinity and NaN) are not permitted.
```

A few languages/libraries serialize these values incorrectly.
If we want to represent these, I think we need to use strings instead.

lib/generic-number.h
MrAnno: The typo has reappeared.

lib/filterx/object-primitive.c
MrAnno: I pushed the `g_ascii_dtostr()` version but the rebase made it disappear.

lib/generic-number.h
bazsi: good catch. this is an unused function, so I am dropping that patch from the series.

Originally this was used to detect infinite values, but then this would have been difficult due to layering issues.

lib/filterx/object-primitive.c
bazsi: sorry, for dropping your change, did not notice you pushed another version. let me redo that.

lib/filterx/object-primitive.c
bazsi: snprintf() also produces "inf" in case the result is infinite.

MrAnno: We discussed that there is no good way to resolve this issue. A few programming languages and libraries decided to violate the RFC as well because in very generic cases, there is no good choice in such cases: encode these values as strings? skip the keys that hold such values?

MrAnno (approved):

PR #516, opened 2025-02-16
Filterx add object str method

bazsi:
This branch separates repr() and str() methods for objects. This works similarly to Python.

The goal of this patch is to resolve an issue around datetime, which is formatted differently when embedded in a JSON (where map_to_json is used) and when formatted using format_json (where repr is used).

The branch also changes a number of call sites to use str() instead of repr(). Convention is:
* whenever we display an object to the developer, we should use repr()
* whenever we include an object in some form of output message, we should use str()

By default, str == repr so the formats can be the same.


lib/filterx/filterx-ref.c
```diff
@@ -195,10 +195,15 @@ _filterx_ref_repr(FilterXObject *s, GString *repr)
 {
   FilterXRef *self = (FilterXRef *) s;
 
-  if (self->value->type->repr)
-    return self->value->type->repr(self->value, repr);
+  return filterx_object_repr(self->value, repr);
```
MrAnno, line 198: This breaks `append()`, because `filterx_object_repr()` truncates the string.

lib/filterx/filterx-ref.c
```diff
@@ -195,10 +195,15 @@ _filterx_ref_repr(FilterXObject *s, GString *repr)
...
+}
 
-  return FALSE;
+static gboolean
+_filterx_ref_str(FilterXObject *s, GString *str)
+{
+  FilterXRef *self = (FilterXRef *) s;
+
+  return filterx_object_str(self->value, str);
```
MrAnno, line 206: We should call the virtual method directly to avoid the truncation.

lib/filterx/filterx-ref.c
bazsi: I noticed this in a follow up patch and now fixed it.

lib/filterx/filterx-ref.c
bazsi: did both for this and the repr() case as well.

lib/filterx/func-repr.c
```diff
@@ -0,0 +1,46 @@
...
+#include "scratch-buffers.h"
+
+FilterXObject *
+filterx_simple_function_repr(FilterXExpr *s, FilterXObject *args[], gsize args_len)
+{
+  FilterXObject *object = filterx_typecast_get_arg(s, args, args_len);
+  if (!object)
+    return NULL;
+
+  GString *buf = scratch_buffers_alloc();
```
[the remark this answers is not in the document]
bazsi: This is going to be reclaimed by the GC, I usually help the GC with an explicit mark and restore in case I am generating a lot of garbage (e.g. loops with potentially large number of items for example).

Adding that would not hurt, but not adding it doesn't hurt that much either, IMHO.

MrAnno (approved):

lib/filterx/func-repr.c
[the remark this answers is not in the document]
MrAnno: Our GC is manual, I just didn't know when exactly this will be reclaimed.

PR #517, opened 2025-02-17
filterx: cache_json_file() inotify based automatic reload

bshifter:
Due to multiple workers, the original frozen objects must be retained until all workers have finished. Since there is currently no way to track the workers' lifecycle, we call reload, which unfreezes and releases all previously stored elements to prevent high memory usage. This is a suboptimal solution and likely temporary. The upcoming RCU feature will provide a better way to handle this issue. However, the soon-to-be-released native dictionary implementation will also support the optimal solution.

### DIFF
--- a/lib/file-monitor.h
+++ b/lib/file-monitor.h
@@ -1,5 +1,7 @@
 /*
+ * Copyright (c) 2024 Axoflow
  * Copyright (c) 2023 László Várady
+ * Copyright (c) 2024 shifter
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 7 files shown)


lib/mainloop.c
```diff
@@ -167,7 +167,7 @@ struct _MainLoop
 
   MainLoopOptions *options;
   ControlServer *control_server;
-  CfgMonitor *cfg_monitor;
+  FileMonitor *file_monitor;
```
MrAnno, line 170: Can we keep the instance name `cfg_monitor` in this unit?

lib/file-monitor.h
```diff
@@ -21,33 +23,35 @@
...
 
-void cfg_monitor_add_watch(CfgMonitor *self, CfgMonitorEventCB cb,  gpointer cb_data);
-void cfg_monitor_remove_watch(CfgMonitor *self, CfgMonitorEventCB cb, gpointer cb_data);
+void file_monitor_add_watch(FileMonitor *self, FileMonitorEventCB cb,  gpointer cb_data);
+void file_monitor_remove_watch(FileMonitor *self, FileMonitorEventCB cb, gpointer cb_data);
 
-void cfg_monitor_start(CfgMonitor *self);
-void cfg_monitor_stop(CfgMonitor *self);
+void file_monitor_start(FileMonitor *self);
+void file_monitor_start_silent(FileMonitor *self);
```
MrAnno, line 54: It does not look like a "silent" version of start, it's more like whether you want an "immediate" check or not.

What do you think of `start_immediate()` or `start_and_check()`?
Don't you need a `bool` return value for the event callback? In that case `start_and_check()` could return that as well.

modules/json/filterx-cache-json-file.c
```diff
@@ -209,36 +250,75 @@ _deep_freeze(FilterXFuntionCacheJsonFile *self, FilterXObject *object)
 
   object = filterx_ref_unwrap_ro(object);
   if (filterx_object_is_type(object, &FILTERX_TYPE_NAME(json_object)))
-    _deep_freeze_dict(self, object);
+    _deep_freeze_dict(self, object, root_object);
```
MrAnno, line 253: This commit contains multiple changes:
- renaming types
- refactoring function signatures

modules/json/filterx-cache-json-file.c
```diff
@@ -53,13 +57,28 @@ cache_json_file_error_quark(void)
...
+  FileMonitor *file_monitor;
+} FilterXFunctionCacheJsonFile;
+
+static void
+_archive_frozen_objects(FilterXFunctionCacheJsonFile *self)
+{
+  if (!self->frozen_objects)
+    return;
+  self->frozen_objects_history[self->history_index] = self->frozen_objects;
+  self->history_index++;
```
MrAnno, line 79: Can we add an assert not to overflow the buffer in case the cleanup mechanism fails or someone refactors the guarantees away?

modules/json/filterx-cache-json-file.c
```diff
@@ -209,36 +250,75 @@ _deep_freeze(FilterXFuntionCacheJsonFile *self, FilterXObject *object)
...
+    _deep_freeze_list(self, object, root_object);
+}
+
+gboolean
+_load_json_file_version(FilterXFunctionCacheJsonFile *self, GError **error)
+{
+  main_loop_assert_main_thread();
+  if (self->history_index >= FROZEN_OBJECTS_HISTORY_SIZE)
+    {
+      main_loop_reload_config(main_loop_get_instance());
```
MrAnno, line 265: Can we move out this condition and the reload call to `_file_monitor_callback` (the name `_load_json_file_version` does not suggest that such intrusive side-effect might happen)?
Once it is done, can we add a comment to `_file_monitor_callback()` that this function may reload the configuration, so nothing complicated should be done with config-related references on the caller side?

MrAnno:
Nice work! :)

lib/file-monitor.h
```diff
@@ -21,33 +23,35 @@
...
 
-void cfg_monitor_add_watch(CfgMonitor *self, CfgMonitorEventCB cb,  gpointer cb_data);
-void cfg_monitor_remove_watch(CfgMonitor *self, CfgMonitorEventCB cb, gpointer cb_data);
+void file_monitor_add_watch(FileMonitor *self, FileMonitorEventCB cb,  gpointer cb_data);
+void file_monitor_remove_watch(FileMonitor *self, FileMonitorEventCB cb, gpointer cb_data);
 
-void cfg_monitor_start(CfgMonitor *self);
-void cfg_monitor_stop(CfgMonitor *self);
+void file_monitor_start_and_check(FileMonitor *self);
+void file_monitor_start_immediate(FileMonitor *self);
```
MrAnno, line 54: I'm confused, now both are named like they do an immediate check. Shouldn't one of them just be `start`?

MrAnno (approved):

PR #518, opened 2025-02-18
filterx: support optimized strings in `regexp_search()` and `regexp_subst()` pattern

alltilla:
I don't think we need a news file entry for this.


OverOrion (approved):
Thanks for this! :fire: :rocket:

bazsi:
we should really try to centralize the iteration on embedded expr instances. something that can be reused for init/deinit/optimize purposes. but that's just something for the future.

PR #521, opened 2025-02-19
Filterx set pri func

sodomelle:
Example usage:
```
set_pri(pri="100");
```

Note: Second argument must be between 0 and 191 inclusive.

### DIFF
--- a/lib/filterx/func-set-pri.c
+++ b/lib/filterx/func-set-pri.c
@@ -23,10 +23,17 @@
  */
 
 #include "func-set-pri.h"
+#include "filterx/object-extractor.h"
+#include "filterx/object-primitive.h"
+#include "filterx/filterx-eval.h"
+
+#define FILTERX_FUNC_SET_PRI_USAGE "Usage: set_pri([value])"
 
 typedef struct FilterXFunctionSetPri_
 {
   FilterXFunction super;
+
+  FilterXExpr *pri_expr;
 } FilterXFunctionSetPri;
 
 static FilterXObject *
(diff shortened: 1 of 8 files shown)


lib/filterx/func-set-pri.c
```diff
@@ -0,0 +1,105 @@
...
+
+  filterx_function_free_method(&self->super);
+}
+
+static gboolean
+_extract_set_pri_arg(FilterXFunctionSetPri *self, FilterXFunctionArgs *args, GError **error)
+{
+  gboolean exists;
+  gboolean arg_error;
+  self->pri = filterx_function_args_get_named_literal_integer(args, "pri", &exists, &arg_error);
```
alltilla, line 65: This should be stored as an expr and be evaluated and used in `eval()` to support setting pri with a message related value instead of a hard coded literal value.

lib/filterx/func-set-pri.c
```diff
@@ -0,0 +1,105 @@
...
+
+/* Takes reference of args */
+FilterXExpr *
+filterx_function_set_pri_new(FilterXFunctionArgs *args, GError **error)
+{
+  FilterXFunctionSetPri *self = g_new0(FilterXFunctionSetPri, 1);
+  filterx_function_init_instance(&self->super, "set_pri");
+
+  self->super.super.eval = _set_pri_eval;
+  self->super.super.free_fn = _set_pri_free;
```
alltilla, line 92: Let's also write `init`, `deinit` and `optimize` functions.

(We should do the same for `set_timestamp()` too, I forgot about that during reviewing that PR.)

lib/filterx/func-set-pri.c
sodomelle: Done.

lib/filterx/func-set-pri.c
sodomelle: Done.

lib/filterx/func-set-pri.c
```diff
@@ -0,0 +1,147 @@
...
+                              NULL);
+      return NULL;
+    }
+
+  FilterXEvalContext *context = filterx_eval_get_context();
+  LogMessage *msg = context->msg;
+
+  msg->pri = pri_val;
+
+  return filterx_boolean_new(TRUE);
```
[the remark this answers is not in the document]
sodomelle: Done. Sorry :(

PR #523, opened 2025-02-20
`network()`, `syslog()` sources: add `$PEERIP` and `$PEERPORT` macros

MrAnno:
The `$PEERIP` and `$PEERPORT` macros always display the address and port of the direct sender.
In most cases, these values are identical to `$SOURCEIP` and `$SOURCEPORT`.
However, when dealing with proxied protocols, `$PEERIP` and `$PEERPORT` reflect the proxy's address and port,
while `$SOURCEIP` and `$SOURCEPORT` indicate the original source of the message.


lib/logreader.c
```diff
@@ -446,6 +446,27 @@ _log_reader_insert_msg_length_stats(LogReader *self, gsize len)
...
+{
+  static NVHandle peer_addr_key = 0;
+
+  GSockAddr *source_addr = self->peer_addr;
+  if (aux->peer_addr)
+    {
+      source_addr = aux->peer_addr;
+
+      if (!peer_addr_key)
+        peer_addr_key = log_msg_get_value_handle("PEER_ADDR");
```
bazsi, line 460: We now have SOURCEIP and DESTIP as macros to extract the relevant addresses.

This time it would contain an IP:port pair (GSA_FULL).

I think we should add a separate IP and maybe port number (although this one is less important, so much so that we might want to skip it, I only added SOURCEPORT recently because of haproxy support).

I'd also think a bit if we need this as a name-value pair, storing the GSockAddr instance directly would mean that we wouldn't need to store it as a name-value pair, (but this is more of a thought than an actual recommendation).

lib/logreader.c
```diff
@@ -446,6 +446,27 @@ _log_reader_insert_msg_length_stats(LogReader *self, gsize len)
...
+
+      gchar buf[MAX_SOCKADDR_STRING];
+      log_msg_set_value(msg, peer_addr_key, g_sockaddr_format(self->peer_addr, buf, sizeof(buf), GSA_FULL), -1);
```
bazsi, line 463: the CI is now crashing, probably because self->peer_addr can be NULL.

lib/logreader.c
bazsi: btw, if PEERIP was a macro to be resolved at runtime, it could easily fall back to being SOURCEIP when PEERIP is unset, because in case there's no proxy, those should be equal.

MrAnno: PEERIP and PEERPORT should be the consistent naming, I agree.
I like the macro idea too, maybe with a tiny twist that we shouldn't add another direct C struct field just for this.

OverOrion (approved):

PR #524, opened 2025-02-21
`webhook()`: proxy and headers support

MrAnno:
Depends on #523

```
webhook(port(8080) proxy-header("x-forwarded-for"));
webhook(port(8080) proxy-header(yes));

$ curl -H "x-forwarded-for: 4.3.3.2"
    -H "X-Forwarded-FOR: 1.2.3.4" -X POST --data "{}" http://127.0.0.1:8080

$PEERIP: 127.0.0.1, $SOURCEIP: 1.2.3.4
```

```
webhook(port(8080) include-request-headers(yes));

filterx {
    headers = json(${webhook.headers});

    $type = headers["Content-Type"][-1];
    $headers = headers;
};
```

$headers:
```json
{
  "Host": ["127.0.0.1:8080"],
  "User-Agent": ["curl/8.5.0"],
  "X-Forwarded-For": [
    "4.3.3.2",
    "1.2.3.4"
  ],
  "Content-Length": ["2"],
  "Content-Type": ["application/x-www-form-urlencoded"]
}
```
```
$type == "application/x-www-form-urlencoded"
```

### DIFF
--- a/modules/python-modules/syslogng/modules/webhook/source.py
+++ b/modules/python-modules/syslogng/modules/webhook/source.py
@@ -21,6 +21,7 @@
 #############################################################################
 
 from syslogng import LogSource, LogMessage
+from collections import defaultdict
 
 import logging
 import asyncio
@@ -28,11 +29,14 @@ import threading
 import tornado
 import ssl
 import signal
+import json
 from typing import Any
 
 signal.signal(signal.SIGINT, signal.SIG_IGN)
 signal.signal(signal.SIGTERM, signal.SIG_IGN)
-WEBHOOK_QUERY_NV_PREFIX = "webhook.query."
+WEBHOOK_NV_PREFIX = "webhook."
+WEBHOOK_QUERY_NV_PREFIX = WEBHOOK_NV_PREFIX + "query."
+WEBHOOK_HEADERS_KEY = WEBHOOK_NV_PREFIX + "headers"
 
 
 class Handler(tornado.web.RequestHandler):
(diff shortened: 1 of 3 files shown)


modules/python-modules/syslogng/modules/webhook/source.py
```diff
@@ -52,6 +66,18 @@ async def post(self, **path_arguments) -> None:
...
         await self.finish({"status": "received"})
 
+    def _set_proxied_ip(self, msg: LogMessage) -> None:
+        proxy_header_candidates = [self.request.headers.get_list(header) for header in WEBHOOK_PROXY_SOURCEIP_HEADERS]
+
+        for header in proxy_header_candidates:
+            if header and len(header) > 0:
+                # the closest/last IP (the behind_proxy flag implies that the last one can be trusted)
+                msg.set_source_ipaddress(header[-1])
+                msg[WEBHOOK_NV_PREFIX + "proxy_ip"] = self.request.remote_ip
```
bazsi, line 76: in the alternate branch for a similar case, we were using $PEER_ADDR to store this information, which should be handled similarly here.

Either using a 3rd GSockAddr in LogMessage or a name-value pair, but one that matches between webhook and syslog-style sources.

MrAnno: I think everything has become consistent.

OverOrion (dismissed):

OverOrion (approved):

PR #527, opened 2025-02-25
filterx: fix switch eval errors

alltilla:

### DIFF
--- a/lib/filterx/expr-switch.c
+++ b/lib/filterx/expr-switch.c
@@ -187,6 +187,8 @@ _find_matching_case(FilterXSwitch *self, FilterXObject *selector)
       FilterXSwitchCase *switch_case = (FilterXSwitchCase *) g_ptr_array_index(self->cases, i);
 
       FilterXObject *value = _eval_switch_case(switch_case);
+      if (!value)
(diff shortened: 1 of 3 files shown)


lib/filterx/expr-switch.c
```diff
@@ -187,8 +187,18 @@ _find_matching_case(FilterXSwitch *self, FilterXObject *selector)
       FilterXSwitchCase *switch_case = (FilterXSwitchCase *) g_ptr_array_index(self->cases, i);
 
       FilterXObject *value = _eval_switch_case(switch_case);
+      if (!value)
+        {
+          filterx_object_unref(value);
```
bazsi, line 192: this is not needed, as value is NULL

bazsi (dismissed):
This is approved. The extra filterx_object_unref() is not harmful, so not critical.

alltilla:
@bazsi True, I updated the code.

lib/filterx/expr-switch.c
sodomelle: It is removed.

sodomelle (approved):
Looks good.

PR #528, opened 2025-02-25
filterx/expr-function: simple functions should get unmarshalled values

bazsi:
Right now, arguments to simple functions are evaluated at the time of the call, but that means that FilterXMessageValue objects are encountered when evaluating them.

This is:
1) inconvinient, as we have to consider both native objects and
   message values
2) somewhat incorrect, as the unmarshalled version of the arguments are
   not cached in a filterx variable.
3) if the function changes the copy, it will not be impacting the
   filterx variable, e.g. unset_empties($json) will be a noop, if $json
   is an LM_VT_JSON

I understand this may have a performance impact, but if it does, we can look into creating FilterXStrings with a borrowed string pointer.

### DIFF
--- a/lib/filterx/tests/test_object_json.c
+++ b/lib/filterx/tests/test_object_json.c
@@ -103,32 +103,12 @@ Test(filterx_json, test_json_function)
   assert_object_json_equals(fobj, "{\"foo\":1}");
   filterx_object_unref(fobj);
 
+  /* message_value is being unmarshalled before passing it to a simple
+   * function, so no need to handle FilterXMessageValue instances */
   fobj = _exec_json_func(filterx_message_value_new("{\"foo\": 1}", -1, LM_VT_JSON));
-  cr_assert(filterx_object_is_type(fobj, &FILTERX_TYPE_NAME(json_object)));
-  assert_object_json_equals(fobj, "{\"foo\":1}");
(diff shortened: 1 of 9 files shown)


lib/filterx/tests/test_object_json.c
```diff
@@ -103,35 +103,27 @@ Test(filterx_json, test_json_function)
...
-  cr_assert(filterx_object_is_type(fobj, &FILTERX_TYPE_NAME(json_object)));
-  assert_object_json_equals(fobj, "{\"foo\":1}");
-  filterx_object_unref(fobj);
+  cr_assert(fobj == NULL);
 
   fobj = _exec_json_func(filterx_message_value_new("[1, 2]", -1, LM_VT_STRING));
-  cr_assert(filterx_object_is_type(fobj, &FILTERX_TYPE_NAME(json_array)));
-  assert_object_json_equals(fobj, "[1,2]");
-  filterx_object_unref(fobj);
+  cr_assert(fobj == NULL);
```
MrAnno, line 126: It looks like these test cases became useless as they try to test something that is "not supported" on the given interface. Can we keep the first one to document the behavior and get rid of the rest?

bazsi:
fixed outstanding review comment and rebased against main.

MrAnno (approved):

PR #531, opened 2025-02-27
rename the table_name parameter of the azure monitor destination

jszigetvari:
The `table_name` parameter of the `azure-monitor` destination is actually referring to the stream name in the DCR, and not the table the logs end up getting written to. The actual destination table name is referenced in the DCR.
For this reason the `table_name` parameter should be rename to `stream_name`.

### DIFF
--- a/scl/azure/azure-monitor.conf
+++ b/scl/azure/azure-monitor.conf
@@ -22,19 +22,22 @@
 #
 #############################################################################
 
-block destination _azure_monitor_internal(
+@requires http "The azure-monitor() driver depends on the AxoSyslog http module, please install the axosyslog-mod-http (Debian & derivatives) or the axosyslog-http (RHEL & co) package"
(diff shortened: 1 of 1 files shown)


scl/azure/azure-monitor.conf
```diff
@@ -49,47 +52,3 @@ block destination _azure_monitor_internal(
     `__VARARGS__`
   );
 };
-
-block destination azure_monitor_builtin(
```
MrAnno, line 53: Please keep the old destinations for a few more releases because this would be a breaking change otherwise.
We can deprecate the old ones and document only the new one, but let's keep them here for 2 more releases.

MrAnno: (I'm sure nobody started using it, so the whole deprecation ceremony is useless, but due to how we version AxoSyslog releases, this is the cleanest and fair way.)

MrAnno (approved):

MrAnno:
Thank you.

PR #534, opened 2025-03-02
light: add docker runner

alltilla:
Commit organization is not the best, I think it is easiest to review this IRL, or I can split it to a lot more commits.

Can be tested with:
```
docker pull ghcr.io/axoflow/axosyslog:nightly
make light-check PYTEST_OPTS='--runner=docker --docker-image=ghcr.io/axoflow/axosyslog:nightly'
```


mitzkia:
Thanks for the PR. 
We agreed in that we will continue the review/discussion of the PR in irl.

alltilla:
@mitzkia Updated as we talked IRL, can you please re-review?

PR #537, opened 2025-03-06
App transform topic

bshifter:
Renamed the flavour field in app_transform to topic and ensured it is referred to as topic throughout. This improves consistency between app-transform and app-parser.
app-transform functionality also covered with unit tests.

PR #541, opened 2025-03-12
`app-transform()`: support `parser {}`

alltilla:
This is a breaking change, as defining a transformation was done with an implicit `filterx` block, which we have
to explicitly state now. However, this is not a documented or widely used feature, so I think we are okay with it.

---

The code generator logic is the following:

Transformations can have either `filterx` only transforms or mixed/`parser` only transform.

First, we gather the transformations that have mixed/`parser` only transforms, and first check them with `log` level `if`-`elif`s and `filterx` pre-checks to the variable and app name.

Then, to keep the `filterx` switch-case optimization, we gather `filterx` transformations that have `filterx` transforms active only (`include/exclude-transforms()` honored) and put them to an `else` block to a switch-case.

There are cases where no mixed/`parser` only or no `filterx` only transformations are available, those are handled correctly, too.

---

Depends on #537


MrAnno (approved):

bshifter (approved):

MrAnno (approved):

alltilla:
rebased to main

PR #546, opened 2025-03-18
Debugger add list command

bazsi:
This is spin off from #340 to make that easier to merge. This part of the PR focuses on the interactive "list" command, which will make it easy to navigate the source code (e.g. the syslog-ng.conf file) we are debugging.


bazsi:
rebased against current main.

PR #548, opened 2025-03-18
Filterx native dict and list object

bazsi:
This branch implements a faster representation for dict(), adds a native list() implementation, eliminates object-json-* and changes how we do copy-on-write for mutable objects.

This depends on a number of PRs to go in first, namely:

- [ ] #528 
- [ ] #543
- [ ] #545


bazsi:
I've found an issue with cache_json_file(), as it uses the weakrefs mechanism but that's not yet available at compile time.

bazsi:
It seems I have found the issue with cache_json_file(), I added a short term fix (longer term the cache_json_file() mechanism could be improved by combining freeze and make_readonly and delegating the deep_freeze code to mutable objects).

bazsi:
I still have an issue with frozen refs. I already have probable fix, but I am still validating it.

bazsi:
Make check runs, light check is green, let's hope the CI agrees and then this is ready for review.

MrAnno:
I'm not done with the review, will need a few rounds, but we discussed a bigger problem today: dirty tracking.
Keeping a weak reference to the root is not enough if we start using FilterXRef multiple levels inside dicts.
What we need is basically a back-reference to the parent object (to the FilterXRef of the parent object to be precise).

This implies a lot of complications during the copy operation, which is a bigger work, so we agreed to merge this with the following clever workaround:
- dirty tracking will work properly through FilterX expressions, see `writable_requested()` (this is far away from objects, so it's not ideal, but works well)
- dirty tracking won't work with the current `root_container`, so we can remove it. This implies that C code that writes dicts will be problematic: the expression workaround is needed for all locations where dicts are modified in the C code (not only unwrap_rw).

lib/filterx/filterx-ref.c
```diff
@@ -55,7 +73,9 @@ _filterx_ref_setattr(FilterXObject *s, FilterXObject *attr, FilterXObject **new_
 
   _filterx_ref_cow(self);
 
-  return filterx_object_setattr(self->value, attr, new_value);
+  gboolean result = filterx_object_setattr(self->value, attr, new_value);
+  filterx_ref_propagate_root(*new_value, self);
```
MrAnno: This implies that dirty tracking will not work purely on objects anymore, which is fine, but it made me think: object, ref, and dict got so coupled that certain functionality won't work properly without using the combinations of their API. `FilterXObject` should help with a facade in the future.

lib/filterx/filterx-object.h
```diff
@@ -480,4 +480,12 @@ filterx_object_set_dirty(FilterXObject *self, gboolean value)
 
 #include "filterx-ref.h"
 
+static inline gboolean
+filterx_object_is_type_or_ref(FilterXObject *object, FilterXType *type)
```
MrAnno, line 530: I wanted to avoid introducing such a function because `FilterXRef` should be completely transparent for the developer unless they want to do downcasts, but it those cases, the developer should get familiar with refs and do an explicitly unwrap on them.

I see this comes in handy in the dict implementation, but I'm wondering: can we move this into the `dict.h` file and specialize it like `is_dict_or_ref()` to make it clear that this should not be used, only in special cases?

lib/filterx/filterx-variable.h
```diff
@@ -117,6 +117,7 @@ filterx_variable_set_value(FilterXVariable *v, FilterXObject **new_value, gboole
   filterx_object_unref(v->value);
   if (new_value)
     {
+      *new_value = filterx_ref_new(*new_value);
```
MrAnno: `filterx_ref_new()` avoids nesting refs in refs or wrapping non-mutable objects, so this won't have too big of an overhead, but my original idea was to create refs only when necessary, when we see
- an assignment
- setsubscript, setattr
- when dealing with dict/list entries internally

I implemented this on the expression level, because something was not right when I tried it on the variable layer.

Generally, we want all mutable objects to be handler through refs, as you mentioned. But if we add this here, can't we remove a few from the expression side?

lib/filterx/tests/test_expr_plus_generator.c
```diff
@@ -233,35 +235,28 @@ Test(expr_plus_generator, test_nested_dict_add_variable_to_generator_with_post_s
...
   filterx_literal_generator_set_elements(lhs, lhs_vals);
 
-  FilterXExpr *rhs = filterx_non_literal_new(filterx_json_object_new_from_repr("{\"tik\":{\"tak\":\"toe\"}}", -1));
+  FilterXExpr *rhs = filterx_non_literal_new(filterx_object_from_json("{\"tik\":{\"tak\":\"toe\"}}", -1, NULL));
 
   FilterXExpr *expr = filterx_operator_plus_generator_new(lhs, rhs);
-  FilterXObject *json_dict = filterx_json_object_new_empty();
-  FilterXExpr *fillable = filterx_literal_new(json_dict);
+  FilterXObject *dict = filterx_dict_new();
+  FilterXExpr *fillable = filterx_literal_new(dict);
```
MrAnno, line 242: I think it would be worth adding the test case we talked about:

```
fx_var = dict($MSG);
b = fx_var.a;
b.c = 2;

fx_var, fx_var.a remains unchanged
```

```
b = $MSG.a;
$MSG.a.b.c = 2;

dirty tracking works ($MSG is synced), b remains unchanged 
```

lib/filterx/filterx-object.h
```diff
@@ -483,4 +483,64 @@ filterx_object_is_type_or_ref(FilterXObject *object, FilterXType *type)
...
+  if (!value || value->readonly || !_filterx_type_is_referenceable(value->type))
+    return;
+  *o = _filterx_ref_new(value);
+}
+
+
+/* NOTE: potentially replaces *value with a FilterXRef, sinking the passed
+ * reference.  It replaces the copied object as a return value */
+static inline FilterXObject *
+filterx_object_cow_fork(FilterXObject **o)
```
MrAnno, line 558: As we are planning to implement dirty tracking on top of FilterXRef, I think this API is no longer only about CoW.

IMHO `mutable`, `mut`, `ref`, or `reference` is the name we are looking for here, copy-on-write is only just one of the features it provides.

MrAnno (commented):
I will need one last round to review carefully the dict implementation itself.

MrAnno:
I checked a few scenarios, the new way of handling refs seems to be working in all cases. It sounds a little complicated with those floating refs, but I couldn't come up with a simpler alternative.

bazsi:
I have cleaned up my patches, removed the various deadends from the history, a rebase is still to be done.

The diff between the current version and the one pushed here the last time is here:
https://github.com/axoflow/axosyslog/compare/ae46224301d4a6eee454d763ff3e55ab85c3c2d2..d40efef47957e20892474cc406e3cf05ea17a5bd

The branch history is very different now, as I said I wiped out the incomplete solutions scattered in the history.

lib/filterx/filterx-ref.c
bazsi: True, dirty tracking can only work for objects wrapped in FilterXRefs. For now OTel classes are mutable but don't use the new API, but we discussed we would rework those and eliminate getattr/setattr for OTel objects.

lib/filterx/filterx-object.h
bazsi: this is 99% used testing code, in cases where we only check if we got the right type but do not actually work with the objects themselves.

In those cases having to typecheck against both ref and the native type is pretty cumbersome.

lib/filterx/filterx-variable.h
bazsi: I don't think we can, but we can still discuss this. Right now we can wrap objects into refs at three places:

1) as a result of an expression, it can already be boxed (for instance create_container for generators)
2) right before an assignment/setattr/set_subscript
3) right before storing it into a mutable container, such as a dict/list

This makes sure that eventually everything is boxed into a ref, even if steps prior did not do it. I think we might be able to remove some of this if we get rid off generators.

lib/filterx/filterx-object.h
bazsi: I didn't change the naming here, but moved the set_parent_container/unset_parent functions back to filterx_ref namespace.

bazsi:
I resolved the conflict between main and this branch.

lib/filterx/expr-comparison.c
```diff
@@ -149,11 +148,11 @@ _evaluate_type_aware(FilterXObject *lhs, FilterXObject *rhs, gint operator)
   if (lhs->type == rhs->type &&
       (filterx_object_is_type(lhs, &FILTERX_TYPE_NAME(string)) ||
        filterx_object_is_type(lhs, &FILTERX_TYPE_NAME(bytes)) ||
-       filterx_object_is_type(lhs, &FILTERX_TYPE_NAME(protobuf)) ||
-       filterx_object_is_type(lhs, &FILTERX_TYPE_NAME(json_object)) || // TODO: we should have generic map and array cmp
-       filterx_object_is_type(lhs, &FILTERX_TYPE_NAME(json_array))))
+       filterx_object_is_type(lhs, &FILTERX_TYPE_NAME(protobuf))))
     return _evaluate_as_string(lhs, rhs, operator);
 
+  /* FIXME: add dict/list comparison */
```
MrAnno, line 153: Can we merge it without comparison? We probably don't have too many jsons/dicts compared at the moment, maybe just equals/notequals.

MrAnno (approved):

lib/filterx/expr-comparison.c
MrAnno: Yes, after the release.

This way, we have time to implement the comparison separately.

MrAnno:
I'll merge this soon, within 1-2 days.

PR #551, opened 2025-03-20
light: release workflow

alltilla:
Triggered either by a tag starting with "light-" or by manually on the same tags.

Based on:
https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/configuring-openid-connect-in-pypi

---

Example run: https://github.com/alltilla/axosyslog/actions/runs/13970900983


bazsi (approved):
As I see this branch contains a fix for a breakage on main.

PR #554, opened 2025-03-25
light: implement src-layout dir structure

alltilla:
When adding a dependency to a poetry-based project in editable mode, for some reason, plain setuptools is used to install the package.

setuptools has a problem finding the right package dir, in the current flat-layout dir structure.

https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#src-layout


mitzkia (approved):
Thx for the changes. Approved

PR #556, opened 2025-03-25
light: support docker execution for loggen

mitzkia:
- This PR add support to use Loggen() with docker execution
 - also adds read_until_eof() for file destination driver

### DIFF
--- a/tests/light/src/axosyslog_light/fixtures.py
+++ b/tests/light/src/axosyslog_light/fixtures.py
@@ -182,8 +182,9 @@ def syslog_ng_ctl(request: pytest.FixtureRequest, testcase_parameters, container
 
 @pytest.fixture
 def container_name(request: pytest.FixtureRequest, testcase_parameters: TestcaseParameters):
-    container_name = f"{testcase_parameters.get_testcase_name()}_{tc_parameters.INSTANCE_PATH.get_instance_name()}"
-    return re.sub(r'[^a-zA-Z0-9_.-]', '_', container_name)
+    testcase_name_and_instance = f"{testcase_parameters.get_testcase_name()}_{tc_parameters.INSTANCE_PATH.get_instance_name()}"
+    tc_parameters.CONTAINER_NAME = re.sub(r'[^a-zA-Z0-9_.-]', '_', testcase_name_and_instance)
+    return tc_parameters.CONTAINER_NAME
 
 
 @pytest.fixture
--- a/tests/light/src/axosyslog_light/fixtures.py
+++ b/tests/light/src/axosyslog_light/fixtures.py
@@ -60,7 +60,7 @@ class InstallDirAction(argparse.Action):
             raise argparse.ArgumentTypeError("{0} is not a valid directory".format(path))
 
         binary = Path(install_dir, "sbin/syslog-ng")
-        if not binary.exists():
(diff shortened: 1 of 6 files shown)


tests/light/src/axosyslog_light/fixtures.py
```diff
@@ -60,7 +60,7 @@ def __call__(self, parser, namespace, path, option_string=None):
             raise argparse.ArgumentTypeError("{0} is not a valid directory".format(path))
 
         binary = Path(install_dir, "sbin/syslog-ng")
-        if not binary.exists():
+        if namespace.runner == "local" and not binary.exists():
```
alltilla, line 63: I think this depends on the order of setting `--runner` and `--installdir`. The user should not use `--installdir` with `--runner=docker`, so I think this can only happen as  a user error. If we have make targets or such, we should make sure to not include these two options at the same time, I believe.

tests/light/src/axosyslog_light/syslog_ng/syslog_ng_docker_executor.py
```diff
@@ -33,10 +33,12 @@
 
 
 class SyslogNgDockerExecutor(SyslogNgExecutor):
-    def __init__(self, container_name: str, image_name: str) -> None:
+    def __init__(self, container_name: str, image_name: str, extra_volume_mounts: typing.Optional[set] = None, extra_env_var: typing.Optional[str] = None) -> None:
         self.__process_executor = ProcessExecutor()
         self.__container_name = container_name
         self.__image_name = image_name
+        self.__extra_volume_mounts = extra_volume_mounts
+        self.__extra_env_var = extra_env_var
```
alltilla, line 41: Let's make this a dict, so we can include more env variables if needed.

tests/light/src/axosyslog_light/syslog_ng/syslog_ng_docker_executor.py
```diff
@@ -59,11 +61,18 @@ def run_process(
...
 
         for path in paths:
             command += ["-v", f"{path}:{path}"]
 
+        if self.__extra_env_var:
+            command += ["-e", self.__extra_env_var]
+
+        if self.__extra_volume_mounts:
+            for mount in self.__extra_volume_mounts:
+                command += ["-v", mount]
```
alltilla, line 74: I think this should be a dict at least, so we can mount something from the host to a different place in the container. Maybe a list or set of 2 element tuples would be the best.

tests/light/src/axosyslog_light/testcase_parameters/testcase_parameters.py
```diff
@@ -26,6 +26,7 @@
 
 
 INSTANCE_PATH = None
+RUNNER = None
```
alltilla, line 29: I think passing the parameter through classes is the way to go. Yes, we have a `--runner` option, but what happens when we want to run a client from a docker runner, but run the server from a local runner? We cannot pinpoit one runner for all our syslog-ng instances, neither we can pinpoint one container name, as we might have multiple containers for a client, server, etc... (this change comes in a future commit)

We tackle the same problem with the `Executor` class (`SyslogNgExecutor` and `SyslogNgCtlExecutor`).

PR #557, opened 2025-03-26
modules/affile/std{in|out}: register stdin/stdout with proper name instead of dash

OverOrion:


mitzkia:
Thanks for the PR. 
The added trace log can help a lot during a debug session.

modules/affile/stdin.c
```diff
@@ -58,7 +58,7 @@ file_opener_for_stdin_new(void)
 LogDriver *
 stdin_sd_new(GlobalConfig *cfg)
 {
-  AFFileSourceDriver *self = affile_sd_new_instance("-", cfg);
+  AFFileSourceDriver *self = affile_sd_new_instance("/dev/stdin", cfg);
```
MrAnno, line 61: What problem does this patch solve?

I'm asking because `/dev/stdin` is not POSIX, so I don't know if this would work outside Linux.
It's basically a symlink to `/proc/self/fd/0`.

OverOrion: https://github.com/axoflow/axosyslog/blob/bff69a5dfbb064e354c0671fe9b55a3673335afe/modules/affile/file-opener.c#L175

The default `-` character for path in this case is easy to miss when debugging trace logs, something closer to `stdin` communicates this intent better I think.

OverOrion: Sorry, just seeing the second comment.
Of course, will do it in a minute, thanks for double checking it thoroughly! :rocket:

MrAnno (approved):

PR #558, opened 2025-03-26
light: syslog source

alltilla:
We already had a `SyslogSource`, but it did not send the logs with framing, so this PR fixes that.

I also observed that the "log" entity all throughout Light is not separated well from some transport specific details, like `\n` delimiting in case of network and file drivers (`\n` was considered part of the log, but it is in reality a delimiter in case of network and file sources), I also fixed these.


mitzkia (approved):
Thanks for the huge refactor. Approved

PR #560, opened 2025-03-27
light: source metrics, `FilterX`, `OpenTelemetrySource`

alltilla:
Also bumped to `0.2.0` so we can release after merging this.


mitzkia (approved):
Thanks a lot for this PR. Approved.

PR #565, opened 2025-03-31
dqtool: Fixed segmentation fault when loading 16 bit buffer

MrAnno:
Cherry-picked from syslog-ng/syslog-ng#5310


lib/serialize.h
```diff
@@ -127,23 +125,23 @@ serialize_read_uint32_array(SerializeArchive *archive, guint32 *values, gsize el
 static inline gboolean
 serialize_read_uint16_array(SerializeArchive *archive, guint32 *values, gsize elements)
```
bazsi, line 126: This is only used for compatibility, so I guess the extra malloc is ok.

MrAnno: Discussed this IRL with Bazsi: I'll change the implementation and use alloca() instead, we know for a fact that `elements` can be max 256, so we'll change the parameter type and revert back to stack allocation.

bazsi (approved):

PR #566, opened 2025-03-31
light: `SessionData` refcount and `LogContextSclBlocks`

alltilla:
Previously, the first worker to finish could remove the file while others were still running, causing race conditions.

After this commit, an active worker counter is maintained, and the file is deleted only when the counter reaches zero.

---

`LogContextSclBlocks` is useful to use things like this in the config:

```
block log foobar() {
   ....
};

log {
  foobar(); # generating this line is now possible
};


mitzkia (approved):
Approved

PR #575, opened 2025-04-11
`dbld` but without the `d`

MrAnno:
I also like to use rootless podman sometimes.

Cherry-picked from syslog-ng/syslog-ng#5326


HofiOne:
shortest contribution ever :D

PR #576, opened 2025-04-14
light: fix multiple implicit source groups

alltilla:
The bug happened when we added the same source driver to multiple logpaths, like:
```py
source = config.create_xy_source()
config.create_logpath(statements=[source, dest_1])
config.create_logpath(statements=[source, dest_2])
```

This implicitly generated two separate source groups for `source`, which failed to start AxoSyslog.

The fix uses the `id()` function in python, which returns a unique id (pointer) to the python object, in this case the `SourceDriver`. This is not doing a real equivalence check on the object level, so if we create two separate, but identically configured sources with `config.create_xy_source()`, they will both have their respective implicitly generated source groups, but I don't think we need to optimize for this use case.

### DIFF
--- a/tests/light/pyproject.toml
+++ b/tests/light/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "axosyslog-light"
-version = "1.1.0"
+version = "1.2.0"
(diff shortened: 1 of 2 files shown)


mitzkia (approved):
Thanks, approved

PR #577, opened 2025-04-14
light: do not call create_statement_group unnecessarily

alltilla:
It has a side effect of adding a new statement
group to the config, although it is not getting
used by anything, because the cached one is
returned.

No need to release this commit on its own, as it
is a non-functional change regarding AxoSyslog's
execution from Light.


mitzkia (approved):
Thanks, approved

PR #580, opened 2025-04-16
Strcasecmp

MrAnno:


lib/filterx/func-str.c
```diff
@@ -497,3 +519,198 @@ filterx_function_includes_new(FilterXFunctionArgs *args, GError **error)
...
+          filterx_object_unref(b_obj);
+          goto exit;
+        }
+    }
+  else
+    b_str = self->b.literal->str;
+
+  /* TODO: utf-8 support together with all func-str-transform */
+  gint cmp = g_ascii_strcasecmp(a_str, b_str);
+  result = filterx_integer_new(cmp);
```
bazsi, line 569: we might need to cache integer(-1) otherwise we will be allocating one everytime we return -1 here

lib/filterx/func-str.c
```diff
@@ -497,3 +519,198 @@ filterx_function_includes_new(FilterXFunctionArgs *args, GError **error)
...
+
+  const gchar *a_str = NULL;
+  if (!self->a_literal)
+    {
+      a_obj = filterx_expr_eval(self->a.expr);
+      if (!a_obj)
+        goto exit;
+
+      gsize a_str_len = 0;
+      if (!filterx_object_extract_string_ref(a_obj, &a_str, &a_str_len))
```
alltilla, line 538: This might return a non NUL terminated string from an indirect LogMessage NVPair's Value. Passing it to a  function that expect it to be NUL terminated is not ideal. We should `APPEND_ZERO` it, right?

alltilla (commented):
One comment, otherwise approve!

lib/filterx/func-str.c
alltilla: Yes, it starts from -1: https://github.com/axoflow/axosyslog/blob/main/lib/filterx/object-primitive.c#L420

lib/filterx/func-str.c
```diff
@@ -497,3 +519,198 @@ filterx_function_includes_new(FilterXFunctionArgs *args, GError **error)
...
+        {
+          filterx_eval_push_error("failed to extract string value", self->b.expr, b_obj);
+          filterx_object_unref(b_obj);
+          goto exit;
+        }
+    }
+  else
+    b_str = self->b.literal->str;
+
+  /* TODO: utf-8 support together with all func-str-transform */
```
bazsi, line 567: we do not want to support utf8 here

lib/filterx/func-str.c
```diff
@@ -497,3 +519,198 @@ filterx_function_includes_new(FilterXFunctionArgs *args, GError **error)
...
+  FilterXStrcasecmp *self = (FilterXStrcasecmp *) s;
+  if (self->a_literal)
+    g_string_free(self->a.literal, TRUE);
+  else
+    filterx_expr_unref(self->a.expr);
+
+  if (self->b_literal)
+    g_string_free(self->b.literal, TRUE);
+  else
+    filterx_expr_unref(self->b.expr);
```
alltilla, line 675: We don't release the expr in optimize, so I think we should free the expr unconditionally here, or do that in optimize.

bazsi: that seems to be in a union, so the expr pointer is not available anymore.

lib/filterx/func-str.c
```diff
@@ -65,6 +67,26 @@ typedef struct _FilterXExprAffix
   FilterXExprAffixProcessFunc process;
 } FilterXExprAffix;
 
+typedef struct _FilterXStrcasecmp
+{
+  FilterXFunction super;
+
+  union
+  {
+    GString *literal;
```
bazsi, line 76: I don't see much value in using a GString wrapper here, but that's only a nit, it adds an extra indirection at runtime.

lib/filterx/func-str.c
MrAnno: The same is available in glib, but we need to calculate the MIN(a,b) then.

alltilla: Fixed.

MrAnno: Thank you.

PR #581, opened 2025-04-16
Filterx: add filterx_object_cow_fork2() function

bazsi:
This branch adds filterx_object_cow_fork2() function that allows us decrease the number of FilterXRef allocations (when set_subscript/setattr/assignment sees a bare, unwrapped FilterXDict or a FilterXList instance.

### DIFF
--- a/lib/filterx/filterx-object.h
+++ b/lib/filterx/filterx-object.h
@@ -552,30 +552,30 @@ filterx_object_is_type_or_ref(FilterXObject *object, FilterXType *type)
  * reference, returning a FilterXRef instead.
  */
 static inline void
-filterx_object_cow_wrap(FilterXObject **o)
+filterx_object_cow_wrap(FilterXObject **pself)
 {
-  FilterXObject *value = *o;
+  FilterXObject *value = *pself;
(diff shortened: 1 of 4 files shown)


lib/filterx/filterx-object.h
```diff
@@ -552,30 +552,64 @@ filterx_object_is_type_or_ref(FilterXObject *object, FilterXType *type)
...
 }
 
 /* */
 static inline FilterXObject *
-filterx_object_cow_store(FilterXObject **o)
+filterx_object_cow_store(FilterXObject **pself)
 {
-  filterx_object_cow_wrap(o);
-  return filterx_object_ref(*o);
+  filterx_object_cow_wrap(pself);
```
MrAnno, line 611: A review comment slipped through a previous PR.
I believe this should be called `filterx_object_wrap_into_ref()` or just simply `filterx_ref_wrap/new()` as wrapping a copy-on-write looks a little confusing.

MrAnno (approved):

lib/filterx/filterx-object.h
bazsi: Let's continue this discussion in https://github.com/axoflow/axosyslog/pull/595

PR #582, opened 2025-04-16
Filterx deduplicate frozen strings

bazsi:
This PR changes frozen FilterXString instances so they use the same instance throughout the execution of a configuration. This deduplicates strings such as common attribute name ("meta", "name", etc), it improves memory, cache use and thereby performance.

It rounds up the allocation strategies for FilterXObject instances to 4:
* normal heap allocated/refcounted
* stack allocated
* hybernated (e.g. created at syslog-ng startup, reloads not affecting them)
* frozen (associated with configurations)

Due to setting the hash value at freeze time, this depends on #583, so I am rebasing this on top of that.

### DIFF
--- a/lib/filterx/object-string.c
+++ b/lib/filterx/object-string.c
@@ -187,6 +187,7 @@ _string_freeze(FilterXObject **pself)
       *pself = frozen_string;
       return;
     }
+  self->hash = _filterx_string_hash(self);
   g_hash_table_insert(global_cache.string_frozen_cache, (gchar *) self->str, self);
 }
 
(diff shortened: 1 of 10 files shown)


lib/filterx/object-string.c
```diff
@@ -187,6 +187,7 @@ _string_freeze(FilterXObject **pself)
       *pself = frozen_string;
       return;
     }
+  self->hash = _filterx_string_hash(self);
```
MrAnno, line 190: `_filterx_string_hash()` already has the side-effect to store its value in `self->hash`.

MrAnno (dismissed):

lib/filterx/object-string.c
MrAnno: Will fix that too.

MrAnno (approved):

PR #583, opened 2025-04-16
Filterx cache string hash values

bazsi:
This branch adds a "hash" value to FilterXString instances, which is calculated either:

1) when the string is frozen, as that happens during compile phase, so it's cheap
2) at runtime at the first invocation (potentially racy, but should resolve itself, review required!)

This way, a hash lookup with a frozen string will avoid having to do a g_str_hash() call

### DIFF
--- a/lib/filterx/object-string.c
+++ b/lib/filterx/object-string.c
@@ -146,6 +146,8 @@ _string_new(const gchar *str, gssize str_len, FilterXStringTranslateFunc transla
   if (str_len == -1)
     str_len = strlen(str);
 
+  g_assert(str_len < G_MAXUINT);
+
   FilterXString *self = g_malloc(sizeof(FilterXString) + str_len + 1);
   memset(self, 0, sizeof(FilterXString));
   filterx_object_init_instance(&self->super, &FILTERX_TYPE_NAME(string));
--- a/lib/filterx/object-string.c
+++ b/lib/filterx/object-string.c
@@ -163,6 +163,31 @@ _string_new(const gchar *str, gssize str_len, FilterXStringTranslateFunc transla
   return self;
 }
 
+static inline guint
+_hash_str(const gchar *str, gsize str_len)
+{
+  const char *p;
+  guint32 h = 5381;
+
+  for (p = str; str_len > 0 && *p != '\0'; p++, str_len--)
+    h = (h << 5) + h + *p;
+
+  return h;
+}
+
+guint
+_filterx_string_hash(FilterXString *self)
+{
+  /* NOTE: this is a data race, as self->hash is written without atomics and
+   * also without mutexes.  Since self->hash is aligned, and is just 32
+   * bits, torn writes do not happen in modern architectures (x86, x86_64,
+   * arm).  */
+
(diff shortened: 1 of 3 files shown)


lib/filterx/object-string.c
```diff
@@ -157,6 +159,25 @@ _string_new(const gchar *str, gssize str_len, FilterXStringTranslateFunc transla
...
+  for (p = str; str_len > 0 && *p != '\0'; p++, str_len--)
+    h = (h << 5) + h + *p;
+
+  return h;
+}
+
+guint
+_filterx_string_hash(FilterXString *self)
+{
+  self->hash = _hash_str(self->str, self->str_len);
```
MrAnno, line 177: The same story that always bothers me:

`self->hash` is not atomic, so if writing it requires multiple instructions on a CPU architecture, parallel writes could produce an invalid value. Such architecture doesn't really exist I guess, but as long as this is not declared atomic, the compiler and processors can do optimizations that doesn't take multiple thread access into account.

I've never been able to catch a bug like this before, so I'm just probably getting stressed on C and other standards.

lib/filterx/object-string.h
```diff
@@ -90,6 +91,29 @@ filterx_string_new(const gchar *str, gssize str_len)
...
+
+  /* although this is racy for parallel access on the same object, it's not
+   * really a problem, as:
+   *
+   * 1) we are only sharing frozen instances of the string, which calculates
+   *    the hash at freeze time
+   *
+   * 2) even if we do share a non-frozen string, the hash algorithm should
+   *    have the same result, so worst case, we calculate the hash 2 times.
+   */
```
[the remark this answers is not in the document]
bazsi: This just means that it would be great to have  a FilterXString constructor variant that takes an allocated space and frees that at destruction time.

This reminded me to this, so I implemented that separately in this PR:

https://github.com/axoflow/axosyslog/pull/596

This will not lead to more sharing though, it's just an initialization trick.

lib/filterx/object-string.c
bazsi: This is a long chat with ChatGPT and it seems to agree with me that I can't get torn writes at least on modern architectures  x86_64/x86/arm, as long as the variable is aligned and volatile.

https://chatgpt.com/share/6802b2d5-67f8-800d-ac5b-da5ca17f38f5

I added volatile now.

I am worried about adding atomics here as the read side of this is a hot path (during hash lookup, potential matches are checked with their hash value), and would probably make the cached version slower.

bazsi: btw, I did try to use new C11 style atomics but it barfs on our C++ code in OTel. I did want to check if there's a measurable impact or no, but this proved to be difficult, while the original solution should also be safe.

bazsi: I added a static assert to make sure that the size of the hash value is 32 bits and I added a comment to explain that it should stay this way, or else.

MrAnno: Thank you.

MrAnno (approved):

PR #584, opened 2025-04-16
Filterx: turn parse_csv() into a function

bazsi:
This turns parse_csv() from a generator to a function, returning a dict instance. It also optimizes dict allocation when
the number of columns are known.

### DIFF
--- a/modules/csvparser/filterx-func-parse-csv.c
+++ b/modules/csvparser/filterx-func-parse-csv.c
@@ -33,7 +33,9 @@
 #include "filterx/object-message-value.h"
 #include "filterx/object-null.h"
 #include "filterx/filterx-object.h"
+#include "filterx/object-dict.h"
 #include "filterx/object-dict-interface.h"
+#include "filterx/object-list.h"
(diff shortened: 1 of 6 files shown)


modules/csvparser/filterx-func-parse-csv.c
```diff
@@ -558,12 +531,12 @@ filterx_function_parse_csv_new(FilterXFunctionArgs *args, GError **error)
...
 
 error:
   filterx_function_args_free(args);
-  filterx_expr_unref(&self->super.super.super);
+  filterx_expr_unref(&self->super.super);
   return NULL;
 }
 
-FILTERX_GENERATOR_FUNCTION(parse_csv, filterx_function_parse_csv_new);
+FILTERX_FUNCTION(parse_csv, filterx_function_parse_csv_new);
```
MrAnno, line 542: There is a `FILTERX_GENERATOR_FUNCTION_DECLARE()` macro in the header file.

MrAnno:
I'll fix this.

MrAnno (approved):

PR #586, opened 2025-04-16
filterx/func-set-timestamp: add get_timestamp() function

bazsi:
This adds get_timestamp() which allows us not to use $R_UNIXSTAMP. evaluating a macro is slower than going directly to the LogStamp structure.

also, we already have a set_timestamp() function


lib/filterx/func-set-timestamp.c
```diff
@@ -200,3 +200,91 @@ filterx_function_set_timestamp_new(FilterXFunctionArgs *args, GError **error)
...
+  if (!idx_str)
+    {
+      g_set_error(error, FILTERX_FUNCTION_ERROR, FILTERX_FUNCTION_ERROR_CTOR_FAIL,
+                  "Second argument must be string type. " FILTERX_FUNC_GET_TIMESTAMP_USAGE);
+      return FALSE;
+    }
+
+  self->timestamp_idx = log_msg_lookup_time_stamp_name(idx_str);
+
+  if ((self->timestamp_idx != LM_TS_STAMP) && (self->timestamp_idx != LM_TS_RECVD))
```
MrAnno, line 260: Shouldn't we just check for the error value -1?

bazsi: set_timestamp() also only accepts these values, although I think we can allow all values.

I am not sure why that was restricted.

bazsi: I've changed this, it sems the lookup_time_stamp_name() function only supports recvd and stamp anyway.

MrAnno (approved):

PR #587, opened 2025-04-17
Filterx: get_sdata() is a function instead of a generator

bazsi:


OverOrion (approved):
Thanks, this looks good to me aside from the failing style check :rocket: 

Also huge thanks for splitting these PRs.

PR #589, opened 2025-04-17
Filterx: handle macros in expr variable

bazsi:
Instead of adding resolved macro values into our variable cache, just resolve them in place without even registering them.

This has two side effects: 
* macros are always resolved and never cached (I think this is a good thing)
* macro values are not bounced into FilterXMessageValue instances but get directly unmarshalled into a typed FilterXObject


MrAnno (approved):
I noticed we have dynamically changing macros like `M_SYSUPTIME`. Not storing them in FilterX may present a more consistent look on those.

PR #593, opened 2025-04-17
light: Add various improvements to NetworkIO() and SyslogNg()

mitzkia:
NetworkIO() improvement:
 - return object of loggen process from NetworkIO.write_messages() functions, after this we will able to wait for process finish on caller side
 - pass "extra_loggen_args" to NetworkIO.write_messages() function, after this we can specify other loggen params on caller side as well

SyslogNg() improvement:
 - eliminate duplicate code of SyslogNg.get_config_version() and SyslogNg.get_version(), also fixes a bug in one of the  returned value.

### DIFF
--- a/tests/light/src/axosyslog_light/driver_io/network/network_io.py
+++ b/tests/light/src/axosyslog_light/driver_io/network/network_io.py
@@ -81,8 +81,8 @@ class NetworkIO():
             return "".join([str(len(message)) + " " + message for message in messages])
         return "".join([message + "\n" for message in messages])
 
-    def write_messages(self, messages, rate=None, transport=None, framed=None):
-        return self.write_raw(self.__format_messages(messages, framed=framed), rate=rate, transport=transport)
+    def write_messages(self, messages, rate=None, transport=None, framed=None, extra_loggen_args=None):
+        return self.write_raw(self.__format_messages(messages, framed=framed), rate=rate, transport=transport, extra_loggen_args=extra_loggen_args)
 
     def write_messages_with_proxy_header(self, proxy_version, src_ip, dst_ip, src_port, dst_port, messages, rate=None, transport=None, framed=None):
         return self.write_raw(
--- a/tests/light/src/axosyslog_light/driver_io/network/network_io.py
+++ b/tests/light/src/axosyslog_light/driver_io/network/network_io.py
@@ -60,7 +60,7 @@ class NetworkIO():
         loggen_input_file = File(loggen_input_file_path)
         loggen_input_file.write_content_and_close(raw_content)
 
-        Loggen().start(
+        loggen_process = Loggen().start(
             LoggenStartParams(
                 target=self.__ip,
                 port=self.__port,
@@ -72,6 +72,7 @@ class NetworkIO():
                 **extra_loggen_args,
             ),
         )
+        return loggen_process
 
     def __format_messages(self, messages, framed=None):
         if framed is None:
@@ -81,10 +82,10 @@ class NetworkIO():
         return "".join([message + "\n" for message in messages])
 
     def write_messages(self, messages, rate=None, transport=None, framed=None):
-        self.write_raw(self.__format_messages(messages, framed=framed), rate=rate, transport=transport)
+        return self.write_raw(self.__format_messages(messages, framed=framed), rate=rate, transport=transport)
 
     def write_messages_with_proxy_header(self, proxy_version, src_ip, dst_ip, src_port, dst_port, messages, rate=None, transport=None, framed=None):
-        self.write_raw(
+        return self.write_raw(
             self.__format_messages(messages, framed=framed),
             rate=rate,
             transport=transport,
--- a/tests/light/pyproject.toml
+++ b/tests/light/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "axosyslog-light"
-version = "1.2.0"
+version = "1.3.0"
 description = "Lightweight End-to-End Test Framework for AxoSyslog."
 authors = ["Andras Mitzki <andras.mitzki@axoflow.com>", "Attila Szakacs <attila.szakacs@axoflow.com>"]
 readme = "README.md"
--- a/tests/light/src/axosyslog_light/syslog_ng/syslog_ng.py
+++ b/tests/light/src/axosyslog_light/syslog_ng/syslog_ng.py
@@ -125,7 +125,7 @@ class SyslogNg(object):
         self.stop()
         self.start(config)
 
-    def get_config_version(self) -> str:
+    def _get_version_info(self, parse_line: typing.Callable[[str], str], error_message: str) -> str:
         stdout_path = Path(f"syslog_ng_{self.instance_paths.get_instance_name()}_version_stdout")
         stderr_path = Path(f"syslog_ng_{self.instance_paths.get_instance_name()}_version_stderr")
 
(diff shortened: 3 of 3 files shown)


tests/light/src/axosyslog_light/driver_io/network/network_io.py
```diff
@@ -81,8 +81,8 @@ def __format_messages(self, messages, framed=None):
             return "".join([str(len(message)) + " " + message for message in messages])
         return "".join([message + "\n" for message in messages])
 
-    def write_messages(self, messages, rate=None, transport=None, framed=None):
-        return self.write_raw(self.__format_messages(messages, framed=framed), rate=rate, transport=transport)
+    def write_messages(self, messages, rate=None, transport=None, framed=None, extra_loggen_args=None):
```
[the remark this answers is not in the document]
mitzkia: I agreed with you that `loggen` is just a "tool" to write/send messages (under the hood). But it has too many different (useful) options. I think we can not foretell which option will be needed in the future from testcases.
I also agreed with your idea (proposal), this could work for now (in this specific case). But it also adds an exception how to handle such cases and my feeling is that this will spread for other options as well, and `write_messages()` will contains too many exceptions to handling different `loggen` parameters.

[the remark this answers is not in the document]
mitzkia: I also assume that we will use in 95% write_messages() from testcases without `extra_loggen_args` parameter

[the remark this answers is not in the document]
mitzkia: I will think over your proposal for our actual case and prevent this to be a pattern here. Thanks.

tests/light/src/axosyslog_light/driver_io/network/network_io.py
```diff
@@ -72,6 +72,7 @@ def write_raw(self, raw_content, rate=None, transport=None, extra_loggen_args=No
                 **extra_loggen_args,
             ),
         )
+        return loggen_process
```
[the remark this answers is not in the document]
mitzkia: Agreed with your thoughts, thanks for the proposal. I will go with  it.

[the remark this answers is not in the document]
mitzkia: fixed

tests/light/src/axosyslog_light/driver_io/network/network_io.py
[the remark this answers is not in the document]
mitzkia: fixed

tests/light/src/axosyslog_light/fixtures.py
```diff
@@ -197,7 +197,7 @@ def log_message():
 
 @pytest.fixture
 def version(syslog_ng):
-    return syslog_ng.get_version()
+    return syslog_ng.get_config_version()
```
alltilla, line 200: AxoSyslog's version and config version is not the same, why do we use the config version getter in the version fixture?

tests/light/src/axosyslog_light/syslog_ng_config/statements/sources/network_source.py
```diff
@@ -65,7 +65,17 @@ def write_log(self, message, rate=None, transport=None, framed=None):
...
-        self.io.write_messages_with_proxy_header(proxy_version, src_ip, dst_ip, src_port, dst_port, messages, rate=rate, transport=transport, framed=framed)
+        logge_process = self.io.write_messages_with_proxy_header(proxy_version, src_ip, dst_ip, src_port, dst_port, messages, rate=rate, transport=transport, framed=framed)
+        logge_process.wait()
+        return logge_process
+
+    def write_logs_async(self, messages, rate=None, transport=None, framed=None):
+        return self.io.write_messages(messages, rate=rate, transport=transport, framed=framed)
+
+    def write_logs_with_proxy_header_async(self, proxy_version, src_ip, dst_ip, src_port, dst_port, messages, rate=None, transport=None, framed=None):
+        return self.io.write_messages_with_proxy_header(proxy_version, src_ip, dst_ip, src_port, dst_port, messages, rate=rate, transport=transport, framed=framed)
```
alltilla, line 81: [optional]

We still leak the fact that sending is done with a process. I think it would be better to have the normal and async functions in the IO layer, do the `wait()` calls there, and not return anything.

tests/light/src/axosyslog_light/syslog_ng_config/statements/sources/network_source.py
```diff
@@ -65,7 +65,17 @@ def write_log(self, message, rate=None, transport=None, framed=None):
         self.io.write_messages([message], rate=rate, transport=transport, framed=framed)
 
     def write_logs(self, messages, rate=None, transport=None, framed=None):
-        self.io.write_messages(messages, rate=rate, transport=transport, framed=framed)
+        loggen_process = self.io.write_messages(messages, rate=rate, transport=transport, framed=framed)
+        loggen_process.wait()
```
alltilla, line 69: Now that we made the `write_logs()` call a blocking function, shouldn't we replace all the uses of it to the async counterpart? With the current state of the PR we changed the behavior of all the testcases.

tests/light/src/axosyslog_light/fixtures.py
mitzkia: Yes, that is correct. Thanks for this. I will fix this part.

tests/light/src/axosyslog_light/syslog_ng_config/statements/sources/network_source.py
mitzkia: The original idea was that `write_logs()` will return when all the logs are written due to actually we write only a few logs in each testcase, what does not need extra time. This change has no effect for the current Light testcases.
In my opinion `write_logs_async()` like functions can be used only in special cases when we want to highlight that the write task should work in background. This is how we can differentiate use-cases.

tests/light/src/axosyslog_light/syslog_ng_config/statements/sources/network_source.py
mitzkia: Ok, thanks. I will rewrite this part.
Please check this commit: ec06f8474f92132dc596e900df2506d8f99702ef
Is this what you have mentioned?

tests/light/src/axosyslog_light/fixtures.py
mitzkia: fixed

tests/light/src/axosyslog_light/syslog_ng_config/statements/sources/network_source.py
MrAnno: I think the process is no longer leaked, so I'm resolving this comment.

tests/light/src/axosyslog_light/syslog_ng_config/statements/sources/network_source.py
MrAnno: I don't fully understand this comment. @mitzkia please press resolve if you believe this has been addressed :)

tests/light/src/axosyslog_light/driver_io/network/network_io.py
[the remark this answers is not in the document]
MrAnno: Same about `write_raw()` returning `Loggen`: we should introduce a `Perf` class or use loggen directly from the tests who require such entities.

mitzkia:
@MrAnno Thanks for the review comment. Agreed with you, I have removed the NetworkIO() related changes from this PR.

MrAnno (approved):

tests/light/src/axosyslog_light/driver_io/network/network_io.py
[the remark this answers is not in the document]
MrAnno: We'll talk about this IRL, do everything in a separate PR.

PR #605, opened 2025-04-30
app-transform: fix auto-parse-include

MrAnno:
`auto-parse-include` should extend the filtering, but the run-time filtering on `filterx_app_variable` was unconditional.


modules/appmodel/app-transform-generator.c
```diff
@@ -181,10 +181,13 @@ _generate_non_filterx_only_app_transform(Transformation *transformation, gpointe
   else
     g_string_append(block, "    elif {\n");
 
-  g_string_append_printf(block,
-                         "        filterx { %s === '%s'; };\n",
-                         self->filterx_app_variable,
-                         transformation->super.name);
+  if (!app_object_generator_has_applications_included(&self->super))
```
MrAnno, line 184: We want to skip the filtering only if any of `auto-parse-include()` matches.

PR #607, opened 2025-05-02
ci: fix apt-packages test

MrAnno:
We now have revisions like this:
`4.11.0.new.dict.34.g0443ae7-snapshot+20250423T230904`


mitzkia (approved):
Thanks for the PR. Approved

PR #609, opened 2025-05-03
Move EOF detection and close-on-input to the proto

### LINKED ISSUE #604: TLS 1.3 Client Log handling of Key Update Handshake Records
# AxoSyslog
## Version of AxoSyslog
syslog-ng 4.8.1 w/ OpenSSL 3.0.8

## Platform
Ubuntu 24.04

## Debug bundle

# Issue
The Syslog-NG TLS client log exporter does not obey Handshake Records which are issued by the server once the TLS 1.3 session has been established. This is of particular concern with regard to TLS 1.3 for the Key Update handshake record which can be issued by the server to require the client to update the keys that are used by the client.

## Steps to reproduce
1. Configure syslog-ng to forward logs to a remote syslog server over TLS 1.3.
2. Start a TLS implementation on the remote syslog server (e.g. OpenSSL s_server).
3. Cause the TLS server to issue the 'Key Update' Handshake Record with the 'update_requested' value set.
4. ISSUE HERE: observe that the syslog-ng client does not respond to the 'Key Update' record as required by RFC 8446 (https://datatracker.ietf.org/doc/html/rfc8446#section-4.6.3) 

## Configuration
cat /syslog-ng.conf
#############################################################################
# Default syslog-ng.conf file which collects all local logs into a
# single file called /var/log/messages.
#

@version: 4.8
@include "scl.conf"

source s_local {
        system();
        internal();
};

source s_network {
        default-network-drivers(
                # NOTE: TLS support
                #
                # the default-network-drivers() source driver opens the TLS
                # enabled ports as well, however without an actual key/cert
                # pair they will not operate and syslog-ng would display a
                # warning at startup.
                #
                #tls(key-file("/path/to/ssl-private-key") cert-file("/path/to/ssl-cert"))
        );
};

destination d_local {
        file("/var/log/messages");
        file("/var/log/messages-kv.log" template("$ISODATE $HOST $(format-welf --scope all-nv-pairs)\n") frac-digits(3));
};

log {
        source(s_local);

        # uncomment this line to open port 514 to receive messages
        #source(s_network);
        destination(d_local);
};

destination d_tls {
        network(
                "172.16.0.7"
                port("7169")
                transport(tls)
                tls(
                        ssl-options(no-sslv3,no-sslv2,no-tlsv12,no-tlsv11,no-tlsv1)
                        ca-file("/etc/syslog-ng/ssl/RootCA.crt")
                )
                time-reopen(5)
        );
};

log { source (s_local); destination(d_tls); };

MrAnno:
Fixes the `collectd()` destination (we never read the server's response).
Fixes possible starvation issue on the destination side when reading server responses
Fixes and workarounds TLS-related state handling issues
Fixes #604

### DIFF
--- a/lib/logproto/logproto-client.h
+++ b/lib/logproto/logproto-client.h
@@ -67,6 +67,7 @@ struct _LogProtoClient
   LogProtoStatus status;
   const LogProtoClientOptions *options;
   LogTransportStack transport_stack;
+  gboolean transport_reversed_io_direction;
   gboolean (*poll_prepare)(LogProtoClient *s, GIOCondition *cond, GIOCondition *idle_cond, gint *timeout);
   LogProtoStatus (*post)(LogProtoClient *s, LogMessage *logmsg, guchar *msg, gsize msg_len, gboolean *consumed);
   LogProtoStatus (*process_in)(LogProtoClient *s);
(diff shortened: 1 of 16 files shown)


lib/logproto/logproto-client.h
```diff
@@ -123,39 +124,53 @@ log_proto_client_handshake(LogProtoClient *s, gboolean *handshake_finished)
...
+  if (log_transport_stack_poll_prepare(&self->transport_stack, &transport_cond))
+    goto exit;
+
+  result = self->poll_prepare(self, cond, idle_cond, timeout);
+
+  /* Protos are not allowed to set both IN and OUT because of "transport_reversed_io_direction" and TLS requirements:
+   * "If you get SSL_ERROR_WANT_WRITE from SSL_write() then you should not do any other operation that could trigger
+   * IO other than to repeat the previous SSL_write() call."
+   */
+  g_assert(!((*cond & G_IO_IN) && (*cond & G_IO_OUT)));
```
MrAnno, line 141: @bazsi This is the consequence of moving everything into the abstract class and trying to use the bool logic  you suggested.

I think my original logic was more allowing. I agree it was ugly, but it is not possible to be transport-transparent in the protos without those callbacks we talked about.

MrAnno:
Will update this in the next 2 days.

MrAnno:
@bazsi Please check the last patch, I could solve this layering problem without introducing additional buffers.
I like this approach very much, it's clear in all layers what is happening (though I still consider it a workaround, which will probably stay with us forever).

HofiOne:
@MrAnno As I could see in the changes, the handling of the IO conditions is correct, just [this tiny piece](https://github.com/syslog-ng/syslog-ng/pull/5392/commits/24dedf8ef3ca0cfbd741659ef4ea95f75c5a54a8) is left out, without that, pipe destinations will close the connection continuously on any input detected on the bi-directionally opened pipes.

MrAnno:
The Python script I attached in my PR can show how syslog-ng would steal messages from the real destination if that destination is reading slowly.

A traditional pipe is not a broadcaster mechanism. If you have multiple readers, they will steal data from each other.

HofiOne:
But, this means we should either check that no more than one source is attached to a given pipe destination, and warn the user that message loss can happen if there are multiple readers are in the log path, but probably it can be a different fix

PR #610, opened 2025-05-05
clickhouse: avoid creating symbols for enum values

MrAnno:
Constructing a map with `FieldDescriptorProto::TYPE_*` produces symbols for enum values:

```
U _ZN6google8protobuf20FieldDescriptorProto11TYPE_DOUBLEE
```

This causes a linkage issue with CMake, which I couldn't resolve, but in the end, these values should not appear in the symbol table, so a `static_cast` is applied as a workaround.


bshifter (approved):
not fully understood the error itself to be honest, but this PR fixes it

PR #612, opened 2025-05-06
file-monitor: fix inotify

MrAnno:
1b13018742fa49c7d75cb88b75bdf9e8c89ffe8f removed the basename.

### DIFF
--- a/lib/file-monitor.c
+++ b/lib/file-monitor.c
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2024 Axoflow
- * Copyright (c) 2023 László Várady
+ * Copyright (c) 2024-2025 Axoflow
+ * Copyright (c) 2023-2025 László Várady
  * Copyright (c) 2024 shifter
  *
  * This library is free software; you can redistribute it and/or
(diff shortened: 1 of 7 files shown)


mitzkia (approved):
approved

lib/file-monitor.c
```diff
@@ -168,7 +169,7 @@ static void _inotify_event_handler(void *c, struct inotify_event *event)
 {
   FileMonitor *self = (FileMonitor *) c;
 
-  if (g_strcmp0(self->file_name, event->name) == 0)
+  if (g_strcmp0(self->file_basename, event->name) == 0)
```
orymate, line 172: shouldn't we accept both the basename and the filename? do all event types name the directory we subscribed for?

MrAnno: In this unit, it is always the file name that inotify returns.

orymate: Sorry, I didn't catch that we watch for the dirname and this is one is the basename.  Also it looks like the iv wrapper takes care of checking the dirname (event->wd) when dispatching the callback. But I couldn't find out if watching multiple files in the same directory would work correctly. It might also be worth checking in an e2e test if atomic writes by rename()ing a temporary file to the final place is handled correctly (it should be based on a quick look at the code).

bshifter (approved):

PR #613, opened 2025-05-07
light: test cache_json_file with atomic move too

orymate:
Extend the `cache_json_file` tests with a create+rename style file modification
in addition to the existing plain open+write+close style one.

### DIFF
--- a/tests/light/functional_tests/filterx/test_filterx_cache_json_file.py
+++ b/tests/light/functional_tests/filterx/test_filterx_cache_json_file.py
@@ -21,10 +21,12 @@
 # COPYING for details.
 #
 #############################################################################
+import os
 
 
 def test_filterx_cache_json_file_reloads_its_content_automatically(syslog_ng, config):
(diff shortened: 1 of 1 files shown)


MrAnno (dismissed):

tests/light/functional_tests/filterx/test_filterx_cache_json_file.py
```diff
@@ -45,3 +47,10 @@ def test_filterx_cache_json_file_reloads_its_content_automatically(syslog_ng, co
         file.write('{"msg": "autoupdated"}')
 
     assert destination.read_until_logs(["autoupdated"])
+
```
bazsi, line 50: I'd prefer this to a be a separate testcase if possible.

orymate: updated

MrAnno (approved):

PR #614, opened 2025-05-07
Fix tls wildcard

MrAnno:
Ported from syslog-ng


bazsi:
I am not merging as it is still marked as draft.

PR #615, opened 2025-05-08
transport-tls: fix clearing BIO retry flags

MrAnno:


OverOrion (approved):
Thanks!

For reference: https://github.com/openssl/openssl/blob/10bd6fa8ca93b4cf53f005f110c827ed923c89a4/test/asynciotest.c#L89:#L115

PR #617, opened 2025-05-09
Filterx add in operator

sodomelle:
Example usage:
```
s_arr = json_array(['foo', 'bar', 'asd']);
if ($MSG in s_arr)
```
or
```
s_arr = json_array(['foo', 'bar', 'asd']);
if ($MSG not in s_arr)
```


lib/filterx/expr-membership.c
```diff
@@ -22,28 +22,97 @@
...
+
+  if (!filterx_object_len(list_obj, &size))
+    return NULL;
+
+  for (guint64 i = 0; i < size; i++)
+    {
+      FilterXObject *elt = filterx_list_get_subscript(list_obj, i);
+      const gchar *val;
+
+      if ((filterx_object_extract_string_ref(elt, &val, &len) && len != 0)
```
MrAnno, line 78: `in` would be nice if it could work with any types. `expr-comparison.c` contains the current logic of how we compare values.

In case we want to start with strings only, can we add some error handling (type checking) so that we don't just fail silently?

tests/light/functional_tests/filterx/test_filterx.py
```diff
@@ -2492,3 +2492,79 @@ def test_otel_repr(config, syslog_ng):
...
+
+    assert file_true.get_stats()["processed"] == 1
+    assert "processed" not in file_false.get_stats()
+    assert file_true.read_log() == "found"
+
+
+def test_in_operator_false(config, syslog_ng):
+    (file_true, file_false) = create_config(
+        config, """
+        s_arr = json_array(['foo', 'bar', 'asd']);
```
MrAnno, line 2519: ` s_arr = ['foo', 'bar', 'asd'];`

This should work too as far as I remember. Once we remove generators, `$MSG in [...]` will work too.

lib/filterx/expr-membership.c
sodomelle: Done.

tests/light/functional_tests/filterx/test_filterx.py
sodomelle: Done.

MrAnno (approved):
🚀

PR #619, opened 2025-05-09
filterx: fix precedence and associativity of "not"

MrAnno:
For example:
https://docs.python.org/3/reference/expressions.html#operator-precedence


sodomelle (approved):
Looks good.

PR #621, opened 2025-05-09
Docker perf support

bazsi:
This branch changes our production docker image to include frame pointers in all key dependencies of axosyslog, in order to be able to run perf in production.

This is how it is implemented:
* I added an axosyslog-builder image, which is Alpine Linux that will then be used to build AxoSyslog itself
* axosyslog-builder builds musl, grpc and json-c (the three dependencies that currently seem to be required), the list can be extended further
* it also contains an initial set of axosyslog build dependencies to speed up subsequent axosyslog compilation
* axosyslog is then built using this image

A successful build of axosyslog-builder can be found here: https://github.com/axoflow/axosyslog/actions/runs/15089475848

A snapshot of axosyslog using the builder image is here: https://github.com/axoflow/axosyslog/actions/runs/15089247873

Since it changes github workflows and those can't run in my fork, I  tested it by pushing it to the upstream repo. The last patch needs to be dropped from the series (e.g. the one with "dropme" in its subject) before merging.


bazsi:
After a few iterations this now properly compiles and installs musl with frame pointers so that perf works within our production container.

you can test it via `ghcr.io/axoflow/axosyslog:dev-docker-perf-support-v6`

if you install perf in that container (apk add perf), then `perf record -g /usr/sbin/syslog-ng -F` (or using -p to trace an
existing syslog-ng process), will produce good backtraces that we can use.

I also added the musl-dbg package for now. Before merging let's discuss what the best way is to do this, maybe we should compile our deps into a separate container and then just pull that during building. With that we can even extend what we need to rebuild relatively easily and we would not increase the build time of the production container much.

bazsi:
Right now the latest runs produced by this branch are:

builder image:
https://github.com/axoflow/axosyslog/actions/runs/15193204673

production image (using the builder built above):
https://github.com/axoflow/axosyslog/actions/runs/15189555679

This is the final production image:
ghcr.io/axoflow/axosyslog:dev-docker-perf-support-v12

I tested it locally.  This does not change any code, only the way we build our containers. There's one important thing to check: the local json-c compilation needs to be done using thread support. This is difficult to validate as it will only cause crashes at high parallelism.

docker/alpine-base.dockerfile
```diff
@@ -0,0 +1,84 @@
...
+
+RUN \
+    cd aports/main/jemalloc && \
+    sed -i -e '/^pkgrel/s,[0-9]*$,999,' APKBUILD && \
+    abuild -r
+
+RUN \
+    cd aports/main/json-c && \
+    sed -i -e '/^pkgrel/s,[0-9]*$,999,' -e '/\$CMAKE_CROSSOPTS/s,$, -DENABLE_THREADING=ON,' APKBUILD && \
+    abuild -r
```
alltilla: @MrAnno In a later commit we remove the forked specfile for json-c, and I believe this replaces its functionality. Can you please validate if this build command does everything we need?

dbld/images/alpine.dockerfile
```diff
@@ -0,0 +1,36 @@
...
+ARG ARG_IMAGE_PLATFORM
+ARG COMMIT
+ENV IMAGE_PLATFORM=${ARG_IMAGE_PLATFORM}
+LABEL COMMIT=${COMMIT}
+
+
+#RUN /dbld/builddeps update_packages
+#RUN /dbld/builddeps install_dbld_dependencies
+#RUN /dbld/builddeps install_apt_packages
+#RUN /dbld/builddeps install_debian_build_deps
```
alltilla, line 15: [nitpick]

dead code

alltilla (commented):
Is the builder image push a temporary error? Will it be fixed after merge, and will it work under PRs from forks?

dbld/images/alpine.dockerfile
bazsi: I would drop this patch from this series. I've added docker-builder and docker targets to dbld/rules and with that we can build production images locally now.

docker/alpine-base.dockerfile
bazsi: Here's the APKBUILD file after the sed script above, which was used to compile json-c:

```
# Maintainer: Natanael Copa <ncopa@alpinelinux.org>
pkgname=json-c
pkgver=0.18
pkgrel=999
pkgdesc="A JSON implementation in C"
url="https://github.com/json-c/json-c/wiki"
arch="all"
license="MIT"
makedepends="cmake doxygen samurai"
subpackages="$pkgname-dev $pkgname-doc"
source="https://s3.amazonaws.com/json-c_releases/releases/json-c-$pkgver.tar.gz"

# secfixes:
#   0.14-r1:
#     - CVE-2020-12762

build() {
	if [ "$CBUILD" != "$CHOST" ]; then
		CMAKE_CROSSOPTS="-DCMAKE_SYSTEM_NAME=Linux -DCMAKE_HOST_SYSTEM_NAME=Linux"
	fi
	cmake -B build -G Ninja \
		-DCMAKE_INSTALL_PREFIX=/usr \
		-DCMAKE_INSTALL_LIBDIR=lib \
		-DBUILD_SHARED_LIBS=ON \
		-DBUILD_STATIC_LIBS=ON \
		-DCMAKE_BUILD_TYPE=MinSizeRel \
		$CMAKE_CROSSOPTS -DENABLE_THREADING=ON
	cmake --build build --target all doc
}

check() {
	cmake --build build --target test
}

package() {
	DESTDIR="$pkgdir" cmake --install build
	mkdir -p "$pkgdir"/usr/share/doc/$pkgname
	mv doc/html "$pkgdir"/usr/share/doc/$pkgname
}

sha512sums="
4763f2352414dac3599bc2183b4fa57dbfaac0ca24de890097bd7d0bdda93c91efa280f6566e949e6d94212ef39a63fc76c5f9d0c54ff3d04b13c859717dba5a  json-c-0.18.tar.gz
"
```

NOTE: on the `-DENABLE_THREADING=ON` line above. you can check the same by downloading `ghcr.io/axoflow/axosyslog-builder` and look at the file in `/home/builder/aports/main/json-c`

With that I think we are fine.

bazsi:
> Is the builder image push a temporary error? Will it be fixed after merge, and will it work under PRs from forks?

I've now dropped both the preliminary alpine dbld container as well as the "dropme" patch. Let's see if this is becoming green or remains red.

bazsi:
I have removed building on PRs, as we can't push to the registry from a PR.

alltilla (approved):

bazsi:
Thanks @alltilla

PR #626, opened 2025-05-16
wallclocktime: strptime and strftime fixes

alltilla:


MrAnno (approved):
Who would have thought that August 7th was a Saturday :)

PR #631, opened 2025-05-22
docker: fix warnings

MrAnno:


bazsi:
I fixed these in #621 too, I can resolve the conflicts if this goes in, but I'd prefer to get that in instead if possible.

PR #632, opened 2025-05-23
transport: fix TLS shutdown by reversing transport stack dealloc order

MrAnno:
`log_transport_tls_free_method()` does some network communication.


bazsi:
I think #633 would be a better fix, what do you think?

PR #644, opened 2025-05-27
docker: don't build glib on armv7

bazsi:
As of now the Glib build fails on armv7, which is not really needed for now


MrAnno:
Waiting for https://github.com/axoflow/axosyslog/actions/runs/15279922707 to finish

PR #651, opened 2025-05-29
Filterx compact expr list

bazsi:
This makes the list of expressions more compact after initialization.


bazsi:
I am splitting off the patches related to expr initialization before eval into a separate PR and then I'll get back to here.

bazsi:
There was indeed an issue with the branch (a missing gboolean return value in a foreach callback) that caused the mac build to fail. This is now fixes and this should be ready for review, but in the meanwhile our Debian based
CI is broken.

https://github.com/Snaipe/Criterion/issues/574

and 
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1096481

PR #652, opened 2025-05-30
fix pubsub message_bytes calculation

bshifter:
The message_bytes pointer was being incremented instead of updating the value it points to.  
Issue discovered by Hofi: https://github.com/syslog-ng/syslog-ng/pull/5389/commits


MrAnno (commented):
Thanks. We should write a bugfix news entry if this was released.

alltilla (approved):

PR #654, opened 2025-06-02
filterx: add eval error stack

alltilla:
I chose to print the errors starting from the ones deeper in the actual call chain. For example we have a binary expression which evaluates lhs. If lhs fails, the lhs eval error will be printed sooner, than the binary expression's.

I also print an error index ascending from 1 to the number of errors (e.g. `[1/2]`, `[2/2]`).

I think these are sensible, as we have a lot more other logs around these, where the events happening sooner are logged sooner, so I tried to mimic that, but we can choose a different printing if needed.

---

Size 8 should be enough for a long time, I believe.

---

Also replaced a lot of `msg_error()`s from `eval()` functions to use the eval error instead, so they do not flood stderr, and added a bunch of new error infos, now that we can have multiple for a call chain.

---

Example:
```
source s_sys {
  example-msg-generator(num(1));
};

log {
  filterx {
    list = [];
    list["this will be suppressed"] ?? 42;
    list["this will cause an error"];
  };
};
```
```
[2025-06-04T18:53:53.031293] >>>>>> filterx rule evaluation begin; rule='#anon-filter0', location='/home/alltilla/repos/axosyslog/build/install/etc/syslog-ng.conf:7:10', msg='0x56fc5fe98510', rcptid='3'
[2025-06-04T18:53:53.031293] FILTERX ESTEP; expr='/home/alltilla/repos/axosyslog/build/install/etc/syslog-ng.conf:8:5|  list = []', value='[]', truthy='1', type='ref/list_object'
[2025-06-04T18:53:53.031293] FilterX: null coalesce suppressing error; err_idx='[1/2]', expr='/home/alltilla/repos/axosyslog/build/install/etc/syslog-ng.conf:9:5|      list["this will be suppressed"]', error='Failed to get element from list: Index must be integer, got: string'
[2025-06-04T18:53:53.031293] FilterX: null coalesce suppressing error; err_idx='[2/2]', expr='/home/alltilla/repos/axosyslog/build/install/etc/syslog-ng.conf:9:5|      list["this will be suppressed"]', error='Failed to get-subscript from object: this will be suppressed'
[2025-06-04T18:53:53.031293] FILTERX ESTEP; expr='/home/alltilla/repos/axosyslog/build/install/etc/syslog-ng.conf:9:5|  list["this will be suppressed"] ?? 42', value='42', truthy='1', type='integer'
[2025-06-04T18:53:53.031293] FILTERX ERROR; err_idx='[1/2]', expr='/home/alltilla/repos/axosyslog/build/install/etc/syslog-ng.conf:10:5|        list["this will cause an error"]', error='Failed to get element from list: Index must be integer, got: string'
[2025-06-04T18:53:53.031293] FILTERX ERROR; err_idx='[2/2]', expr='/home/alltilla/repos/axosyslog/build/install/etc/syslog-ng.conf:10:5|        list["this will cause an error"]', error='Failed to get-subscript from object: this will cause an error'
[2025-06-04T18:53:53.031293] <<<<<< filterx rule evaluation result; result='unmatched', rule='#anon-filter0', location='/home/alltilla/repos/axosyslog/build/install/etc/syslog-ng.conf:7:10', dirty='1', msg='0x56fc5fe98510', rcptid='3'
```

### DIFF
--- a/lib/filterx/expr-assign.c
+++ b/lib/filterx/expr-assign.c
@@ -42,21 +42,7 @@ _assign(FilterXBinaryOp *self, FilterXObject *value)
 static inline FilterXObject *
 _suppress_error(void)
 {
-  if (debug_flag)
-    {
-      gint error_count = filterx_eval_get_error_count();
-      gchar buf[FILTERX_EVAL_ERROR_IDX_FMT_SIZE];
-
-      for (gint err_idx = 0; err_idx < error_count; err_idx++)
-        {
-          msg_debug("FilterX: null coalesce assignment suppressing error",
-                    filterx_eval_format_error_index_tag(err_idx, buf),
-                    filterx_eval_format_error_location_tag(err_idx),
-                    filterx_eval_format_error_tag(err_idx));
-        }
-    }
-
-  filterx_eval_clear_errors();
+  filterx_eval_dump_errors("FilterX: null coalesce assignment suppressing error");
 
   return filterx_null_new();
 }
--- a/lib/filterx/expr-null-coalesce.c
+++ b/lib/filterx/expr-null-coalesce.c
@@ -45,23 +45,7 @@ _eval_null_coalesce(FilterXExpr *s)
           && filterx_message_value_get_type(lhs_object) == LM_VT_NULL))
     {
       if (!lhs_object)
-        {
-          if (debug_flag)
-            {
-              gint error_count = filterx_eval_get_error_count();
-              gchar buf[FILTERX_EVAL_ERROR_IDX_FMT_SIZE];
-
-              for (gint err_idx = 0; err_idx < error_count; err_idx++)
-                {
-                  msg_debug("FilterX: null coalesce suppressing error",
-                            filterx_eval_format_error_index_tag(err_idx, buf),
-                            filterx_eval_format_error_location_tag(err_idx),
-                            filterx_eval_format_error_tag(err_idx));
-                }
-            }
-
-          filterx_eval_clear_errors();
-        }
+        filterx_eval_dump_errors("FilterX: null coalesce suppressing error");
 
       FilterXObject *rhs_object = filterx_expr_eval(self->super.rhs);
       if (!rhs_object)
--- a/lib/filterx/filterx-eval.c
+++ b/lib/filterx/filterx-eval.c
@@ -253,6 +253,26 @@ filterx_eval_format_error_location_tag(gint index)
   return filterx_error_format_location_tag(&context->errors[index]);
 }
 
+void
+filterx_eval_dump_errors(const gchar *message)
+{
+  if (debug_flag)
+    {
+      gint error_count = filterx_eval_get_error_count();
+      gchar buf[FILTERX_EVAL_ERROR_IDX_FMT_SIZE];
+
+      for (gint err_idx = 0; err_idx < error_count; err_idx++)
+        {
+          msg_debug(message,
+                    filterx_eval_format_error_index_tag(err_idx, buf),
+                    filterx_eval_format_error_location_tag(err_idx),
+                    filterx_eval_format_error_tag(err_idx));
+        }
+    }
+
+  filterx_eval_clear_errors();
+}
+
 static inline FilterXFailureInfo *
 _new_failure_info_entry(FilterXEvalContext *context)
 {
@@ -300,6 +320,7 @@ filterx_eval_exec(FilterXEvalContext *context, FilterXExpr *expr)
   FilterXObject *res = filterx_expr_eval(expr);
   if (!res)
     {
+      /* Open coded as this function is context specific. */
       if (debug_flag)
         {
           gint error_count = context->error_count;
--- a/modules/metrics-probe/filterx/func-update-metric.c
+++ b/modules/metrics-probe/filterx/func-update-metric.c
@@ -96,21 +96,7 @@ exit:
   if (!success)
     {
       /* It would be nice to introduce a counter for this. */
-      if (debug_flag)
-        {
-          gint error_count = filterx_eval_get_error_count();
-          gchar buf[FILTERX_EVAL_ERROR_IDX_FMT_SIZE];
-
-          for (gint err_idx = 0; err_idx < error_count; err_idx++)
-            {
-              msg_debug("FilterX: update_metric(): eval error, skipping",
-                        filterx_eval_format_error_index_tag(err_idx, buf),
-                        filterx_eval_format_error_location_tag(err_idx),
-                        filterx_eval_format_error_tag(err_idx));
-            }
-        }
-
-      filterx_eval_clear_errors();
+      filterx_eval_dump_errors("FilterX: update_metric() eval error, skipping");
     }
 
   return filterx_boolean_new(TRUE);
--- a/lib/filterx/expr-set-subscript.c
+++ b/lib/filterx/expr-set-subscript.c
@@ -59,21 +59,7 @@ _set_subscript(FilterXSetSubscript *self, FilterXObject *object, FilterXObject *
 static inline FilterXObject *
 _suppress_error(void)
 {
-  if (debug_flag)
-    {
-      gint error_count = filterx_eval_get_error_count();
-      gchar buf[FILTERX_EVAL_ERROR_IDX_FMT_SIZE];
-
-      for (gint err_idx = 0; err_idx < error_count; err_idx++)
-        {
-          msg_debug("FilterX: null coalesce assignment suppressing error",
-                    filterx_eval_format_error_index_tag(err_idx, buf),
-                    filterx_eval_format_error_location_tag(err_idx),
-                    filterx_eval_format_error_tag(err_idx));
-        }
-    }
-
-  filterx_eval_clear_errors();
+  filterx_eval_dump_errors("FilterX: null coalesce assignment suppressing error");
 
   return filterx_null_new();
 }
--- a/lib/filterx/expr-setattr.c
+++ b/lib/filterx/expr-setattr.c
@@ -60,21 +60,7 @@ _setattr(FilterXSetAttr *self, FilterXObject *object, FilterXObject *new_value)
 static inline FilterXObject *
 _suppress_error(void)
 {
-  if (debug_flag)
-    {
-      gint error_count = filterx_eval_get_error_count();
-      gchar buf[FILTERX_EVAL_ERROR_IDX_FMT_SIZE];
-
-      for (gint err_idx = 0; err_idx < error_count; err_idx++)
-        {
-          msg_debug("FilterX: null coalesce assignment suppressing error",
-                    filterx_eval_format_error_index_tag(err_idx, buf),
-                    filterx_eval_format_error_location_tag(err_idx),
-                    filterx_eval_format_error_tag(err_idx));
-        }
-    }
-
-  filterx_eval_clear_errors();
+  filterx_eval_dump_errors("FilterX: null coalesce assignment suppressing error");
 
   return filterx_null_new();
 }
--- a/lib/filterx/filterx-eval.h
+++ b/lib/filterx/filterx-eval.h
@@ -90,6 +90,7 @@ EVTTAG *filterx_eval_format_error_location_tag(gint index);
 void filterx_eval_clear_errors(void);
 EVTTAG *filterx_eval_format_error_index_tag(gint index, gchar *buf);
 EVTTAG *filterx_format_eval_result(FilterXEvalResult result);
+void filterx_eval_dump_errors(const gchar *message);
 
 void filterx_eval_begin_context(FilterXEvalContext *context, FilterXEvalContext *previous_context,
                                 FilterXScope *scope_storage, LogMessage *msg);
--- a/lib/filterx/func-set-fields.c
+++ b/lib/filterx/func-set-fields.c
@@ -213,21 +213,7 @@ _set_with_fallbacks(FilterXObject *dict, FilterXObject *key, GPtrArray *values)
       FilterXObject *value_obj = filterx_expr_eval(value);
       if (!value_obj)
         {
-          if (debug_flag)
-            {
-              gint error_count = filterx_eval_get_error_count();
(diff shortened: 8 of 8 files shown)


lib/filterx/filterx-eval.c
```diff
@@ -58,79 +58,181 @@ filterx_eval_set_context(FilterXEvalContext *context)
...
+      filterx_error_clear(&context->errors[i]);
     }
+
+  context->error_count = 0;
 }
 
 void
 filterx_eval_clear_errors(void)
+{
+  _clear_errors(filterx_eval_get_context());
```
sodomelle: I think we should NULL check the context here.

lib/filterx/filterx-eval.c
```diff
@@ -303,7 +306,8 @@ filterx_eval_begin_context(FilterXEvalContext *context,
 static void
 _failure_info_clear_entry(FilterXFailureInfo *failure_info)
 {
-  filterx_error_clear(&failure_info->error);
+  for (gint i = 0; i < failure_info->error_count; i++)
+    filterx_error_clear(&failure_info->errors[i]);
```
sodomelle, line 388: Set the `failure_info->error_count` to `0`.

modules/metrics-probe/filterx/func-update-metric.c
```diff
@@ -87,7 +87,20 @@ _eval(FilterXExpr *s)
...
+      if (trace_flag)
+        {
+          gint error_count = filterx_eval_get_error_count();
+          gchar buf[FILTERX_EVAL_ERROR_IDX_FMT_SIZE];
+
+          for (gint err_idx = 0; err_idx < error_count; err_idx++)
+            {
+              msg_debug("FilterX: update_metric(): eval error, skipping",
+                        filterx_eval_format_error_index_tag(err_idx, buf),
+                        filterx_eval_format_error_location_tag(err_idx),
```
sodomelle: Isn't it possible to make this a function, and put it somewhere common? I saw this repeated a few times, with the only difference being the first parameter of `msg_debug`.

libtest/filterx-lib.c
```diff
@@ -148,12 +148,83 @@ filterx_test_unknown_object_new(void)
...
+  return NULL;
+}
+
+static gboolean
+_non_literal_init(FilterXExpr *s, GlobalConfig *cfg)
+{
+  FilterXNonLiteralExpr *self = (FilterXNonLiteralExpr *) s;
+  if (!filterx_expr_init(self->block, cfg))
+    {
+      filterx_expr_deinit(self->block, cfg);
```
sodomelle: I don't think this deinit is needed.

lib/filterx/tests/Makefile.am
```diff
@@ -137,3 +138,6 @@ lib_filterx_tests_test_scope_LDADD   = $(TEST_LDADD) $(JSON_LIBS)
 
 lib_filterx_tests_test_func_format_json_CFLAGS	= $(TEST_CFLAGS)
 lib_filterx_tests_test_func_format_json_LDADD	= $(TEST_LDADD)
+
+lib_filterx_tests_test_filterx_eval_CFLAGS	= $(TEST_CFLAGS)
+lib_filterx_tests_test_filterx_eval_LDADD	= $(TEST_LDADD)
```
sodomelle: nitpick: I enjoy having new lines at the end of files.

lib/filterx/tests/test_filterx_eval.c
```diff
@@ -0,0 +1,174 @@
+/*
+ * Copyright (c) 2025 Axoflow
+ * Copyright (c) 2024 Attila Szakacs <attila.szakacs@axoflow.com>
```
sodomelle: Date seems off. Or we have a time traveler. Noice :D

sodomelle (commented):
I reviewed another chunk.
Next commit to check is:  filterx: use eval_error instead msg_error, where sensible

lib/filterx/expr-arithmetic-operators.c
```diff
@@ -56,11 +65,19 @@ _eval_arithmetic_operators_common(FilterXArithmeticOperator *self, GenericNumber
...
 
-  if(!filterx_object_extract_generic_number(rhs_object, rhs_number))
+  if (!filterx_object_extract_generic_number(rhs_object, rhs_number))
     {
-      filterx_object_unref(rhs_object);
+      gchar type_name_buf[FILTERX_OBJECT_TYPE_NAME_BUF_SIZE];
+      gchar *info = g_strdup_printf("right hand side must be a double or integer, got: %s. ",
+                                    filterx_object_format_type_name(lhs_object, type_name_buf));
+      filterx_eval_push_error_info("Failed to evaluate arithmetic operator", &self->super.super, info, TRUE);
+      filterx_object_unref(lhs_object);
```
sodomelle, line 79: FYI: modulo operator doesn't use the common part of eval, because it only works with integers. Just a note that it might worth adding error logs to `_eval_modulo` as well. I did this. I'm sorry.

lib/filterx/tests/test_filterx_expr.c
```diff
@@ -312,6 +312,23 @@ Test(filterx_expr, test_filterx_set_subscript)
...
+static gboolean
+_is_string_in_filterx_eval_errors(const gchar *error_str)
+{
+  gint error_count = filterx_eval_get_error_count();
+  if (!error_count)
+    return FALSE;
+
+  for (gint i = 0; i < error_count; i++)
+    {
+      printf("%s\n", filterx_eval_get_error(i));
```
sodomelle: Is this `printf` intentional?

lib/filterx/filterx-eval.c
alltilla: We don't need to clear out the fields here, as in `_clear_failure_info()` we release each `FilterXFailureInfo` (g_array_set_size(failure_info, 0);).

In `context->failure_info = g_array_sized_new(FALSE, TRUE, sizeof(FilterXFailureInfo), FAILURE_INFO_PREALLOC_SIZE);` (second argument) we ask the array to zero out each new element.

So at the end of `_clear_failure_info()` we set the array size to 0, the elements will still hold values, but as soon as we increase the size from 0, we will get a zeroed out element.

---

I checked out this behavior in a quick unit test locally:
```
  GArray *failure_info = g_array_sized_new(FALSE, TRUE, sizeof(FilterXFailureInfo), 0);

  g_array_set_size(failure_info, 1);
  FilterXFailureInfo *entry = &g_array_index(failure_info, FilterXFailureInfo, 0);
  entry->error_count = 12;
  cr_assert(entry->error_count == 12);

  g_array_set_size(failure_info, 0);
  cr_assert(entry->error_count == 12);

  g_array_set_size(failure_info, 1);
  cr_assert(entry->error_count == 0);
```
It passed.

modules/metrics-probe/filterx/func-update-metric.c
alltilla: We can, but I prefer to keep `msg_...` calls direct, as it is a lot easier to find them in the code with grep.
I think we won't have many more calls like this.

lib/filterx/tests/test_filterx_expr.c
alltilla: No, fixed.

lib/filterx/expr-arithmetic-operators.c
alltilla: Ah, I didn't catch it! Thanks for reminding me, fixed.

lib/filterx/tests/test_filterx_eval.c
alltilla: Thanks, updated :)

lib/filterx/tests/Makefile.am
alltilla: Me too, added a newline.

libtest/filterx-lib.c
alltilla: True, removed.

lib/filterx/filterx-eval.c
alltilla: Nice catch, added a NULL check.

lib/filterx/filterx-eval.c
sodomelle: I see, okay!

lib/filterx/expr-arithmetic-operators.c
sodomelle: Thank you.

modules/metrics-probe/filterx/func-update-metric.c
sodomelle: Makes sense, I agree.

sodomelle (approved):
As far as I can tell, this looks good. My questions and findings were fixed and/or answered.

sodomelle (dismissed):

lib/filterx/filterx-eval.c
```diff
@@ -57,47 +57,65 @@ filterx_eval_set_context(FilterXEvalContext *context)
...
-    {
-      if (free_info)
-        g_free(info);
-    }
+  filterx_error_set_values(error, message, expr, NULL);
+  filterx_error_set_info(error, info, free_info);
+
+  _log_to_stderr_if_needed(context, error);
+
+  if (!context && free_info)
```
bazsi: I think this last if should be improved a bit to make it more readable. 

what about
```
  if (error == &local_error)
    filterx_error_clear(error);
```

or even extract this into a local function and also call that from the rest of the functions, to make it more future proof.

lib/filterx/expr-assign.c
```diff
@@ -42,7 +42,20 @@ _assign(FilterXBinaryOp *self, FilterXObject *value)
...
 {
-  msg_debug("FILTERX null coalesce assignment supressing error", filterx_eval_format_last_error_tag());
+  if (trace_flag)
+    {
+      gint error_count = filterx_eval_get_error_count();
+      gchar buf[FILTERX_EVAL_ERROR_IDX_FMT_SIZE];
+
+      for (gint err_idx = 0; err_idx < error_count; err_idx++)
+        {
+          msg_debug("FilterX: null coalesce assignment suppressing error",
```
bazsi: is this trace or debug level? it used to be debug now it's actually trace due to the conditional.

lib/filterx/filterx-eval.c
alltilla: I have refactor it like this. Not the cleanest, I agree. What do you think?

```c
/* takes ownership of info */
void
filterx_eval_push_error_info(const gchar *message, FilterXExpr *expr, gchar *info, gboolean free_info)
{
  FilterXEvalContext *context = filterx_eval_get_context();
  FilterXError local_error;
  FilterXError *error = _init_local_or_context_error(context, &local_error);

  /* We might have a NULL FilterXError * if the error stack has filled up. */
  if (!error)
    goto exit;

  filterx_error_set_values(error, message, expr, NULL);
  filterx_error_set_info(error, info, free_info);

  _backfill_error_expr(context);
  _log_to_stderr_if_needed(context, error);

exit:
  if (!free_info)
    return;

  gboolean is_error_in_stack = context && error;
  if (!is_error_in_stack)
    g_free(info);
}
```

lib/filterx/expr-assign.c
alltilla: Nice catch, fixed.

lib/filterx/filterx-eval.c
bazsi: I was wondering if error == &local_error would be a better indication, especially if we could factor it into a separate function.

e.g. _clear_local_error(context, &local_error); that mirrors the _init_local_or_context_error() function and does the cleanup required.

lib/filterx/expr-null-coalesce.c
```diff
@@ -46,11 +46,23 @@ _eval_null_coalesce(FilterXExpr *s)
...
+
+              for (gint err_idx = 0; err_idx < error_count; err_idx++)
+                {
+                  msg_debug("FilterX: null coalesce suppressing error",
+                            filterx_eval_format_error_index_tag(err_idx, buf),
+                            filterx_eval_format_error_location_tag(err_idx),
+                            filterx_eval_format_error_tag(err_idx));
+                }
+            }
+
```
bazsi: I feel a bit bad about the open-coded loop here as I am sure it will get repeated a number of times further down.

lib/filterx/expr-membership.c
```diff
@@ -51,8 +52,8 @@ _eval_in(FilterXExpr *s)
   filterx_object_unref(rhs_obj);
   if (!filterx_object_is_type(list_obj, &FILTERX_TYPE_NAME(list)))
     {
-      msg_error("FilterX: in operator right hand side must be list type",
-                evt_tag_str("type", list_obj->type->name));
+      gchar *info = g_strdup_printf("Right hand side must be list type, got: %s", list_obj->type->name);
+      filterx_eval_push_error_info("Failed to evaluate 'in' operator", &self->super.super, info, TRUE);
```
bazsi, line 59: we might want to add a push_error_info() variant that takes a format string instead of having to do it in the caller.

lib/filterx/expr-null-coalesce.c
alltilla: @sodomelle had the same fear, I think I will lean to doing that now.

lib/filterx/expr-membership.c
alltilla: Yeah that would be a good idea. I'll check how much work that needs tomorrow. If it is quick I'll do it, otherwise let's do it in another PR later.

lib/filterx/filterx-eval.c
alltilla: Refactored it again in a similar manner that you mentioned.

lib/filterx/expr-membership.c
alltilla: I won't have time to check this today unfortunately, so I would rather do this in another PR.

bshifter (approved):

alltilla:
Talked about this IRL, and we concluded that the PR is good enough now, the formatting improvements can go to another PR.

Also concluded that all comments can be resolved and the PR merged.

PR #660, opened 2025-06-06
Light improved crash detection

mitzkia:
This PR helps to identify axosyslog crash from Light testcases


MrAnno (approved):
I think we could check the exit code of `syslog-ng`. In that case , we won't need `bail out`-like greps and we could catch more error types.

PR #662, opened 2025-06-06
FilterX ref count type clarification and multiple bugfixes

MrAnno:
There were multiple problems with the frozen/hibernated object duality:

- hibernate() called the freeze() virtual method, causing confusion and possibly incorrect implementations:
  - FilterXRef implemented freeze() by calling filterx_object_freeze() on the inside value, but
    that call may have been coming from a hibernate() invocation.
  - Dict and list implementations did the same: they froze the key-values even if we were hibernating the dict
  - string implemented the freeze() method in a way that literal strings coming from the configuration were
    cached both in FilterXGlobalCache and FilterXConfig, but after a few reloads, FilterXConfig probably freed
    the objects that remained in FilterXGlobalCache:

```
strcmp (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
g_str_equal (in /usr/lib/libglib-2.0.so.0.8200.5)
g_hash_table_lookup (in /usr/lib/libglib-2.0.so.0.8200.5)
_string_freeze.lto_priv.0 (object-string.c:183)
UnknownInlinedFun (filterx-object.c:136)
UnknownInlinedFun (filterx-object.c:186)
filterx_object_freeze (filterx-object.c:173)
filterx_config_freeze_object (filterx-config.c:66)
construct_template_expr (filterx-grammar.y:74)
```

and

```
ERROR:./lib/filterx/filterx-object.h:308:filterx_object_ref: code should not be reached
Bail out! ERROR:./lib/filterx/filterx-object.h:308:filterx_object_ref: code should not be reached
```

These could have been fixed by introducing the `preserve()` and `thaw()` virtual methods, but there are other problems as well:

- the description in filterx-object says that hibernated objects have a lifetime bound to the config or to the syslog-ng process, and frozen objects can be unfrozen during runtime, but:
  - the configuration still uses freezing (filterx_config_freeze_object), not hibernating
  - _filterx_ref_unfreeze() has assumptions that are no longer true and there is no clean way to fix this
  - json-cache-file() does not use the ref-count behavior of freeze(), and I believe it can never do it safely
- dict() and list() shouldn't even be hibernated or frozen, their recursive nature would cause various complications
  that the current code does not handle correctly (more bugs).

----

So instead of trying to fix the current complex behavior, this PR tries to clarify and simplify reference types by concentrating on the storage during these operations:
- hibernated objects are stored until syslog-ng stops, they are global, the storage is handled by the callers
- frozen objects are stored until the configuration is reloaded, the storage is handled by the freeze() call
- recursive/mutable types can't be frozen or hibernated anymore
- deduplication of immutable values is now done explicitly, without hiding it inside the freeze() implementation
  (this is what actually fixes the crashes I mentioned).

TODO:

- [ ] reimplement `cache_json_file()` value deduplication using `filterx_object_dedup()`.

### DIFF
--- a/lib/filterx/filterx-eval.h
+++ b/lib/filterx/filterx-eval.h
@@ -138,7 +138,7 @@ static inline void
 filterx_eval_store_weak_ref(FilterXObject *object)
 {
   /* Preserved objects do not need weak refs. */
-  if (object && filterx_object_is_preserved(object))
+  if (object && (filterx_object_is_preserved(object) || filterx_object_is_readonly(object)))
     return;
 
   if (object && !object->weak_referenced)
--- a/lib/filterx/filterx-object.h
+++ b/lib/filterx/filterx-object.h
@@ -118,7 +118,7 @@ filterx_type_is_cowable(FilterXType *type)
  *
  *    The ref_cnt is always equal to FILTERX_OBJECT_REFCOUNT_STACK
  *
- * 3) hybernated objects
+ * 3) hibernated objects
  *
  *    Hybernated objects are preserved for the entire lifecycle of a
  *    filterx-based configuration.  They are allocated at startup or when
@@ -131,9 +131,9 @@ filterx_type_is_cowable(FilterXType *type)
  *    these objects can be replace the unfrozen version of the same object by
  *    filterx_object_freeze() (e.g. when they represent the same string).
  *
- *    The ref_cnt will always be: FILTERX_OBJECT_REFCOUNT_HYBERNATED
+ *    The ref_cnt will always be: FILTERX_OBJECT_REFCOUNT_HIBERNATED
  *
- *    NOTE: hybernated objects are considered "preserved"
+ *    NOTE: hibernated objects are considered "preserved"
  *
  * 4) frozen objects
  *
@@ -165,13 +165,13 @@ enum
   FILTERX_OBJECT_REFCOUNT_BARRIER_MAX=FILTERX_OBJECT_REFCOUNT_BARRIER + 1023,
   /* stack based allocation */
   FILTERX_OBJECT_REFCOUNT_STACK,
-  /* anything above this point is preserved: either hybernated or frozen */
+  /* anything above this point is preserved: either hibernated or frozen */
   FILTERX_OBJECT_REFCOUNT_PRESERVED,
 
-  /* hybernated object (considered preserved) */
-  FILTERX_OBJECT_REFCOUNT_HYBERNATED=FILTERX_OBJECT_REFCOUNT_PRESERVED,
+  /* hibernated object (considered preserved) */
+  FILTERX_OBJECT_REFCOUNT_HIBERNATED=FILTERX_OBJECT_REFCOUNT_PRESERVED,
 
-  /* hybernated object (considered preserved) */
+  /* hibernated object (considered preserved) */
   FILTERX_OBJECT_REFCOUNT_FROZEN,
 };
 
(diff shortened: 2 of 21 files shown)


lib/filterx/filterx-eval.h
```diff
@@ -138,7 +138,7 @@ static inline void
 filterx_eval_store_weak_ref(FilterXObject *object)
 {
   /* Preserved objects do not need weak refs. */
-  if (object && filterx_object_is_preserved(object))
+  if (object && (filterx_object_is_preserved(object) || filterx_object_is_readonly(object)))
```
MrAnno, line 141: Is this correct?

What if I put the subobject of a readonly dict into a writable dict?

lib/filterx/func-cache-json-file.c
```diff
@@ -223,15 +183,11 @@ _load_json_file_version(FilterXFunctionCacheJsonFile *self, GError **error)
     {
       return FALSE;
     }
-  _archive_frozen_objects(self);
-  self->frozen_objects = g_ptr_array_new_with_free_func((GDestroyNotify) filterx_object_unfreeze_and_free);
-  filterx_object_freeze(&cached_json);
-  g_ptr_array_add(self->frozen_objects, cached_json);
+  filterx_object_make_readonly(cached_json);
   g_atomic_pointer_set(&self->cached_json, cached_json);
```
MrAnno: `unref()`

MrAnno:
I'd like to ask at least 2 reviewers to check this.

lib/filterx/filterx-ref.c
```diff
@@ -289,7 +289,7 @@ FilterXObject *
 _filterx_ref_new(FilterXObject *value)
 {
 #if SYSLOG_NG_ENABLE_DEBUG
-  if (!filterx_object_is_cowable(value) || filterx_object_is_ref(value))
+  if (!value->type->is_mutable || filterx_object_is_ref(value))
```
alltilla, line 282: This is an equivalent change, right?

lib/filterx/filterx-object.h
```diff
@@ -225,10 +225,8 @@ _filterx_object_is_type(FilterXObject *object, FilterXType *type)
 static inline gboolean
 filterx_object_is_type(FilterXObject *object, FilterXType *type)
 {
-#if SYSLOG_NG_ENABLE_DEBUG
-  if (filterx_type_is_cowable(type) && filterx_object_is_ref(object))
+  if (type->is_mutable && filterx_object_is_ref(object))
```
alltilla, line 214: Is the DEBUG check removed on purpose?

lib/filterx/filterx-object.h
```diff
@@ -257,41 +239,8 @@ filterx_object_ref(FilterXObject *self)
...
-       * If less than 1024 CPUs lose the race, then the refcount would end
-       * up in the range between FILTERX_OBJECT_REFCOUNT_OFLOW_MARK and
-       * FILTERX_OBJECT_REFCOUNT_STACK, causing the assertion at the end of
-       * this function to trigger an abort.
-       *
-       * The non-racy solution would be to use a
-       * g_atomic_int_exchange_and_add() call and checking the old_value
-       * against FILTERX_OBJECT_REFCOUNT_OFLOW_MARK another time, but that's
-       * an extra conditional in a hot-path.
-       */
```
alltilla, line 290: Why did we remove this safety net? I think we can still hit this case with the cached json object for example. Do we consider this adding too much complexity while having nearly zero chance? If so, IMO a comment should still be added here to explain this calculated risk.

MrAnno: It was useful when the original refcount was divided by 2, but without that, it's just 3 less than MAXINT, so it's basically a traditional refcount.

lib/filterx/filterx-object.h
MrAnno: Yes, I believe we won't really find those bugs in dev builds (we don't have enough tests and we don't spend enough time just "trying things out"), we need it in production. If this has a measurable perf impact, let's remove it later.

lib/filterx/filterx-ref.c
MrAnno: No, I believe I fixed a NOOP, because:
```
static inline gboolean
filterx_type_is_cowable(FilterXType *type)
{
  return type->is_mutable && !filterx_type_is_ref(type);
}
```

MrAnno: A yes, here it is equivalent. In your next comment, it was a NOOP.

bshifter (dismissed):

PR #665, opened 2025-06-10
pipe: fix of unexpected input errors on pipe destinations

MrAnno:
Cherry-picked from syslog-ng/syslog-ng#5392


MrAnno:
I think this is not correct, we'll be reading or own writes. We should disable `G_IO_IN` when using it with pipes or with the program destination.

I can reproduce message loss easily:

```
import time
import os

pipe_path = '/tmp/asd'
print(f"Opening named pipe {pipe_path} for reading...")
with open(pipe_path, 'r') as pipe:
    print("Reading from pipe:")
    for line in pipe:
        print("Sleeping for 5 seconds...")
        time.sleep(5)
        print(f"Received: {line.strip()}")
```

alltilla (approved):

lib/logproto/logproto-text-client.c
```diff
@@ -84,6 +96,14 @@ log_proto_text_client_process_input(LogProtoClient *s)
   return LPS_SUCCESS;
 }
 
+static LogProtoStatus
+_prohibit_in(LogProtoClient *s)
```
sodomelle, line 100: Nitpick: static function names in this file don't start with '_' character. Is there a reason for that?

MrAnno: `static` functions can have shorter non-prefixed names if they start with an underscore.

I usually use short names in general, and long prefixed names when the short name would be ambiguous, for example, when settings a breakpoint in `gdb`.

I prefer `classname_init()` over _`init()`, but `_fairly_unique_operation()` over `classname_fairly_unique_operation()`.

sodomelle: Makes sense. Thanks for clarification.

sodomelle (approved):
Thanks for the clarification. Looks good to me.

PR #672, opened 2025-06-16
Light: Improved crash detection

mitzkia:
After this PR:
- in case of those functional tests where we want to check if syslog-ng could not start we also check the exception error message, which help to validate more precisely why syslog-ng could not start
- Light will check syslog-ng's return code more stricter, instead of checking various patterns in stdout or stderr outputs (which can be different in various platforms).

### DIFF
--- a/tests/light/src/axosyslog_light/syslog_ng/syslog_ng.py
+++ b/tests/light/src/axosyslog_light/syslog_ng/syslog_ng.py
@@ -28,7 +28,6 @@ from subprocess import Popen
 
 from axosyslog_light.common.blocking import wait_until_false
 from axosyslog_light.common.blocking import wait_until_true
-from axosyslog_light.common.random_id import get_unique_id
 from axosyslog_light.syslog_ng.console_log_reader import ConsoleLogReader
 from axosyslog_light.syslog_ng.syslog_ng_executor import SyslogNgExecutor
 from axosyslog_light.syslog_ng.syslog_ng_executor import SyslogNgStartParams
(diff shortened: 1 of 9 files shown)


tests/light/src/axosyslog_light/syslog_ng/syslog_ng.py
```diff
@@ -226,29 +228,9 @@ def __start_syslog_ng(self) -> None:
...
-                )
-                core_file.replace(Path(core_file))
-            if core_file_found:
-                raise Exception("syslog-ng core file was found and processed")
-            if self._process.returncode in [-6, -9, -11]:
-                ret_code = self._process.returncode
-                self._process = None
-                raise Exception("syslog-ng process crashed with signal {}".format(ret_code))
+    def __validate_returncode(self, returncode: int):
+        if returncode not in [0, 2]:
```
MrAnno: 1: error in reading configuration file, invalid command line arguments, etc.
2: init-time or persist failure

MrAnno (dismissed):

tests/light/src/axosyslog_light/syslog_ng/syslog_ng.py
sodomelle: For example: test_regexp_search_error_in_pattern fails with error code 2. And the content of the test indicates that it should fail.

mitzkia: I have extended the list with return code 1 also, because we can write a testcase when after a syslog-ng start we set a new config with syntax error for which we would like to assert

sodomelle: Error code 1 also added as non crash returncode, after discussion with @mitzkia.

sodomelle: Error code 1 is also added here.

PR #673, opened 2025-06-16
filterx/dict: fix crash when iterator receive NULL for value parameter

sodomelle:
If the value parameter is NULL for `_format_and_append_dict_elem`, then the underlying `format_json` crashed.

Input that caused the crash:
`$MSG = regexp_search("bar", "(?<f>foo)?(?<b>bar)?");`


lib/filterx/object-dict-interface.c
```diff
@@ -191,6 +191,9 @@ _format_and_append_dict_elem(FilterXObject *key, FilterXObject *value, gpointer
   GString *json = (GString *) args[0];
   gboolean *first = (gboolean *) args[1];
 
+  if(!value)
```
MrAnno: I think we shouldn't allow null-pointers in dict values (FilterX null values are okay).

MrAnno (approved):

PR #676, opened 2025-06-17
axosyslog: change license to GPL 3+

bazsi:
This branch unifies the licensing terms of all code in this repository to the General Public License and at the same time updates it to version 3 (or later) of that license.

According to the LGPL 2.1 section 3

 3.  You may opt to apply the terms of the ordinary GNU General Public License instead of this License to a given copy of the Library.  To do this, you must alter all the notices that refer to this License, so that they refer to the ordinary GNU General Public License, version 2, instead of to this License.  (If a newer version than version 2 of the ordinary GNU General Public License has appeared, then you can specify that version instead if you wish.) Do not make any other change in these notices.

Once this change is made in a given copy, it is irreversible for that copy, so the ordinary GNU General Public License applies to all subsequent copies and derivative works made from that copy.

This option is useful when you wish to copy part of the code of the Library into a program that is not a library.


bazsi:
todo: light

MrAnno:
The license headers were updated via a script due to the large number of files, so errors may have slipped in. I reviewed the changes to the best of my human abilities, and the copyright check also validated them programmatically. I went through the files multiple times to ensure the copyright lines remained in place. If I happened to miss anything, please let us know and we'll fix it immediately.

MrAnno (approved):

PR #678, opened 2025-06-17
filterx: `protobuf_message()`

alltilla:
This PR adds a lot of refactors around gRPC schema loading and protobuf field conversion.
It also adds a `protobuf_message()` filterx function that can generate arbitrary protobuf data with the help of a proto file.
It also adds the `proto_var()` option to `clickhouse()` and `bigquery()`

The code is a refactored version of https://github.com/bshifter/axosyslog/commit/d221d3d0d1c8aa5aeb3e6efc6957b99e96f55f60 by @bshifter.

### DIFF
--- a/modules/cloud-auth/cloud-auth.hpp
+++ b/modules/cloud-auth/cloud-auth.hpp
@@ -38,6 +38,8 @@ public:
 }
 }
 
+#include "compat/cpp-start.h"
+
 struct _CloudAuthenticator
 {
   syslogng::cloud_auth::Authenticator *cpp;
@@ -45,5 +47,7 @@ struct _CloudAuthenticator
   gboolean (*init)(CloudAuthenticator *s);
   void (*free_fn)(CloudAuthenticator *s);
 };
+
+#include "compat/cpp-end.h"
 
 #endif
--- a/modules/examples/sources/random-choice-generator/random-choice-generator.hpp
+++ b/modules/examples/sources/random-choice-generator/random-choice-generator.hpp
@@ -81,6 +81,8 @@ private:
 }
 }
 
+#include "compat/cpp-start.h"
+
 struct RandomChoiceGeneratorSourceWorker_
 {
   LogThreadedSourceWorker super;
@@ -92,5 +94,7 @@ struct RandomChoiceGeneratorSourceDriver_
   LogThreadedSourceDriver super;
   syslogng::examples::random_choice_generator::SourceDriver *cpp;
 };
+
+#include "compat/cpp-end.h"
 
 #endif
--- a/modules/grpc/common/credentials/grpc-credentials-builder.hpp
+++ b/modules/grpc/common/credentials/grpc-credentials-builder.hpp
@@ -97,6 +97,8 @@ private:
 }
 }
 
+#include "compat/cpp-start.h"
+
 struct GrpcServerCredentialsBuilderW_
 {
   syslogng::grpc::ServerCredentialsBuilder *self;
@@ -106,5 +108,7 @@ struct GrpcClientCredentialsBuilderW_
 {
   syslogng::grpc::ClientCredentialsBuilder *self;
 };
+
+#include "compat/cpp-end.h"
 
 #endif
--- a/modules/grpc/common/grpc-dest-worker.hpp
+++ b/modules/grpc/common/grpc-dest-worker.hpp
@@ -63,6 +63,8 @@ protected:
 }
 }
 
+#include "compat/cpp-start.h"
+
 GrpcDestWorker *grpc_dw_new(GrpcDestDriver *o, gint worker_index);
 
 struct GrpcDestWorker_
@@ -70,5 +72,7 @@ struct GrpcDestWorker_
   LogThreadedDestWorker super;
   syslogng::grpc::DestWorker *cpp;
 };
+
+#include "compat/cpp-end.h"
 
 #endif
--- a/modules/grpc/common/grpc-dest.hpp
+++ b/modules/grpc/common/grpc-dest.hpp
@@ -246,6 +246,8 @@ protected:
 }
 }
 
+#include "compat/cpp-start.h"
+
 struct GrpcDestDriver_
 {
   LogThreadedDestDriver super;
@@ -253,6 +255,8 @@ struct GrpcDestDriver_
 };
 
 GrpcDestDriver *grpc_dd_new(GlobalConfig *cfg, const gchar *stats_name);
+
+#include "compat/cpp-end.h"
 
 #endif
 
--- a/modules/grpc/common/grpc-source-worker.hpp
+++ b/modules/grpc/common/grpc-source-worker.hpp
@@ -49,6 +49,8 @@ public:
 }
 }
 
+#include "compat/cpp-start.h"
+
 GrpcSourceWorker *grpc_sw_new(GrpcSourceDriver *o, gint worker_index);
 
 struct GrpcSourceWorker_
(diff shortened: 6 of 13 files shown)


modules/grpc/common/filterx/protobuf-field-converter.cpp
```diff
@@ -93,21 +107,35 @@ class i32FieldConverter : public ProtobufFieldConverter
 
   bool set(Message *message, ProtoReflectors reflectors, FilterXObject *object, FilterXObject **assoc_object)
   {
-    gint64 i;
-    if (!filterx_object_extract_integer(object, &i))
+    try
+      {
+        reflectors.reflection->SetInt32(message, reflectors.fieldDescriptor, this->extract(object));
+      }
+    catch (const std::exception &e)
```
MrAnno, line 135: Can we catch the exact exception type instead of everything?

modules/grpc/common/filterx/protobuf-field-converter.cpp
```diff
@@ -397,6 +501,20 @@ class BytesFieldConverter : public ProtobufFieldConverter
       }
     return true;
   }
+
+  bool add(Message *message, ProtoReflectors reflectors, FilterXObject *object)
+  {
+    try
+      {
+        reflectors.reflection->AddString(message, reflectors.fieldDescriptor, this->extract(object));
```
MrAnno: It was an older decision, but is it safe to use string for bytes?

modules/grpc/common/filterx/protobuf-field-converter.hpp
```diff
@@ -116,6 +117,59 @@ class ProtobufFieldConverter
...
+                return false;
+              }
+
+            filterx_object_unref(elem);
+          }
+
+        *assoc_object = filterx_object_ref(object);
+        return true;
+      }
+    catch(const std::exception &ex)
```
MrAnno, line 166: What exceptions are we expecting here?

If `add()` does not and will not cross C/C++ boundaries, can't we just let them throw an exception instead of doing C-like error handling inside of them?

modules/grpc/common/schema/grpc-schema.hpp
```diff
@@ -75,26 +75,95 @@ struct NameValueTemplatePair
...
 
     return *this;
   }
 
 };
 
+class ProtoSchemaProvider
+{
+public:
+  virtual bool init() = 0;
```
MrAnno: The existence of a pure virtual `init()` method suggests that something may be wrong with the class hierarchy (the name sounds like a virtualized constructor).

Can we rename this method or fix the hierarchy so that constructors can do their work without dynamic dispatch?

modules/grpc/common/filterx/filterx-protobuf-formatter.cpp
```diff
@@ -0,0 +1,101 @@
...
+        }
+      else
+        {
+          if (!converter->set(message, field_name, value, &assoc_object))
+            return FALSE;
+        }
+
+      filterx_object_unref(assoc_object);
+    }
+  catch (const std::exception &e)
```
MrAnno, line 68: Optional: I think the scope of this and similar try-catch blocks could be reduced.

modules/grpc/common/filterx/filterx-protobuf-formatter.cpp
```diff
@@ -0,0 +1,101 @@
...
+  FilterXObject *object_unwrapped = filterx_ref_unwrap_ro(object);
+  if (!filterx_object_is_type(object, &FILTERX_TYPE_NAME(dict)))
+    {
+      throw std::runtime_error("Expected a dictionary object for protobuf formatting");
+    }
+
+  google::protobuf::Message *message = this->proto_schema_file_loader.get_schema_prototype().New();
+
+  gpointer user_data = static_cast<gpointer>(message);
+  gboolean success = filterx_dict_iter(object_unwrapped, this->format_element, user_data);
```
MrAnno: I'm not sure this is defined behavior, `format_element()` is a C++ method even if it's static. Can we write an extern C wrapper for the iterator function just to be sure?

modules/grpc/common/filterx/protobuf-field-converter.cpp
```diff
@@ -517,6 +518,156 @@ class BytesFieldConverter : public ProtobufFieldConverter
...
+
+      filterx_object_unref(assoc_key_object);
+
+      FilterXObject *assoc_val_object = NULL;
+      if (!value_converter->set(elem_message, value_name, value, &assoc_val_object))
+        return FALSE;
+
+      filterx_object_unref(assoc_val_object);
+    }
+  catch (const std::exception &e)
```
MrAnno, line 594: Optional: more specific exception type + reduced try-catch scope here and similar places?

modules/grpc/common/filterx/protobuf-field-converter.cpp
```diff
@@ -517,6 +518,156 @@ class BytesFieldConverter : public ProtobufFieldConverter
...
+
+  FilterXObject *dict = filterx_ref_unwrap_ro(object);
+  if (!filterx_object_is_type(dict, &FILTERX_TYPE_NAME(dict)))
+    {
+      log_type_error(reflectors, object->type->name);
+      return false;
+    }
+
+  gpointer user_data[] = { message, &reflectors };
+  return filterx_dict_iter(dict, MapFieldConverter::add_elem, user_data);
```
MrAnno: extern C wrapper instead?

modules/grpc/common/grpc-dest.cpp
```diff
@@ -288,6 +289,13 @@ grpc_dd_set_protobuf_schema(LogDriver *d, const gchar *proto_path, GList *values
   schema->set_protobuf_schema(proto_path, values);
 }
 
+void
+grpc_dd_set_proto_var(LogDriver *d, LogTemplate *proto_var)
+{
+  GrpcDestDriver *self = (GrpcDestDriver *) d;
```
MrAnno, line 295: It's out of scope of this PR, but the definition of `struct GrpcDestDriver_` and all other C structs should be in an extern C block.

MrAnno (commented):
Can we change the author of the commits that were used or influenced by Shifter's work? He has weeks of work in this.

modules/grpc/common/schema/grpc-schema.hpp
alltilla: I believe this is the exact virtual init that we have for `LogPipe`s. The `ProtoSchemaProvider` implementations can have an initial empty state produced by the ctor, then grammar rules can fill them with values, then we have point where we want to initialize them. I intentionally named it `init()`, so it matched the `LogPipe`-like convention.

If we wanted to do the clean RAII, we would need to store its config-driven values next to them in the driver and copy/share those, which does not seem clean to me.

modules/grpc/common/filterx/protobuf-field-converter.cpp
alltilla: There is not bytes API, string is used for that.

The API does not explicitly state this, though, but I cannot imagine anything else being used for that:
https://protobuf.dev/reference/cpp/api-docs/google.protobuf.message/

modules/grpc/common/schema/grpc-schema.hpp
MrAnno: I understand, thank you.

bshifter (approved):

modules/grpc/common/filterx/protobuf-field-converter.hpp
alltilla: Yeah we should. I am working on a patch that makes every C++ method throw exceptions instead of returning a bool, however it is turning out to be a huge change, so I would rather do that in a separate PR.

alltilla: Closing this comment, so we can merge the PR. I assume this is not a blocker thing for you as you have already given an approve.

PR #684, opened 2025-06-20
Filterx add format xml

sodomelle:
Example:

$MSG = format_xml({"a":{"b":"foo"}});

### DIFF
--- a/modules/xml/filterx-func-format-windows-eventlog-xml.c
+++ b/modules/xml/filterx-func-format-windows-eventlog-xml.c
@@ -22,6 +22,11 @@
  */
 #include "filterx-func-format-windows-eventlog-xml.h"
 
+static gboolean
+_append_inner_dict(FilterXObject *key, FilterXObject *dict, gpointer user_data)
+{
+  return TRUE;
+}
 
 FilterXExpr *
 filterx_function_format_windows_eventlog_xml_new(FilterXFunctionArgs *args, GError **error)
@@ -31,6 +36,7 @@ filterx_function_format_windows_eventlog_xml_new(FilterXFunctionArgs *args, GErr
 
   if (!self)
     return NULL;
+  self->append_inner_dict = _append_inner_dict;
 
   return s;
 }
--- a/modules/xml/filterx-func-format-windows-eventlog-xml.c
+++ b/modules/xml/filterx-func-format-windows-eventlog-xml.c
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2025 Axoflow
+ * Copyright (c) 2025 Tamas Kosztyu <tamas.kosztyu@axoflow.com>
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+#include "filterx-func-format-windows-eventlog-xml.h"
+
+
+FilterXExpr *
+filterx_function_format_windows_eventlog_xml_new(FilterXFunctionArgs *args, GError **error)
+{
+  FilterXExpr *s = filterx_function_format_xml_new(args, error);
+  FilterXFunctionFormatXML *self = (FilterXFunctionFormatXML *)s;
+
+  if (!self)
+    return NULL;
+
+  return s;
+}
+
+FILTERX_FUNCTION(format_windows_eventlog_xml, filterx_function_format_windows_eventlog_xml_new);
--- a/modules/xml/filterx-func-format-windows-eventlog-xml.c
+++ b/modules/xml/filterx-func-format-windows-eventlog-xml.c
@@ -21,10 +21,132 @@
  *
  */
 #include "filterx-func-format-windows-eventlog-xml.h"
+#include "filterx/filterx-eval.h"
+#include "filterx/object-extractor.h"
+#include "filterx/object-dict-interface.h"
+#include "filterx/object-list-interface.h"
+
+static gboolean
+_append_inner_data_dict_element(FilterXObject *key, FilterXObject *value, gpointer user_data)
+{
+  FilterXFunctionFormatXML *self = ((gpointer *) user_data)[0];
+  GString *buffer = ((gpointer *) user_data)[1];
+  const gchar *key_str;
+  gsize key_str_len;
+
+  if (!filterx_object_extract_string_ref(key, &key_str, &key_str_len))
+    {
+      filterx_eval_push_error_info_printf(XML_ERROR_STR, &self->super.super,
+                                          "Dict key must be a string, got %s",
+                                          filterx_object_get_type_name(key));
+      return FALSE;
+    }
+
+  const gchar *value_str;
+  gsize value_str_len;
+
+  if (!filterx_object_extract_string_ref(value, &value_str, &value_str_len))
+    {
+      filterx_eval_push_error_info_printf(XML_ERROR_STR, &self->super.super,
+                                          "Dict value must be a string, got %s",
+                                          filterx_object_get_type_name(value));
+      return FALSE;
+    }
+
+  g_string_append_printf(buffer, "<Data Name='%s'>%s</Data>", key_str, value_str);
+  return TRUE;
+}
+
+static gboolean
+_append_data_element(FilterXObject *key, FilterXObject *value, gpointer user_data)
+{
+  FilterXFunctionFormatXML *self = ((gpointer *) user_data)[0];
+  GString *buffer = ((gpointer *) user_data)[1];
+  const gchar *key_str;
+  gsize key_str_len;
+
+  if (!filterx_object_extract_string_ref(key, &key_str, &key_str_len))
+    {
+      filterx_eval_push_error_info_printf(XML_ERROR_STR, &self->super.super,
+                                          "Dict key must be a string, got %s",
+                                          filterx_object_get_type_name(key));
+      return FALSE;
+    }
+
+  const gchar *value_str;
+  gsize value_str_len;
+
+  if (!filterx_object_extract_string_ref(value, &value_str, &value_str_len))
+    {
+      filterx_eval_push_error_info_printf(XML_ERROR_STR, &self->super.super,
+                                          "Dict value must be a string, got %s",
+                                          filterx_object_get_type_name(value));
+      return FALSE;
+    }
+
+  append_leaf(key_str, value_str, value_str_len, buffer);
+  return TRUE;
+}
+
+static gboolean
+_append_data_dict(FilterXObject *key, FilterXObject *value, gpointer user_data)
+{
+  FilterXObject *value_unwrapped = filterx_ref_unwrap_ro(value);
+  if(filterx_object_is_type(value_unwrapped, &FILTERX_TYPE_NAME(dict)))
+    {
+      if(!filterx_dict_iter(value_unwrapped, _append_inner_data_dict_element, user_data))
+        return FALSE;
+
+      return TRUE;
+    }
+  else if(filterx_object_is_type(value_unwrapped, &FILTERX_TYPE_NAME(list)))
+    {
+      if (!append_list(key, value_unwrapped, user_data))
+        return FALSE;
+
+      return TRUE;
+    }
+  else if(!_append_data_element(key, value, user_data))
+    return FALSE;
+
+  return TRUE;
+}
 
 static gboolean
 _append_inner_dict(FilterXObject *key, FilterXObject *dict, gpointer user_data)
 {
+  GString *buffer = ((gpointer *) user_data)[1];
+  gboolean *is_only_attribute_present = ((gpointer *) user_data)[2];
+  const gchar *key_str;
+  gsize key_str_len;
+
+  if (!filterx_object_extract_string_ref(key, &key_str, &key_str_len))
+    {
+      FilterXFunctionFormatXML *self = ((gpointer *) user_data)[0];
+      filterx_eval_push_error_info_printf(XML_ERROR_STR, &self->super.super,
+                                          "Dict key must be a string, got %s",
+                                          filterx_object_get_type_name(key));
+      return FALSE;
+    }
+
+  append_inner_dict_start_tag(key_str, buffer);
+  gsize prev_buffer_len = buffer->len;
+
+  if (strncmp(key_str, "EventData", 9) == 0)
+    {
+      FilterXObject *dict_unwrapped = filterx_ref_unwrap_ro(dict);
+      if(!filterx_dict_iter(dict_unwrapped, _append_data_dict, user_data))
+        return FALSE;
+    }
+  else
+    {
+      *is_only_attribute_present = FALSE;
+      if (!filterx_dict_iter(dict, append_object, user_data))
+        return FALSE;
+    }
+
+  append_inner_dict_end_tag(key_str, user_data, prev_buffer_len);
+
   return TRUE;
 }
 
--- a/modules/xml/filterx-func-format-xml.c
+++ b/modules/xml/filterx-func-format-xml.c
@@ -21,11 +21,92 @@
  *
  */
 #include "filterx-func-format-xml.h"
+#include "filterx/object-extractor.h"
+#include "filterx/filterx-eval.h"
+#include "scratch-buffers.h"
+#include "filterx/object-dict-interface.h"
+
+#define FILTERX_FUNC_FORMAT_XML_USAGE "Usage: format_xml([dict])"
+const char *XML_ERROR_STR = "Failed to convert to xml";
+
+static void
+_append_leaf(const char *key_str, const char *value_str, gsize value_str_len, GString *buffer)
+{
+  if(value_str_len)
+    {
+      g_string_append_printf(buffer, "<%s>%s</%s>", key_str, value_str, key_str);
+    }
+  else
+    {
+      g_string_append_printf(buffer, "<%s/>", key_str);
+    }
+}
+
+static gboolean
+_append_entry(FilterXObject *key, FilterXObject *value, gpointer user_data)
+{
+  GString *buffer = ((gpointer *) user_data)[1];
+  const gchar *key_str;
+  gsize key_str_len;
+
+  g_assert(filterx_object_extract_string_ref(key, &key_str, &key_str_len));
+
+  GString *val_buf = scratch_buffers_alloc();
+  if (!filterx_object_str(value, val_buf))
+    {
+      FilterXFunctionFormatXML *self = ((gpointer *) user_data)[0];
+      filterx_eval_push_error_info_printf(XML_ERROR_STR, &self->super.super,
+                                          "Couldn't convert value to string, type: %s",
+                                          filterx_object_get_type_name(value));
+      return FALSE;
+    }
+
+  append_leaf(key_str, val_buf->str, val_buf->len, buffer);
+  return TRUE;
+}
+
+static gboolean
+_append_object(FilterXObject *key, FilterXObject *value, gpointer user_data)
+{
+
+  if(!_append_entry(key, value, user_data))
+    return FALSE;
+
+  return TRUE;
+}
 
 static FilterXObject *
 _eval(FilterXExpr *s)
 {
-  return NULL;
+  FilterXFunctionFormatXML *self = (FilterXFunctionFormatXML *) s;
+
+  FilterXObject *input_dict = filterx_expr_eval_typed(self->input);
+  if (!input_dict)
+    return NULL;
+
+  ScratchBuffersMarker marker;
+  GString *formatted = scratch_buffers_alloc_and_mark(&marker);
+  FilterXObject *input_dict_unwrapped = filterx_ref_unwrap_ro(input_dict);
+  if (!filterx_object_is_type(input_dict_unwrapped, &FILTERX_TYPE_NAME(dict)))
+    {
+      scratch_buffers_reclaim_marked(marker);
+      filterx_object_unref(input_dict);
+      filterx_eval_push_error("input must be a dict. " FILTERX_FUNC_FORMAT_XML_USAGE, s, input_dict);
+      return NULL;
+    }
+
+  gpointer user_data[] = { self, formatted };
+
+  if (!filterx_dict_iter(input_dict_unwrapped, append_object, user_data))
+    {
+      scratch_buffers_reclaim_marked(marker);
+      filterx_object_unref(input_dict);
+      return NULL;
+    }
+
+  scratch_buffers_reclaim_marked(marker);
+  filterx_object_unref(input_dict);
+  return filterx_string_new(formatted->str, formatted->len);
 }
 
 static FilterXExpr *
--- a/modules/xml/filterx-func-format-xml.c
+++ b/modules/xml/filterx-func-format-xml.c
@@ -180,6 +180,7 @@ _append_entry(FilterXObject *key, FilterXObject *value, gpointer user_data)
 static gboolean
 _append_object(FilterXObject *key, FilterXObject *value, gpointer user_data)
 {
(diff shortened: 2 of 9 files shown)


tests/light/functional_tests/filterx/test_filterx.py
```diff
@@ -2604,3 +2604,140 @@ def test_arithmetic_operators_type(config, syslog_ng):
...
+
+def test_format_xml_invalid_arg_number(config, syslog_ng):
+    _ = create_config(
+        config, """
+        $MSG = format_xml("test", 2);
+""",
+    )
+    with pytest.raises(Exception):
+        syslog_ng.start(config)
+
```
mitzkia, line 2614: Please add check why syslog-ng could not start in this case.
like:
```
    with pytest.raises(Exception) as exec_info:
        syslog_ng.start(config)
    assert "syslog-ng is not running" in str(exec_info.value)
```

modules/xml/filterx-func-format-xml.c
```diff
@@ -21,11 +21,80 @@
...
-  return NULL;
+  FilterXFunctionFormatXML *self = (FilterXFunctionFormatXML *) s;
+
+  FilterXObject *input_dict = filterx_expr_eval_typed(self->input);
+  if (!input_dict)
+    return NULL;
+
+  GString *formatted = scratch_buffers_alloc();
+  FilterXObject *input_dict_unwrapped = filterx_ref_unwrap_ro(input_dict);
+  filterx_object_unref(input_dict);
```
alltilla: `input_dict_unwrapped` does not add a new ref to `input_dict`. We should keep `input_dict` alive while we are still working on `input_dict_unwrapped`.

We should only unref it when we return from this function.

This probably works as we are not the only ref holders here, but it is logically not sound.

modules/xml/filterx-func-format-xml.h
```diff
@@ -30,6 +30,7 @@ FILTERX_FUNCTION_DECLARE(format_xml);
 typedef struct FilterXFunctionFormatXML_
 {
   FilterXFunction super;
+  gboolean has_just_attribute;
```
alltilla: A single `FilterXFunctionFormatXML` instance is shared between processing (source) threads. We have one expression, and multiple threads call its `eval()` in parallel.

Because of this, we cannot save evaluation related info here. In other words, this class must be stateless after its initialization.

In C++, we would have something like this
```cpp
class FilterXExpr {
...
  virtual FilterXObject *eval() const = 0;
...
};
```
Emphasis on `const`, which means this function cannot modify the instance.
Unlike C++, in C `const`-ness does not distribute to the struct's inner structs, etc, so we don't bother with marking it as `const`. This adds a development complexity of thinking about `const`-ness, but we have to live with it.

Please move this field to a separate eval context struct (or similar), which gets initialized locally in the beginning of `eval()`, then you can pass it to the inner processing functions.

modules/xml/filterx-func-format-xml.c
```diff
@@ -28,6 +28,26 @@
...
 
+static void
+_append_attribute(const char *key_str, const char *value_str, GString *buffer)
+{
+  if (buffer->str[buffer->len - 1] == '>')
+    g_string_overwrite(buffer, buffer->len - 1, " ");
+  else
+    g_string_append_c(buffer, ' ');
+
+  g_string_append_printf(buffer, "%s=\'%s\'>", &key_str[1], value_str);
```
alltilla: Don't we need some escaping here?

modules/xml/filterx-func-format-xml.c
```diff
@@ -25,9 +25,28 @@
...
+
+static gboolean
+_append_list(FilterXObject *key, FilterXObject *list, gpointer user_data)
+{
+  guint64 len;
+  filterx_object_len(list, &len);
+
+  for (gsize i = 0; i < len; i++)
+    {
+      FilterXObject *elem = filterx_list_get_subscript(list, i);
```
alltilla, line 97: I think we need to release the ref for `elem` when we are done with it.

modules/xml/filterx-func-format-xml.c
```diff
@@ -31,6 +31,56 @@
...
+  if(!filterx_object_extract_string_ref(key, &key_str, &key_str_len))
+    return FALSE;
+
+  _append_inner_dict_start_tag(key_str, buffer);
+
+  self->has_just_attribute = FALSE;
+  if(!filterx_dict_iter(dict, _append_to_buffer, user_data))
+    return FALSE;
+
+  _append_inner_dict_end_tag(key_str, user_data);
```
alltilla: It is possible that after xml parsing we got something like:
```
"key":{"@attr": "val"}
```
Then the user removes the `attr` as they don't need it, then we have:
```
"key":{}
```
I think we should format this in this case:
```
<key/>
```

---

Also, how does this work if we have a similar case to what I mentioned, but `"#text"` will still remain.

mitzkia:
We have discussed a few things IRL:
 - negative (parametrized) testcases for different data types (int, date, ...) in values
 - negative (parametrized) testcases for special characters other than "#" and "@" in key

tests/copyright/policy
```diff
@@ -293,6 +293,8 @@ modules/correlation/group-lines.h
 modules/xml/windows-eventlog-xml-parser\.h
 modules/xml/filterx-parse-xml\.[ch]$
 modules/xml/filterx-parse-windows-eventlog-xml\.[ch]$
+modules/xml/filterx-func-format-xml\.[ch]
+modules/xml/filterx-func-format-windows-eventlog-xml\.[ch]
```
alltilla: [note]

FYI, we don't need to set these anymore manually, everything is GPLv3+ by default IIRC.

alltilla (commented):
Very clean commit history, it was a breeze to review!

modules/xml/filterx-func-format-xml.c
sodomelle: When `#text` remains, I think it's still working just fine, as the text will be put as the value of the element.
The first one currently produces `<key></key>`.

tests/light/functional_tests/filterx/test_filterx.py
sodomelle: Done.

modules/xml/filterx-func-format-xml.c
sodomelle: Done.

modules/xml/filterx-func-format-xml.h
sodomelle: Done.

modules/xml/filterx-func-format-xml.c
sodomelle: I don't really know what you mean by that. Either way I removed the `\` characters before `'` :D

modules/xml/filterx-func-format-xml.c
sodomelle: Done.

tests/copyright/policy
sodomelle: Removed.

modules/xml/filterx-func-format-xml.c
alltilla: Can we make a `len == 0` check and do the `<key/>` insertion?

alltilla (commented):
Thanks for the updates, after doing the TODOs from the description and fixing the empty dict case, I am ready to approve.

modules/xml/filterx-func-format-xml.c
sodomelle: Done some length checking ;)

modules/xml/filterx-func-format-xml.c
```diff
@@ -21,11 +21,89 @@
...
+}
+
+static gboolean
+_append_entry(FilterXObject *key, FilterXObject *value, gpointer user_data)
+{
+  GString *buffer = ((gpointer *) user_data)[1];
+  const gchar *key_str;
+  gsize key_str_len;
+
+  (void)filterx_object_extract_string_ref(key, &key_str, &key_str_len);
```
alltilla: A failed call to this function will leave `key_str` and `key_str_len` uninitialized. We are using those values later on, and it can cascade to very strange undefined behaviors.

I know that dicts always has string keys, but the API does not enforce this right now.
I would recommend to either `g_assert()` the success of this (builds on the current non-enforceable dict API), or check for the result and handle a possible error, even if it never happens (follows the general API usage).

modules/xml/filterx-func-format-xml.c
```diff
@@ -25,10 +25,33 @@
...
 #define FILTERX_FUNC_FORMAT_XML_USAGE "Usage: format_xml([dict])"
 const char *XML_ERROR_STR = "Failed to convert to xml";
 
+static gboolean _append_to_buffer(FilterXObject *key, FilterXObject *value, gpointer user_data);
+
+static gboolean
+_append_list(FilterXObject *key, FilterXObject *list, gpointer user_data)
+{
+  guint64 len;
+  filterx_object_len(list, &len);
```
alltilla: A failing `len()` call would leave `len` uninitialized.

If we check the code at its current state, we can see that only lists can appear here, so this `len()` call will always succeed. However, this is not enforced with the current function signatures and the compiler, so it can break in the future if we mess something up. Can you `g_assert()` the result of this call, so we catch possible coding errors sooner in the future?

modules/xml/filterx-func-format-xml.c
```diff
@@ -32,6 +32,61 @@ const char *XML_ERROR_STR = "Failed to convert to xml";
...
+
+static gboolean
+_append_inner_dict(FilterXObject *key, FilterXObject *dict, gpointer user_data)
+{
+  GString *buffer = ((gpointer *) user_data)[1];
+  gboolean *is_only_attribute_present = ((gpointer *) user_data)[2];
+  const gchar *key_str;
+  gsize key_str_len;
+
+  (void)filterx_object_extract_string_ref(key, &key_str, &key_str_len);
```
alltilla: Same as a previous comment of mine.

modules/xml/filterx-func-format-xml.c
```diff
@@ -176,12 +177,65 @@ _append_entry(FilterXObject *key, FilterXObject *value, gpointer user_data)
...
 
+static gboolean
+_is_valid_xml_tag_name(FilterXObject *tag, gpointer user_data)
+{
+  const gchar *tag_str;
+  gsize tag_str_len;
+
+  if (!filterx_object_extract_string_ref(tag, &tag_str, &tag_str_len))
+    {
+      FilterXFunctionFormatXML *self = ((gpointer *) user_data)[0];
```
alltilla: [optional]

You can cast this once in the beginning of the function to avoid code duplication.

modules/xml/filterx-func-format-xml.c
```diff
@@ -176,12 +177,65 @@ _append_entry(FilterXObject *key, FilterXObject *value, gpointer user_data)
...
+  if (tag_str_len > 2 &&
+      (tolower(tag_str[0]) == 'x' &&
+       tolower(tag_str[1]) == 'm' &&
+       tolower(tag_str[2]) == 'l'))
+    {
+      FilterXFunctionFormatXML *self = ((gpointer *) user_data)[0];
+      filterx_eval_push_error_info(XML_ERROR_STR, &self->super.super,
+                                   "Dict key can't start with with \"xml\" string", FALSE);
+      return FALSE;
+    }
```
alltilla: If I understand the specs right, in an element's name, the prefix `xml` is valid, but it is reserved for standardized future use. So there can be a valid XML that has an element name starting with `xml`, as long as it follows the standardized reserved values (which currently there is not any, I believe, but it can be in the future).

modules/xml/filterx-func-format-xml.c
```diff
@@ -176,12 +177,65 @@ _append_entry(FilterXObject *key, FilterXObject *value, gpointer user_data)
...
+  // '@' and '#' characters are reserved for attributes and text, so we consider them valid
+  if (!(isalpha(tag_str[0]) || tag_str[0] == '_' || tag_str[0] == '@' || tag_str[0] == '#'))
+    {
+      FilterXFunctionFormatXML *self = ((gpointer *) user_data)[0];
+      filterx_eval_push_error_info(XML_ERROR_STR, &self->super.super,
+                                   "Dict key must begin with a letter or '_' character", FALSE);
+      return FALSE;
+    }
+
+  for(gint i = 1; i < tag_str_len -1; ++i)
```
alltilla, line 210: We can arrive here with `tag_str_len` being 0 and `tag_str_len - 1` underflowing, causing this to be a really long loop.

I think we should mark empty strings as invalid, too, in the beginning.

modules/xml/filterx-func-format-xml.c
alltilla: Escaping is a strategy of handling special characters in strings, for example if you want to use a single quote `'` character in a python single-quote string, you need to escape it with a backslash:

```py
my_string = 'Attila\'s string'
```

I was afraid of injections by special XML characters (`'`, `>`, ...). Now that I read the XML specs, I know that another special character handling strategy is used in XMLs: invalidation. You cannot have special characters in any way in your XML at all, end of story. And you have implemented this in the formatter, so this is now fixed.

modules/xml/filterx-func-format-xml.c
alltilla: Thanks!

modules/xml/filterx-func-format-xml.c
sodomelle: For every `key`, `_is_valid_xml_tag_name` is called, which does this check for extracting. I didn't want to add error handling again. But you're right, I'll add a g_assert here, just in case.

sodomelle: Nice catch! Added assertion.

modules/xml/filterx-func-format-xml.c
sodomelle: Nice catch! Added assertion.

modules/xml/filterx-func-format-xml.c
sodomelle: Added assertion.

modules/xml/filterx-func-format-xml.c
sodomelle: Done.

modules/xml/filterx-func-format-xml.c
sodomelle: Removed checking for "xml" string.

modules/xml/filterx-func-format-xml.c
sodomelle: Marked empty tag names as invalid, so if  `tag_str_len` is `0` we return `FALSE`. (They are invalid)

alltilla (approved):
Thanks for the fixes, great work!

mitzkia (dismissed):
approved for the Light tests

alltilla (approved):

alltilla:
@sodomelle I have put one followup commit to this branch for your convenience:
https://github.com/axoflow/axosyslog/pull/684/commits/76f952a3ef6dc380c144cec9775a6f4605e1d5a7

tests/light/functional_tests/filterx/test_filterx.py
```diff
@@ -2600,3 +2600,164 @@ def test_arithmetic_operators_type(config, syslog_ng):
...
+def test_format_windows_eventlog_list(config, syslog_ng):
+    (file_true, file_false) = create_config(
+        config, windows_eventlog_dict.replace("{eventdata}", """{"Data":["foo","bar"]}"""),
+    )
+    syslog_ng.start(config)
+
+    assert file_true.get_stats()["processed"] == 1
+    assert "processed" not in file_false.get_stats()
+    assert file_true.read_log() == r"""<Event xmlns='http://schemas.microsoft.com/win/2004/08/events/event'><System><Provider Name='EventCreate'/><EventID Qualifiers='0'>999</EventID><Version>0</Version><Level>2</Level><Task>0</Task><Opcode>0</Opcode><Keywords>0x80000000000000</Keywords><TimeCreated SystemTime='2024-01-12T09:30:12.1566754Z'/><EventRecordID>934</EventRecordID><Correlation/><Execution ProcessID='0' ThreadID='0'/><Channel>Application</Channel><Computer>DESKTOP-2MBFIV7</Computer><Security UserID='S-1-5-21-3714454296-2738353472-899133108-1001'/></System><RenderingInfo Culture='en-US'><Message>foobar</Message><Level>Error</Level><Task/><Opcode>Info</Opcode><Channel/><Provider/><Keywords><Keyword>Classic</Keyword></Keywords></RenderingInfo><EventData><Data>foo</Data><Data>bar</Data></EventData></Event>"""
```
MrAnno, line 2827: Optional: It may be worth writing a test where an XML and a Windows-specific XML are passed to `parse_xml()` and then back to `format_xml()` to assert on nothing changing.

modules/xml/filterx-func-format-xml.c
```diff
@@ -21,11 +21,88 @@
...
 _eval(FilterXExpr *s)
 {
-  return NULL;
+  FilterXFunctionFormatXML *self = (FilterXFunctionFormatXML *) s;
+
+  FilterXObject *input_dict = filterx_expr_eval_typed(self->input);
+  if (!input_dict)
+    return NULL;
+
+  GString *formatted = scratch_buffers_alloc();
```
MrAnno: It's worth using `alloc_and_mark()` as those XMLs can be huge.

modules/xml/filterx-func-format-xml.c
```diff
@@ -61,6 +82,19 @@ _append_entry(FilterXObject *key, FilterXObject *value, gpointer user_data)
...
+  if (key_str_len && (key_str[0] == '@'))
+    {
+      *is_only_attribute_present = TRUE;
+      _append_attribute(key_str, val_buf->str, buffer);
+      return TRUE;
+    }
+  if (key_str_len && (key_str[0] == '#'))
+    {
+      *is_only_attribute_present = FALSE;
+      _append_text(val_buf->str, buffer);
```
MrAnno, line 171: This seems invalid, especially the `>` character between the 2 subvalues:
```
    $json = {"#toplevel": { "#sub": "blablabla", "#sub2": "mamama" }};
    <#toplevel>blablabla>mamama</#toplevel>
```

`#` is currently allowed in tag names on top level. Inside a tag, anything written after `#` will be lost.

I think we should match for `#text` exactly. The `>` problem will disappear because dicts can't have 2 identical keys.
Anything else starting with `#` should be treated somehow else. (Making a tag out of them no matter where we are seems acceptable for now, even if it's not valid XML. Alternatively, we can make it an error, but I wouldn't bother implementing it in this PR.)

modules/xml/filterx-func-format-xml.c
```diff
@@ -102,6 +125,15 @@ _append_entry(FilterXObject *key, FilterXObject *value, gpointer user_data)
 static gboolean
 _append_to_buffer(FilterXObject *key, FilterXObject *value, gpointer user_data)
 {
+  FilterXObject *value_unwrapped = filterx_ref_unwrap_ro(value);
+
+  if (filterx_object_is_type(value_unwrapped, &FILTERX_TYPE_NAME(list)))
+    {
+      if (!_append_list(key, value_unwrapped, user_data))
```
MrAnno: Top-level list wouldn't make sense, right?

modules/xml/filterx-func-format-windows-eventlog-xml.c
```diff
@@ -31,8 +36,9 @@ filterx_function_format_windows_eventlog_xml_new(FilterXFunctionArgs *args, GErr
 
   if (!self)
     return NULL;
+  self->append_inner_dict = _append_inner_dict;
 
   return s;
 }
 
-FILTERX_FUNCTION(format_windows_eventlog_xml, filterx_function_format_windows_eventlog_xml_new);
+FILTERX_FUNCTION(format_windows_eventlog_xml, filterx_function_format_windows_eventlog_xml_new);
```
MrAnno: This can help you avoid those line ending mistakes in the future:
```
"files.insertFinalNewline": true,
"files.trimFinalNewlines": true,
```

modules/xml/filterx-func-format-xml.c
```diff
@@ -0,0 +1,369 @@
...
+      filterx_eval_push_error_static_info(XML_ERROR_STR, &self->super.super,
+                                          "XML tag name can't be empty");
+      return FALSE;
+    }
+
+  // '@' and '#' characters are reserved for attributes and text, so we consider them valid
+  if (!(isalpha(tag_str[0]) || tag_str[0] == '_' || tag_str[0] == '@' || tag_str[0] == '#'))
+    {
+      filterx_eval_push_error_static_info(XML_ERROR_STR, &self->super.super,
+                                          "Dict key must begin with a letter or '_' character");
```
MrAnno, line 206: Shouldn't we talk about `@` and `#text` too?

modules/xml/filterx-func-format-xml.h
```diff
@@ -0,0 +1,47 @@
...
+  FilterXFunction super;
+  FilterXExpr *input;
+  gboolean (*append_inner_dict)(FilterXObject *key, FilterXObject *dict, gpointer user_data);
+} FilterXFunctionFormatXML;
+
+FilterXExpr *filterx_function_format_xml_new(FilterXFunctionArgs *args, GError **error);
+
+extern const char *XML_ERROR_STR;
+
+gboolean append_to_buffer(FilterXObject *key, FilterXObject *value, gpointer user_data);
```
MrAnno: This looks more like `append_element` or `append_object`.

MrAnno (approved):
Excellent commit history :)

MrAnno:
Sorry, I didn't want to approve it yet as I left some notes that affect functionality.

tests/light/functional_tests/filterx/test_filterx.py
sodomelle: Done. Event Data part is handled differently.

modules/xml/filterx-func-format-xml.c
sodomelle: Done.

modules/xml/filterx-func-format-xml.c
sodomelle: Replaced `'#'` check with check for `#text` string.

modules/xml/filterx-func-format-xml.c
sodomelle: I don't think so.

modules/xml/filterx-func-format-windows-eventlog-xml.c
sodomelle: Thank you. Fixed the accidental removal of the new line at the end of file.

modules/xml/filterx-func-format-xml.c
sodomelle: Changed it to `"#text"` check.

modules/xml/filterx-func-format-xml.h
sodomelle: You're right, renamed function.

modules/xml/filterx-func-format-xml.c
```diff
@@ -0,0 +1,372 @@
...
+  gboolean is_only_attribute_present = FALSE;
+  gpointer user_data[] = { self, formatted, &is_only_attribute_present };
+
+  if (!filterx_dict_iter(input_dict_unwrapped, append_object, user_data))
+    {
+      filterx_object_unref(input_dict);
+      return NULL;
+    }
+
+  scratch_buffers_reclaim_marked(marker);
```
MrAnno, line 283: Opt: This should go to the early return locations too (or a goto error handling), but it's not a big problem.

modules/xml/filterx-func-format-xml.c
```diff
@@ -0,0 +1,372 @@
...
+      return FALSE;
+    }
+
+  if (key_str_len && (key_str[0] == '@'))
+    {
+      *is_only_attribute_present = TRUE;
+      _append_attribute(key_str, val_buf->str, buffer);
+      return TRUE;
+    }
+  if (key_str_len && (strncmp(key_str, "#text", 5) == 0))
```
MrAnno: Just a note: this is a substring match because of the hard-coded 5.

MrAnno (dismissed):

modules/xml/filterx-func-format-xml.c
sodomelle: fixed

modules/xml/filterx-func-format-xml.c
sodomelle: fixed

MrAnno: `strncmp()` will always be substr, unless you add a a_len == b_len precheck.
`#t` will now match for example.

But I think it's good enough for now :)

modules/xml/filterx-func-format-xml.c
MrAnno: Thanks

MrAnno (dismissed):

modules/xml/filterx-func-format-xml.c
sodomelle: changed it to `g_strcmp0` :D

MrAnno (approved):

PR #688, opened 2025-06-23
light: stop message timeout increase

alltilla:


tests/light/src/axosyslog_light/syslog_ng/syslog_ng.py
```diff
@@ -96,7 +96,7 @@ def stop(self, unexpected_messages: typing.List[str] = None) -> None:
                 self.__error_handling("syslog-ng did not stop")
             if self.start_params.stderr and self.start_params.debug and self.start_params.verbose:
                 if not self._console_log_reader.wait_for_stop_message():
-                    self.__error_handling("Stop message not arrived")
+                    logger.warning("Stop message has not been found, this might be because of a very long console log")
```
mitzkia, line 99: This is ok, due to before this we check the return code of syslog-ng if something went wrong during the stopping phase. So we can still catch the crash after this.

mitzkia (approved):
approved

PR #691, opened 2025-06-24
filterx-scope: avoid log_msg_unset_value() side-effects

MrAnno:
When an unset key enters FilterX, we shouldn't unset it again during sync as it can cause side-effects, for example, unsetting `PID` or `PROGRAM` may invalidate the legacy MSGHDR.


alltilla (approved):
wow that is strange!

PR #697, opened 2025-06-30
grpc/filterx: add missing DEPENDENCIES to Makefile

alltilla:


alltilla:
Trellix MPS logs in splunk after reduction (notice that `rt` and `dvchost` got removed by reduction):

image

PR #699, opened 2025-07-02
`filterx`: `parse_cef()` and `parse_leef()` parsing fixes

alltilla:
There are some parsing errors with the code on main.

CEF and LEEF does some strange escaping rules, which we did not correctly support.

This new version intends to make the parsing code compatible with the spec (see Character Encoding):
https://www.microfocus.com/documentation/arcsight/arcsight-smartconnectors-8.3/cef-implementation-standard/Content/CEF/Chapter%201%20What%20is%20CEF.htm


lib/scanner/kv-scanner/kv-scanner.c
```diff
@@ -303,6 +331,9 @@ _decode_value(KVScanner *self)
     _decode_value_until_next_key(self, input);
   else
     _decode_value_until_next_pair_separator(self, input);
+
+  if (self->flags & KV_SCANNER_ESCAPE_VALUE_SEPARATOR)
```
MrAnno: Is it right that `KV_SCANNER_ESCAPE_VALUE_SEPARATOR` does not really make sense without `KV_SCANNER_VALUE_EXTENDS_TO_NEXT_KEY`?

If so, wouldn't make more sense to just add support for additional escape sequences  in
`_decode_backslash_escape` (`StrReprDecodeOptions`) and avoid any additional kv-scanner parsing logic?

lib/scanner/kv-scanner/kv-scanner.c
```diff
@@ -294,6 +307,21 @@ _decode_value_until_next_pair_separator(KVScanner *self, const gchar *input)
     }
 }
 
+static void
+_unescape_value_separator(KVScanner *self)
+{
+  for (gsize i = 0; i < self->value->len; i++)
```
sodomelle: Nitpick: I think the condition in line 316 could be used instead:
`for (gsize i = 0; i < self->value->len; i++)`
to
`for (gsize i = 0; i + 1 < self->value->len; i++)`

modules/cef/event-format-parser.c
```diff
@@ -51,11 +51,41 @@ _get_current_field(EventParserContext *ctx)
...
+          g_string_append_c(unescaped, '\\');
+          i++;
+        }
+      else
+        {
+          g_string_append_c(unescaped, value[i]);
+        }
+    }
+
+  gssize unascaped_len = unescaped->len;
```
sodomelle, line 80: Nitpick: I don't think it's necessary to create a new variable here.

modules/cef/filterx-func-format-leef.c
```diff
@@ -54,49 +53,53 @@ filterx_function_format_leef_format_delimiter(EventFormatterContext *ctx, GStrin
...
-      if (delimiter_len)
-        {
-          append_unsafe_utf8_as_escaped(formatted, delimiter_str, delimiter_len, 0, "\\x%02x", "\\x%02x");
-          ctx->config.extensions.pair_separator[0] = delimiter_str[0];
-          ctx->config.extensions.pair_separator[1] = '\0';
-        }
+      gchar type_name_buf[FILTERX_OBJECT_TYPE_NAME_BUF_SIZE];
+      gchar *info = g_strdup_printf("Header value for must be a string, got: %s, header: delimiter",
+                                    filterx_object_format_type_name(delimiter_obj, type_name_buf));
+      filterx_eval_push_error_info("Failed to evaluate event formatter function",
```
sodomelle: Don't forget to update eval_push_error_info related calls.

sodomelle (commented):
Looks good, added some nitpics, and a caution for rebasing.

modules/cef/event-format-parser.c
alltilla: `g_string_free()` frees the `GString` object that holds this `len` information, but I need the `len` after it got freed, so it is necessary to store it beforehand.

lib/scanner/kv-scanner/kv-scanner.c
alltilla: Very nice observation, fixed!

lib/scanner/kv-scanner/kv-scanner.c
alltilla: I'll write some unit tests for each separately, and will see if those make sens of their own. If not, I am open to set them in a different way.

modules/cef/filterx-func-format-leef.c
alltilla: Done!

modules/cef/event-format-parser.c
sodomelle: Oh yes, my bad.

sodomelle (dismissed):
My findigs were fixed. Looks good to me.

lib/scanner/kv-scanner/kv-scanner.c
alltilla: We don't need both. I think the important difference is the handling of stray words. As I found this out, I chose a different implementation. I could not fit it into `decode.c`, but I think with the new implementation it is a lot more reasonable addition to `KVScanner`.

alltilla:
Added the support for empty, but not missing header values. Re-running the tests.

MrAnno (approved):

lib/scanner/kv-scanner/kv-scanner.c
MrAnno: Thanks.

PR #701, opened 2025-07-03
filterx-error: do not store info on stack

alltilla:
musl provides a relateviely small (128K) stack size: https://wiki.musl-libc.org/functional-differences-from-glibc.html#Thread-stack-size

The 8 errors * 1024 bytes of memory occupied for each filterx scope reaches 128K after 16 scopes chained.

We should consider raising the limit explicitly.

### DIFF
--- a/lib/filterx/filterx-error.c
+++ b/lib/filterx/filterx-error.c
@@ -101,7 +101,7 @@ filterx_falsy_error_set_values(FilterXError *error, const gchar *message, Filter
 }
 
 void
-filterx_error_set_info(FilterXError *error, const gchar *info)
+filterx_error_set_static_info(FilterXError *error, const gchar *info)
 {
   error->info = (gchar *) info;
   error->free_info = FALSE;
--- a/lib/filterx/expr-arithmetic-operators.c
+++ b/lib/filterx/expr-arithmetic-operators.c
@@ -44,8 +44,8 @@ _eval_arithmetic_operators_common(FilterXArithmeticOperator *self, GenericNumber
                               : filterx_expr_eval_typed(self->super.lhs);
   if (!lhs_object)
     {
-      filterx_eval_push_error_info("Failed to evaluate arithmetic operator", &self->super.super,
-                                   "Failed to evaluate left hand side");
+      filterx_eval_push_error_static_info("Failed to evaluate arithmetic operator", &self->super.super,
+                                          "Failed to evaluate left hand side");
       return FALSE;
     }
 
@@ -64,8 +64,8 @@ _eval_arithmetic_operators_common(FilterXArithmeticOperator *self, GenericNumber
 
   if (!rhs_object)
     {
-      filterx_eval_push_error_info("Failed to evaluate arithmetic operator", &self->super.super,
-                                   "Failed to evaluate right hand side");
+      filterx_eval_push_error_static_info("Failed to evaluate arithmetic operator", &self->super.super,
+                                          "Failed to evaluate right hand side");
       return FALSE;
     }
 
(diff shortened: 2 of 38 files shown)


lib/filterx/filterx-error.c
```diff
@@ -101,7 +101,9 @@ filterx_falsy_error_set_values(FilterXError *error, const gchar *message, Filter
 void
 filterx_error_set_info(FilterXError *error, const gchar *info)
 {
-  g_strlcpy(error->info, info, FILTERX_ERROR_INFO_LEN);
+  GString *buf = scratch_buffers_alloc();
```
MrAnno: This means we can't call these functions in the middle of a `mark/reclaim` block and we shouldn't reclaim all buffers in FilterX code ever (we'll have to remember), only when all scopes are released because we store these errors in an optional feature `failure info`.

We have a few of those:
```
expr-compound.c:79:      scratch_buffers_reclaim_marked(mark);
expr-condition.c:119:      scratch_buffers_reclaim_marked(mark);
filterx-metrics-labels.c:154:  scratch_buffers_mark(&marker);
filterx-metrics-labels.c:158:  scratch_buffers_reclaim_marked(marker);
filterx-metrics.c:143:  scratch_buffers_mark(&marker);
filterx-metrics.c:149:      scratch_buffers_reclaim_marked(marker);
filterx-metrics.c:156:  scratch_buffers_reclaim_marked(marker);
filterx-metrics.c:193:  scratch_buffers_mark(&marker);
filterx-metrics.c:203:  scratch_buffers_reclaim_marked(marker);
func-failure-info.c:112:  scratch_buffers_mark(&mark);
func-failure-info.c:123:  scratch_buffers_reclaim_marked(mark);
func-str.c:371:  scratch_buffers_mark(&marker);
func-str.c:384:  scratch_buffers_reclaim_marked(marker);
func-vars.c:92:  scratch_buffers_reclaim_marked(marker);
json-repr.c:186:  scratch_buffers_reclaim_marked(marker);
object-metrics-labels.c:121:  scratch_buffers_mark(&marker);
object-metrics-labels.c:141:  scratch_buffers_reclaim_marked(marker);
```

alltilla: Yeah, you are right. We might have to dup these :/

MrAnno: Why don't we just lazy- or pre-allocate a 1024 long buffer on the heap and reuse it until the eval context or FilterXFailureInfo ends?

Or just a GString with a good default capacity.

MrAnno: We'll need some kind of cache/storage to keep error-absorbing features fast.

lib/filterx/filterx-error.c
```diff
@@ -101,7 +103,8 @@ filterx_falsy_error_set_values(FilterXError *error, const gchar *message, Filter
 void
 filterx_error_set_info(FilterXError *error, const gchar *info)
 {
-  g_strlcpy(error->info, info, FILTERX_ERROR_INFO_LEN);
+  error->info = (gchar *) info;
+  error->free_info = FALSE;
```
MrAnno, line 107: Originally `filterx_eval_push_error_info()` and `filterx_error_set_info()` had a bool parameter free_info.

Now that we don't have those, can we rename functions that operate on `free_info = FALSE` objects to make it clear that the info is not owned and has to live long enough? (probably we're talking about static strings)

MrAnno: set_static_info?
filterx_eval_push_error_static_info?

(or anything similar)

We can rename them back if we figured out the lifetimes and the storage.

alltilla: Sure, done!

MrAnno:
@sodomelle This PR became a partial revert + renaming after the initial discussions, so I merged it without your review to fix `main`.

PR #705, opened 2025-07-06
filterx: `parse_leef()`: support missing delimiter in v2

alltilla:
#699 removed the support for missing delimiter header field in LEEF 2.0, as it is not marked to be optional in the spec, and the parsing fix was a lot easier to implement without it.

However a lot of other 3rd party docs do mark it optional, and certain apps omit it completely (which we just found out), so it is better to support it.


alltilla:
@MrAnno It would be nice to have this in the release as well, though it is not strictly necessary. We did't have any examples of LEEF 2.0 no delimiter header field until now, so everything that have worked before is working right now even with the current state of `main`.

modules/cef/filterx-func-parse-leef.c
```diff
@@ -87,6 +87,21 @@ _is_delmiter_empty(gchar delimiter)
   return delimiter == 0;
 }
 
+static gboolean
+_fallback_to_parse_extensions(EventParserContext *ctx, const gchar *input, gint input_len, FilterXObject **result,
+                              GError **error, gpointer user_data)
+{
+  if (!csv_scanner_append_rest(ctx->csv_scanner))
```
MrAnno, line 94: Shouldn't `error` be set?

alltilla: You are right, added.

Also while I was fixing this I thought about another edge case, which was not handled correctly.
Fixed that also.

alltilla: if separate extensions was set, the key-values were erroneously added under the delimiter key if delimiter was missing.

alltilla: The fact that the set subscript call is in the common code introduces a lot of complicated code to overrule it in some cases. Maybe we should just move the set subscript to the parse callback and we would have a simpler code.

If you think it is worth to do I will do it on Monday, but if you think the code right now is good enough we can merge it. It is just an idea.

alltilla: Pushed a commit, we can drop it if you don't like it.

PR #706, opened 2025-07-06
filterx: make `parse_cef()` and `parse_leef()` non-generators

alltilla:
Also added some patches to improve code readability and make it closer to conventions.

Depends on #705

### DIFF
--- a/modules/cef/event-format-parser.c
+++ b/modules/cef/event-format-parser.c
@@ -105,7 +105,8 @@ parse_default(EventParserContext *ctx, const gchar *value, gint value_len, Filte
 }
 
 gboolean
-parse_version(EventParserContext *ctx, const gchar *value, gint value_len, FilterXObject *parsed_dict)
+event_format_parser_parse_version(EventParserContext *ctx, const gchar *value, gint value_len,
+                                  FilterXObject *parsed_dict)
 {
   const gchar *log_signature = ctx->parser->config.signature;
   gchar *colon_pos = memchr(value, ':', value_len);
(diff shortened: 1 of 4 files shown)


modules/cef/event-format-parser.c
```diff
@@ -426,13 +433,12 @@ gboolean
...
-  self->super.super.super.optimize = _optimize;
-  self->super.super.super.init = _init;
-  self->super.super.super.deinit = _deinit;
-  self->super.super.super.free_fn = _free;
+  filterx_function_init_instance(&self->super, fn_name);
+  self->super.super.eval = _eval;
+  self->super.super.optimize = _optimize;
+  self->super.super.init = _init;
+  self->super.super.deinit = _deinit;
+  self->super.super.free_fn = _free;
```
MrAnno, line 437: There is a macro at the bottom of each file called `FILTERX_GENERATOR_FUNCTION()` and `FILTERX_GENERATOR_FUNCTION_DECLARE()` in each header, which needs to be replaced with normal function macros.

alltilla: Thanks, fixed.

MrAnno: The `FILTERX_GENERATOR_FUNCTION()` definitions at the bottom of the C files are still there.
They forwarded to the same macro, but it would be nice to get rid of them.

alltilla: Yeah. I think it is fixed now. I also found another remaining fillable reference in the unit tests, which I also removed.

PR #709, opened 2025-07-08
`loggen`: add `--client-port`

alltilla:
It is useful if we want to control which outbounds
port is used to avoid possible collisions with
server ports.

### DIFF
--- a/news/feature-709.md
+++ b/news/feature-709.md
@@ -1,0 +1,1 @@
+`loggen`: Added `--client-port` option to set the outbound (client) port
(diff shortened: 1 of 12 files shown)


MrAnno:
When no client address or port is bound, an ephemeral port is chosen on Linux that is guaranteed to be free.
Where does the collision happen?

MrAnno:
I think the best would be if preallocated ports in Light using the OS's ephemeral port allocation ability. That would avoid all future collisions as well.

Nevertheless, `--client-port` for loggen may be useful for something else in the future, so I can approve the PR.

PR #710, opened 2025-07-08
light: use lower initial value for port allocator

alltilla:
Increases the number of available non-ephemeral
ports.


mitzkia (approved):
Thanks

PR #716, opened 2025-07-15
filterx/test_scope: use simpler filterx_scope_new() constructor

bazsi:


MrAnno:
This was intentionally since `filterx_scope_new()` is only used in test code, not in production.

MrAnno (approved):
I don't mind replacing these, but my intention was to test functions as we use them in production code.

bazsi:
> I don't mind replacing these, but my intention was to test functions as we use them in production code.

we could do that in a single testcase, but I don't think we need to do that in all of them.

PR #718, opened 2025-07-16
syslog-parser: accept timestamps without colon in time-numoffset

MrAnno:
For example, 2025-07-16T04:42:54+0000


MrAnno:
Please merge this only if we found a real device producing such logs.

PR #720, opened 2025-07-16
filterx: add string slicing operator 

MrAnno:
For example:
```
filterx {
    str = "example";
    idx = 3;
    str[idx..5] == "mp";
    str[..idx] == "exa";
    str[idx..] == "mple";
};
```

### DIFF
--- a/lib/filterx/expr-string-operators.c
+++ b/lib/filterx/expr-string-operators.c
@@ -1,0 +1,250 @@
+/*
+ * Copyright (c) 2025 László Várady
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "expr-string-operators.h"
+#include "filterx/expr-literal.h"
+#include "filterx/filterx-eval.h"
+#include "filterx/object-string.h"
+#include "filterx/object-extractor.h"
+
+
+typedef struct FilterXSlicingOperator_
+{
+  FilterXExpr super;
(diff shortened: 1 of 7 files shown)


MrAnno:
I ended up using Rust's slicing syntax instead of Python's `:`, because we have plans to support ranges in switch-case, and we'd like to use the same `..` keyword there too.

lib/filterx/expr-string-operators.c
```diff
@@ -0,0 +1,247 @@
...
+  const gchar *str = filterx_string_get_value_ref(lhs_object, &str_len);
+  if(!str)
+    {
+      filterx_eval_push_error_info_printf("Failed to evaluate string slicing operator", &self->super,
+                                          "Left hand side must be a string, got: %s",
+                                          filterx_object_get_type_name(lhs_object));
+      filterx_object_unref(lhs_object);
+      return NULL;
+    }
+
```
sodomelle, line 142: nitpick: extra new line

lib/filterx/expr-string-operators.c
```diff
@@ -0,0 +1,247 @@
...
+                                          "Failed to evaluate start index");
+      return FALSE;
+    }
+
+  if (!filterx_object_extract_integer(start, start_idx))
+    {
+      filterx_eval_push_error_info_printf("Failed to evaluate string slicing operator", &self->super,
+                                          "Start index must be an integer, got: %s",
+                                          filterx_object_get_type_name(start));
+      return FALSE;
```
sodomelle, line 82: I think `filterx_object_unref(start);` is missing here.

lib/filterx/expr-string-operators.c
```diff
@@ -0,0 +1,247 @@
...
+                                          "Failed to evaluate end index");
+      return FALSE;
+    }
+
+  if(!filterx_object_extract_integer(end, end_idx))
+    {
+      filterx_eval_push_error_info_printf("Failed to evaluate string slicing operator", &self->super,
+                                          "End index must be an integer, got: %s",
+                                          filterx_object_get_type_name(end));
+      return FALSE;
```
sodomelle, line 112: I think `filterx_object_unref(end);` is missing here.

lib/filterx/expr-string-operators.c
MrAnno: FilterX error handling is so lenghty, I would prefer keeping this extra line of separation here.
An alternative would be to introduce an `_extract_lhs()` function too.

lib/filterx/expr-string-operators.c
MrAnno: Thanks for catching this. I only did positive tests under valgrind.

lib/filterx/expr-string-operators.c
sodomelle: I see. Both of them makes sense to me, maybe the `_extract_lhs()` is a bit more intuitive. But it really is nitpicking, so I resolve the conversation.

PR #721, opened 2025-07-17
filterx/filterx-scope: write-protected FilterXScope

bazsi:
By making FilterXScope initially write protected and only making it writable at the first actual write, we can delay cloning of variables for read-only access.

For instance, given foo is a dict:

foo = {"foo":{"foo2":"foovalue"}};

And then in a subsequent filterx block we do this:

foo.foo.foo2

Then the following happens without this patch:
  * "foo" is fetched from the parent scope
  * filterx_scope_clone() is invoked, which will create another ref
  * we continue evaluation

This means that unmatched branches will cause clones. This branch changes this by:
  * making the scope initially write protected
  * making it writable at strategic points
  * as long as it is read-only, mutable variables are shared between the current and the previous FilterXScope, no cloning is performed

We mark the scope as writable at the following points:
  * set-subscript, unset, assign, set-attr, update_repr

### DIFF
--- a/tests/light/functional_tests/filterx/test_filterx_scope.py
+++ b/tests/light/functional_tests/filterx/test_filterx_scope.py
@@ -339,6 +339,57 @@ def test_declared_variables_are_retained_across_scopes(config, syslog_ng):
     assert file_true.read_log() == "barka"
 
 
+def test_mutable_variables_are_inherited_and_cowed_in_parallel_branches(config, syslog_ng):
+    (file_true, file_false, file_final) = create_config(
+        config,
+        # initialize foo in the prelude, dict in dict
+        init_exprs=[
+            """
+            declare foo = {"foo":{"foo2":"foovalue"}};
+            isset(foo);
+            """,
+        ],
+        # in the true branch, inherit the variables from the parent scope and then change "foo"
+        # after changing foo, bail out to the false branch
+        true_exprs=[
+            """
+            isset(foo);
+            foo.foo.foo2 == "foovalue";
+            foo.foo.bar2 = "barvalue";
+            """,
+            """
+            isset(foo);
+            foo.foo.foo2 == "foovalue";
+            foo.foo.bar2 == "barvalue";
+            false;
+            """,
+        ],
+        # the original foo is still intact
+        false_exprs=[
+            """
+            isset(foo);
+            foo.foo.foo2 == "foovalue";
+            not isset(foo.foo.bar2);
+            $MSG = "done";
+            """,
+        ],
+        final_exprs=[
+            """
+            isset(foo);
+            foo.foo.foo2 == "foovalue";
+            not isset(foo.foo.bar2);
+            $MSG == "done";
+            """,
+        ],
+    )
+    syslog_ng.start(config)
+
+    assert file_false.get_stats()["processed"] == 1
+    assert "processed" not in file_true.get_stats()
+    assert file_false.read_log() == "done"
+    assert file_final.read_log() == "done"
+
+
 def test_declared_variables_are_retained_across_scopes_and_junctions(config, syslog_ng):
     (file_true, file_false, _) = create_config(
         config, init_exprs=[
--- a/lib/filterx/filterx-scope.c
+++ b/lib/filterx/filterx-scope.c
@@ -160,11 +160,11 @@ _register_variable(FilterXScope *self,
 }
 
 static FilterXVariable *
-_pull_variable_from_parent_scope(FilterXScope *self, FilterXScope *scope, FilterXVariable *previous_v)
+_pull_variable_from_parent_scope(FilterXScope *self, FilterXVariable *parent_variable)
 {
-  FilterXVariable *v = _register_variable(self, previous_v->variable_type, previous_v->handle);
+  FilterXVariable *v = _register_variable(self, parent_variable->variable_type, parent_variable->handle);
 
-  *v = *previous_v;
+  *v = *parent_variable;
   if (v->value)
     v->value = filterx_object_clone(v->value);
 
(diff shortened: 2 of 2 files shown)


lib/filterx/expr-compound.c
```diff
@@ -43,6 +43,8 @@ static gboolean
 _eval_expr(FilterXExpr *expr, FilterXObject **result)
 {
   FilterXObject *res = NULL;
+  if (expr->mutates_scope)
+    filterx_eval_context_make_writable(NULL);
```
MrAnno, line 55: This doesn't seem recursive.

I think this should go to `filterx_expr_eval()`.

lib/filterx/expr-variable.c
```diff
@@ -117,6 +119,7 @@ _assign(FilterXExpr *s, FilterXObject **new_value)
     }
 
   FilterXEvalContext *context = filterx_eval_get_context();
+  filterx_eval_context_make_writable(context);
```
MrAnno: I think these should go to `filterx_expr_assign()`, `filterx_expr_unset`, etc.

It seems `unset` is implemented even for `FilterXGetSubscript` and `FilterXGetAttr`.


So overall with my previous comment, everything related to scope write protection would remain inside the `FilterXExpr` abstract class (except for simple functions, but I would solve that too by introducing a mutable function declaration).

lib/filterx/expr-setattr.c
```diff
@@ -195,6 +195,7 @@ filterx_setattr_new(FilterXExpr *object, FilterXObject *attr_name, FilterXExpr *
 
   self->new_value = new_value;
   self->super.ignore_falsy_result = TRUE;
+  self->super.mutates_scope = TRUE;
```
MrAnno: This is relatively easy to miss when new expressions, operators, functions are implemented.

Can we make this somehow "mandatory to set/unset"?
For example, by adding it as a parameter for `filterx_expr_init_instance()` (which would involve all subclass constructors).

Or if we can come up with a method that would catch these errors in any other way, that would be fine too.

tests/light/functional_tests/filterx/test_filterx_scope.py
```diff
@@ -339,6 +339,57 @@ def test_declared_variables_are_retained_across_scopes(config, syslog_ng):
...
+            $MSG == "done";
+            """,
+        ],
+    )
+    syslog_ng.start(config)
+
+    assert file_false.get_stats()["processed"] == 1
+    assert "processed" not in file_true.get_stats()
+    assert file_false.read_log() == "done"
+    assert file_final.read_log() == "done"
```
MrAnno, line 390: I think I didn't fully understand the scenario this patch tries to improve.

> This means that unmatched branches will cause clone

This is probably what I don't really get.

If we have CoW working across FilterX blocks, wouldn't it just work the cheapest possible way automatically?

lib/filterx/expr-compound.c
bazsi: While I agree that this would belong to FilterXExpr, the reason this was here is that FilterXExpr does not know anything about the FilterXEvalContext layer and filterx_expr_eval() itself is an inline function, meaning we would have a circular dependency in the .h files.

One way to do this is to pass FilterXScope instance to eval() and then depend on FilterXScope, let me see how that looks like (apart from the huge patch required to pass that instance all along the eval call chain)

lib/filterx/expr-variable.c
bazsi: same applies (header circular dependency). the unset() and assign() cases are not needed anymore as those
exprs are set to mutable_scope=TRUE, meaning that by the time we arrive here, the scope is already marked as writable.

So unset/assign is covered.

There are however some special cases that need handling:
  * variable exprs do not set mutates_scope to TRUE and I don't think they should
  * but even read-only variable expression can cause the scope to be written when 
     1) a $variable is used (it is registered in the scope even if it is a read-only access) 
     2) eval_typed() is used and the variable is unmarshalled and update_repr() is called.
     These cases are now handled by the implementation of eval and update_repr in expr-variable.c

lib/filterx/expr-setattr.c
bazsi: I did this, added an explicit mutates_scope argument to filterx_expr_init_instance() and other generic initializers (function, come to mind). I am not sure this is more readable though, but it does force an explicit decision on the mutates_scope setting.

lib/filterx/expr-compound.c
bazsi: I made an attempt at this, but this is a huge changeset, and I think your LLVM jit patches would definitely break because of that. I abandoned this for now.

tests/light/functional_tests/filterx/test_filterx_scope.py
bazsi: I improved the description of the patch, hopefully that helps what the objective was.

bazsi:
Here's the result of my perf test with and without the patch:

bazsi@bzorp3:~/src/syslog-ng/syslog-ng/dbld/install$ time for i in `seq 1 200`; do cat pan_dump.txt ; done | nc -q0 localhost 2000

real	0m20.895s
user	0m0.105s
sys	0m0.730s
bazsi@bzorp3:~/src/syslog-ng/syslog-ng/dbld/install$ time for i in `seq 1 200`; do cat pan_dump.txt ; done | nc -q0 localhost 2000

real	0m19.146s
user	0m0.095s
sys	0m0.768s

The config has 48 non-matching branches and 1 matching one (the last).

MrAnno:
We talked about it privately and the recursive mutates_scope check is needed in order to make it work correctly.

> real 0m19.146s

Is this reproducible in reverse order? (I experienced deviations at this scale on my laptop while measuring FilterX performance)

lib/filterx/expr-compound.c
bazsi: I've just pushed an update to this. Now the recursivity is fixed (thanks again for doing the filterx_expr_walk() work), so the top-level expr is guaranteed to have the right effect (FXE_WRITE) in the implemented semantics.

So any FXE_WRITE expr triggers filterx_scope_make_writable()

I was again pondering on where the right place is to actually do this, but actually it is still the same spot. Apart from the inline and circular dependency thing, the right spot to make the scope writable is at the top-level statement (e.g. stmt in the grammar) and not higher.

lib/filterx/expr-variable.c
bazsi: as variable access in general is not considered a writable operation, so I'd retain these on-demand "make-writable" parts as a part of expr-variables, as update_repr and variable registration are not inherently writable but can mean that we need a writable scope.

MrAnno (approved):

PR #725, opened 2025-07-17
filterx-str: add str_replace() function

MrAnno:
For example:
```
filterx {
    dal = "érik a szőlő, hajlik a vessző";
    str_replace(dal, "a", "egy") == "érik egy szőlő, hegyjlik egy vessző";

    dal = "érik a szőlő, hajlik a vessző";
    str_replace(dal, "a", "egy", 1) == "érik egy szőlő, hajlik a vessző";
};
```

### DIFF
--- a/tests/light/functional_tests/filterx/test_filterx_funcs.py
+++ b/tests/light/functional_tests/filterx/test_filterx_funcs.py
@@ -956,3 +956,19 @@ def test_format_leef_version_2_space_delimiter(config, syslog_ng):
         "dummy_key3=value\\=containing\\=value\\=separators\\=that\\=needs\\=to\\=be\\=escaped "
         "dummy_key4=value\\\\containing\\\\backslash\\\\that\\\\needs\\\\to\\\\be\\\\backslash\\\\escaped"
     )
+
+
+def test_str_replace(config, syslog_ng):
+    (file_final,) = create_config(
+        config, r"""
+    dal = "érik a szőlő, hajlik a vessző";
+    $MSG = {
+        "replace_all": str_replace(dal, "a", "egy"),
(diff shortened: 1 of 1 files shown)


bazsi:
This looks good to me, but I think a light test in test_filterx_funcs would be a great addition.

lib/filterx/func-str-transform.c
```diff
@@ -91,3 +92,59 @@ filterx_simple_function_upper(FilterXExpr *s, FilterXObject *args[], gsize args_
...
+  gint64 limit_num = 0;
+  if (limit && !filterx_object_extract_integer(limit, &limit_num))
+    {
+      filterx_simple_function_argument_error(s, "Fourth argument must be integer");
+      return NULL;
+    }
+
+  GString *replaced = g_string_new(str);
+
+  g_string_replace(replaced, old_str, new_str, limit_num);
```
sodomelle, line 144: Nitpick: I like it when `(void)` is used to express ignoring the return value, not to generate compiler warnings.

sodomelle (approved):
Only a nitpick. Looks good.

lib/filterx/func-str-transform.c
MrAnno: We don't use `Wunused-result`, because there would be hundreds of calls in AxoSyslog that violate this.

I think it is okay to ignore return values as long as they are not about error handling (like here).
If it's about error handling, it should be either handled, or if failure is impossible in the given invocation, then it should be marked with `g_assert(f())`;

MrAnno: But if you get a warning with any compilers, that should be fixed, of course.

PR #729, opened 2025-07-22
light: Add various improvements

mitzkia:
- light: bump version to 1.9.0
- light/syslog-ng: Mark process as None during a detected crash
- light/ProcessExecutor: print a debug log why process can not started
- light/Start Parameter: add gdb_for_bt as a new --run-under start parameter

### DIFF
--- a/tests/light/src/axosyslog_light/syslog_ng/syslog_ng_local_executor.py
+++ b/tests/light/src/axosyslog_light/syslog_ng/syslog_ng_local_executor.py
@@ -27,20 +27,24 @@ import typing
 from pathlib import Path
 from subprocess import Popen
 
+from axosyslog_light.common.file import copy_shared_file
 from axosyslog_light.executors.command_executor import CommandExecutor
 from axosyslog_light.executors.process_executor import ProcessExecutor
 from axosyslog_light.syslog_ng.syslog_ng_executor import SyslogNgExecutor
 from axosyslog_light.syslog_ng.syslog_ng_executor import SyslogNgStartParams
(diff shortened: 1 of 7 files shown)


tests/light/src/axosyslog_light/syslog_ng/syslog_ng_local_executor.py
```diff
@@ -99,6 +101,31 @@ def run_process_with_gdb(
...
+        self,
+        start_params: SyslogNgStartParams,
+        stderr_path: Path,
+        stdout_path: Path,
+    ) -> Popen:
+        gdb_exec_file_content = f"""#!/bin/bash
+    exec gdb <<EOF
+    file {self.__syslog_ng_binary_path}
+    run {shlex.join(start_params.format())}
+    bt full
```
MrAnno: We probably want thread apply all bt full here.

Wouldn't it be better to commit this into a separate sh file?

mitzkia: ok, I agreed with you.
1, I will replace 'bt ful' with 'thread apply all bt full'
2, I will move this script into a separate shared file

mitzkia: fixed

MrAnno: Thank you.

mitzkia:
Thanks, fixed

PR #731, opened 2025-07-23
`http`: support templating in `body-prefix()`

alltilla:
In case of batching the templates in `body-prefix()` will be calculated from the first message. Make sure to use `worker-partition-key()` to group similar messages together.

Literal dollar signs (`$`) used in `body-prefix()` must be escaped like `$$`.

Example usage:
```
http(
  ...
  body-prefix('{"log_type": "$log_type", "entries": [')
  body('"$MSG"')
  delimiter(",")
  body-suffix("]}")

  batch-lines(1000)
  batch-timeout(1000)
  worker-partition-key("$log_type")
);
```

### DIFF
--- a/news/feature-731.md
+++ b/news/feature-731.md
@@ -1,0 +1,22 @@
+`http`: Added templating support to `body-prefix()`
+
+In case of batching the templates in `body-prefix()` will be calculated
+from the first message. Make sure to use `worker-partition-key()` to
+group similar messages together.
+
+Literal dollar signs (`$`) used in `body-prefix()` must be escaped like `$$`.
+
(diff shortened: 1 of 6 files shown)


news/feature-731.md
```diff
@@ -0,0 +1,20 @@
+`http`: Added templating support to `body-prefix()`
```
MrAnno, line 1: I would mention that if someone used `$` previously in this option, it should be escaped  to `$$`.

alltilla: Thanks, added.

MrAnno (approved):

PR #732, opened 2025-07-24
Add ADC credentials option

sodomelle:
Example usage:
```
destination {
    	google-pubsub-grpc(
      		project("test")
                topic("test")
      		auth(adc(service-account-key("absolute path to file")))
       );
};
```

### DIFF
--- a/modules/grpc/common/credentials/grpc-credentials-builder.cpp
+++ b/modules/grpc/common/credentials/grpc-credentials-builder.cpp
@@ -231,6 +231,12 @@ ClientCredentialsBuilder::set_service_account_validity_duration(guint64 validity
   service_account.validity_duration = validity_duration;
 }
 
+void
+ClientCredentialsBuilder::set_adc_service_account_key(const char *key_path)
+{
+  adc_service_account_key = key_path;
+}
+
 bool
 ClientCredentialsBuilder::validate() const
 {
@@ -271,7 +277,31 @@ ClientCredentialsBuilder::build() const
     case GCAM_ALTS:
       return ::grpc::experimental::AltsCredentials(alts_credentials_options);
     case GCAM_ADC:
+    {
+      if (!adc_service_account_key.empty())
+        {
+          const char *env_var_name = "GOOGLE_APPLICATION_CREDENTIALS";
+          const char *key_path_orig = getenv(env_var_name);
+          if (setenv(env_var_name, adc_service_account_key.c_str(), 1) != 0)
+            {
+              msg_error("gRPC: setenv failed for GOOGLE_APPLICATION_CREDENTIALS");
+              return nullptr;
+            }
+          std::shared_ptr<::grpc::ChannelCredentials> creds = ::grpc::GoogleDefaultCredentials();
+          if (!key_path_orig)
+            {
+              unsetenv(env_var_name);
+              return creds;
+            }
+          if (setenv(env_var_name, key_path_orig, 1) != 0)
+            {
+              msg_error("gRPC: resetting GOOGLE_APPLICATION_CREDENTIALS failed");
+              return nullptr;
+            }
+          return creds;
+        }
       return ::grpc::GoogleDefaultCredentials();
+    }
     case GCAM_SERVICE_ACCOUNT:
     {
       auto channel_creds = ::grpc::SslCredentials(::grpc::SslCredentialsOptions());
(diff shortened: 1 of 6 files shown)


modules/grpc/common/credentials/grpc-credentials-builder.hpp
```diff
@@ -92,6 +94,8 @@ class ClientCredentialsBuilder
     std::string key;
     guint64 validity_duration = 3600L;
   } service_account;
+
+  std::string application_credentials;
```
alltilla: [nitpick]

Could you put an ADC comment above it? Thanks!

modules/grpc/common/credentials/grpc-credentials-builder.cpp
```diff
@@ -271,7 +277,14 @@ ClientCredentialsBuilder::build() const
...
       return ::grpc::experimental::AltsCredentials(alts_credentials_options);
     case GCAM_ADC:
+    {
+      if (!application_credentials.empty())
+        {
+          if (setenv("GOOGLE_APPLICATION_CREDENTIALS", application_credentials.c_str(), 1) != 0)
+            msg_info("gRPC: setenv failed for GOOGLE_APPLICATION_CREDENTIALS");
+        }
       return ::grpc::GoogleDefaultCredentials();
+    }
```
alltilla: We should store the old value and restore it after the `GoogleDefaultCredentials` call.

modules/grpc/common/credentials/grpc-credentials-builder.cpp
```diff
@@ -231,6 +231,12 @@ ClientCredentialsBuilder::set_service_account_validity_duration(guint64 validity
   service_account.validity_duration = validity_duration;
 }
 
+void
+ClientCredentialsBuilder::set_application_credentials(const char *key_path)
```
alltilla: Can we call this something similar to `service-account()`'s `key()`? They serve the same purpose.
For example `service-account-key()`?

modules/grpc/common/credentials/grpc-credentials-builder.cpp
```diff
@@ -271,7 +277,14 @@ ClientCredentialsBuilder::build() const
     case GCAM_ALTS:
       return ::grpc::experimental::AltsCredentials(alts_credentials_options);
     case GCAM_ADC:
+    {
+      if (!application_credentials.empty())
+        {
+          if (setenv("GOOGLE_APPLICATION_CREDENTIALS", application_credentials.c_str(), 1) != 0)
+            msg_info("gRPC: setenv failed for GOOGLE_APPLICATION_CREDENTIALS");
```
alltilla: This might be a failure scenario.

modules/grpc/common/credentials/grpc-credentials-builder.hpp
sodomelle: Done. Added one for the new method above too.

modules/grpc/common/credentials/grpc-credentials-builder.cpp
sodomelle: Done.

modules/grpc/common/credentials/grpc-credentials-builder.cpp
sodomelle: Renamed it.

modules/grpc/common/credentials/grpc-credentials-builder.cpp
sodomelle: Changed the logging to `error` level. Returning `nullptr` when it happens.

modules/grpc/common/credentials/grpc-credentials-builder.cpp
```diff
@@ -271,7 +277,25 @@ ClientCredentialsBuilder::build() const
...
+            {
+              msg_error("gRPC: setenv failed for GOOGLE_APPLICATION_CREDENTIALS");
+              return nullptr;
+            }
+          std::shared_ptr<::grpc::ChannelCredentials> creds = ::grpc::GoogleDefaultCredentials();
+          if (setenv("GOOGLE_APPLICATION_CREDENTIALS", key_path_orig, 1) != 0)
+            {
+              msg_error("gRPC: resetting GOOGLE_APPLICATION_CREDENTIALS failed");
+              return nullptr;
+            }
```
alltilla: I think we should handle the case where `key_path_orig` is `null` (was not set before) with `unsetenv()`.

sodomelle: Done.

alltilla (approved):

bazsi:
this looks good to me. is there a reason this is not yet merged?

PR #733, opened 2025-07-24
light: Add clickhouse dest related testcases

mitzkia:

PR #735, opened 2025-07-25
dbld/clickhouse: install clickhouse-server for devshell image

mitzkia:
- This PR is required for #733 
-

### DIFF
--- a/dbld/images/devshell.dockerfile
+++ b/dbld/images/devshell.dockerfile
@@ -11,3 +11,4 @@ RUN /dbld/builddeps enable_dbgsyms
 RUN /dbld/builddeps install_perf
 
 RUN /dbld/builddeps install_apt_packages
+RUN /dbld/builddeps install_clickhouse_server_for_light
--- a/dbld/builddeps
+++ b/dbld/builddeps
@@ -231,6 +231,14 @@ function install_bison_from_source {
     make && make install
 }
 
+function install_clickhouse_server_for_light {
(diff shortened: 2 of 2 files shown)


dbld/images/devshell.dockerfile
```diff
@@ -11,3 +11,4 @@ RUN /dbld/builddeps enable_dbgsyms
 RUN /dbld/builddeps install_perf
 
 RUN /dbld/builddeps install_apt_packages
+RUN /dbld/builddeps install_clickhouse_server_for_light
```
MrAnno, line 14: Do we want to add all module's server components to the devshell when we write test to them?
An alternative would be to start a separate container in the CI with shared networking or create an image built upon devshell.

But running light tests is part of the dev process, so maybe it's the right place.

mitzkia: Thanks for the review. 
- With the current approach we can manage clickhouse-server start/stop in each testcase (we need to start/stop only a process in a same container where syslog-ng + testcases are running), which is a good thing.
- if a clickhouse-server would live in a different container it needs an extra effort to add it to dbld logic - how - and when to start it (we can not start/stop it from light - due to it runs in a different docker container) - and because of this complexity I would not like to open this scope now - my goal was to concentrate only for clickhouse destination tests. 

I would like to also mention I am not against to think over this idea again - to put 3rd party test dependencies (like clickhouse-server) into a separate container and my first idea was this too :)

MrAnno: Okay, it's not worth the trouble.

PR #736, opened 2025-07-25
FilterX: null and error-safe dict elements

MrAnno:
For example, the following fields won't be set:
```
$MSG = {
    "nullidontwant":?? null,
    "erroridontwant":?? nonexistingvar,
};
```

### DIFF
--- a/lib/filterx/expr-literal-container.c
+++ b/lib/filterx/expr-literal-container.c
@@ -26,6 +26,9 @@
 #include "filterx/object-dict.h"
 #include "filterx/object-list.h"
 #include "filterx/filterx-eval.h"
+#include "filterx/filterx-object.h"
+#include "filterx/object-null.h"
+#include "filterx/object-message-value.h"
 
 /* Object Members (e.g. key-value) */
 
@@ -33,7 +36,8 @@ struct FilterXLiteralElement_
 {
   FilterXExpr *key;
   FilterXExpr *value;
-  gboolean literal;
+  guint8 nullv:1,
+         literal:1;
 };
 
 static gboolean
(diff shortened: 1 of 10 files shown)


lib/filterx/expr-literal-container.c
```diff
@@ -126,6 +141,23 @@ _literal_container_eval(FilterXExpr *s)
         }
 
       FilterXObject *value = filterx_expr_eval(elem->value);
+      if (elem->nullv)
+        {
+          if (!value)
+            filterx_eval_dump_errors("FilterX: null coalesce assignment suppressing error");
+
+          gboolean value_is_null_or_error = !value || filterx_object_is_type(value, &FILTERX_TYPE_NAME(null))
```
alltilla: [optional]

```c
gboolean value_is_null_or_error = !value || filterx_object_extract_null(value);
```

alltilla (dismissed):

sodomelle (approved):
Looks good. I won't merge as @alltilla  has an optional comment. Please decide if you want to do it or not. Then we can merge IMO.

lib/filterx/expr-literal-container.c
MrAnno: Fixed.

lib/filterx/expr-literal-container.c
```diff
@@ -126,6 +141,23 @@ _literal_container_eval(FilterXExpr *s)
...
+                                            || (filterx_object_is_type(value, &FILTERX_TYPE_NAME(message_value))
+                                                && filterx_message_value_get_type(value) == LM_VT_NULL);
```
bazsi: just one question: 

The same null/error coalescing logic is implemented in other places as well (like expr-assign), shouldn't we introduce a filterx_object_is_coalescable() (or a better name)

bazsi:
I like this, however I'd be happy to see the centralization of the decision  whether a value is to be coalesced

lib/filterx/expr-literal-container.c
MrAnno: The repetition has been reduced to a simple `filterx_object_extract_null()` call.

alltilla (approved):

PR #738, opened 2025-07-26
`cloud-auth()`: add `scope()` option for `gcp(service-account())`

alltilla:
Can be used for authentications using scope instead of audiance.
For more info, see: https://google.aip.dev/auth/4111#scope-vs-audience

Example usage:
```
http(
  ...
  cloud-auth(
    gcp(
      service-account(
        key("/path/to/secret.json")
        scope("https://www.googleapis.com/auth/example-scope")
      )
    )
  )
);
```

### DIFF
--- a/modules/cloud-auth/cloud-auth-grammar.ym
+++ b/modules/cloud-auth/cloud-auth-grammar.ym
@@ -1,5 +1,6 @@
 /*
- * Copyright (c) 2023 Attila Szakacs
+ * Copyright (c) 2025 Axoflow
+ * Copyright (c) 2025 Attila Szakacs <attila.szakacs@axoflow.com>
  *
  * This program is free software: you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


modules/cloud-auth/cloud-auth-grammar.ym
```diff
@@ -1,5 +1,6 @@
 /*
- * Copyright (c) 2023 Attila Szakacs
+ * Copyright (c) 2025 Axoflow
+ * Copyright (c) 2025 Attila Szakacs <attila.szakacs@axoflow.com>
```
MrAnno: `2023-2025` here and the other files too (would be important to keep the original year too).

modules/cloud-auth/google-auth.cpp
```diff
@@ -321,6 +337,15 @@ google_authenticator_set_service_account_audience(CloudAuthenticator *s, const g
   self->service_account_options.audience = g_strdup(audience);
 }
 
+void
+google_authenticator_set_service_account_scope(CloudAuthenticator *s, const gchar *scope)
+{
+  GoogleAuthenticator *self = (GoogleAuthenticator *) s;
+
+  g_free(self->service_account_options.scope);
+  self->service_account_options.scope = g_strdup(scope);
```
MrAnno, line 346: This is leaked.

modules/cloud-auth/cloud-auth-grammar.ym
alltilla: Fixed.

modules/cloud-auth/google-auth.cpp
alltilla: Nice catch! Fixed.

MrAnno (approved):

PR #741, opened 2025-07-29
modules/xml: add error message on missing xmlns attribute

bazsi:
I tried to parse a sample of Windows  Eventlog XML and I didn't get an error message, which took me a while to recognize.

I used an incorrect sample, which I didn't realize. This should make it easier to find these cases.

### DIFF
--- a/modules/xml/filterx-parse-windows-eventlog-xml.c
+++ b/modules/xml/filterx-parse-windows-eventlog-xml.c
@@ -153,21 +153,22 @@ _collect_attrs(const gchar **attribute_names, const gchar **attribute_values,
 static gboolean
 _has_valid_schema_url(const gchar **attribute_names, const gchar **attribute_values, GError **error)
 {
-  if (!attribute_names[0])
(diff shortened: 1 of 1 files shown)


modules/xml/filterx-parse-windows-eventlog-xml.c
```diff
@@ -153,21 +153,22 @@ _collect_attrs(const gchar **attribute_names, const gchar **attribute_values,
...
     {
-      _set_error(error, "unexpected schema URL: %s", attribute_values[0]);
+      _set_error(error, "unexpected schema URL in the Event element: %s", attribute_values[0]);
       return FALSE;
     }
 
   if (attribute_names[1])
     {
-      _set_error(error, "unexpected attribute in Event, number of attributes must be 1, got: %s", attribute_names[1]);
+      _set_error(error, "unexpected attribute in the Event element:", attribute_names[1]);
```
alltilla: `%s` is missing

PR #742, opened 2025-07-29
Filterx: expr init must be called before eval

bazsi:
This PR makes sure that all expressions are initialized before being evaluated.

High level summary of the changes:
* unit tests will now invoke filterx_expr_init() before calling eval
* extracting literal arguments in function calls will now use filterx_literal_get_value() instead of evaluating the expr. This seems to be the same in the case of literals, but get_value() can be called without init and eval() can assert on being initialized
* fixes an strptime() error with init/deinit/optimize
*

### DIFF
--- a/lib/filterx/func-failure-info.c
+++ b/lib/filterx/func-failure-info.c
@@ -47,7 +47,7 @@ typedef struct FilterXFunctionFailureInfoEnable_
 typedef struct FilterXFunctionFailureInfoMeta_
 {
   FilterXFunction super;
-  FilterXObject *metadata;
+  FilterXExpr *metadata_expr;
 } FilterXFunctionFailureInfoMeta;
 
 static inline void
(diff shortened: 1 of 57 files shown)


lib/filterx/func-failure-info.c
```diff
@@ -47,7 +47,7 @@ typedef struct FilterXFunctionFailureInfoEnable_
 typedef struct FilterXFunctionFailureInfoMeta_
 {
   FilterXFunction super;
-  FilterXObject *metadata;
+  FilterXExpr *metadata_expr;
```
alltilla, line 50: Don't we need init, deinit and optimize for this?

bazsi: I think we have that, here:

https://github.com/axoflow/axosyslog/pull/742/commits/ad8990f5670f03da0f4ea01ebcdb78d717102427#diff-e3269fc27150fef2e2c2758bb2089f7c33929854e037102b9b58cba424766171R136

alltilla: Yes, I missed that, sorry.

PR #743, opened 2025-07-30
FilterX: fix escaping in format_xml and format_windows_eventlog_xml functions

sodomelle:
Example:
`<b>` -> `&lt;b&gt;`

### DIFF
--- a/modules/xml/filterx-func-format-windows-eventlog-xml.c
+++ b/modules/xml/filterx-func-format-windows-eventlog-xml.c
@@ -54,7 +54,11 @@ _append_inner_data_dict_element(FilterXObject *key, FilterXObject *value, gpoint
     }
 
   if (value_str_len)
-    g_string_append_printf(buffer, "<Data Name='%s'>%s</Data>", key_str, value_str);
+    {
+      gchar *escaped_value = g_markup_escape_text(value_str, value_str_len);
+      g_string_append_printf(buffer, "<Data Name='%s'>%s</Data>", key_str, escaped_value);
+      g_free(escaped_value);
+    }
   else
     g_string_append_printf(buffer, "<Data Name='%s' />", key_str);
   return TRUE;
@@ -87,7 +91,9 @@ _append_data_element(FilterXObject *key, FilterXObject *value, gpointer user_dat
       return FALSE;
     }
 
-  self->append_leaf(key_str, value_str, value_str_len, buffer);
+  gchar *escaped_value = g_markup_escape_text(value_str, value_str_len);
+  self->append_leaf(key_str, escaped_value, strlen(escaped_value), buffer);
+  g_free (escaped_value);
   return TRUE;
 }
 
--- a/modules/xml/filterx-func-format-xml.c
+++ b/modules/xml/filterx-func-format-xml.c
@@ -107,23 +107,27 @@ append_list(FilterXObject *key, FilterXObject *list, gpointer user_data)
 }
 
 static void
-_append_attribute(const char *key_str, const char *value_str, GString *buffer)
+_append_attribute(const char *key_str, GString *value_buffer, GString *buffer)
 {
   if (buffer->str[buffer->len - 1] == '>')
     g_string_overwrite(buffer, buffer->len - 1, " ");
   else
     g_string_append_c(buffer, ' ');
 
-  g_string_append_printf(buffer, "%s='%s'>", &key_str[1], value_str);
+  gchar *escaped_value = g_markup_escape_text(value_buffer->str, value_buffer->len);
+  g_string_append_printf(buffer, "%s='%s'>", &key_str[1], escaped_value);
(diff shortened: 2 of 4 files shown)


modules/xml/filterx-func-format-windows-eventlog-xml.c
```diff
@@ -54,7 +54,7 @@ _append_inner_data_dict_element(FilterXObject *key, FilterXObject *value, gpoint
     }
 
   if (value_str_len)
-    g_string_append_printf(buffer, "<Data Name='%s'>%s</Data>", key_str, value_str);
+    g_string_append_printf(buffer, "<Data Name='%s'>%s</Data>", key_str, g_markup_escape_text(value_str, value_str_len));
```
MrAnno: All `g_markup_escape_text()` invocations leak here, we need to pass ownership where possible, and deallocate the memory otherwise.

modules/xml/filterx-func-format-windows-eventlog-xml.c
```diff
@@ -54,7 +54,7 @@ _append_inner_data_dict_element(FilterXObject *key, FilterXObject *value, gpoint
...
   else
     g_string_append_printf(buffer, "<Data Name='%s' />", key_str);
```
MrAnno, line 63: Shouldn't we escape keys?

modules/xml/filterx-func-format-xml.c
```diff
@@ -158,17 +158,17 @@ _append_entry(FilterXObject *key, FilterXObject *value, gpointer user_data)
   if (key_str_len && (key_str[0] == '@'))
     {
       *is_only_attribute_present = TRUE;
-      _append_attribute(key_str, val_buf->str, buffer);
+      _append_attribute(key_str, g_markup_escape_text(val_buf->str, val_buf->len), buffer);
```
MrAnno: I think it would be more secure to implement escaping inside `_append_text`, so we won't miss any call sites.

Do we need this in ``_append_attribute()` as well?

modules/xml/filterx-func-format-windows-eventlog-xml.c
MrAnno: No, `g_markup_escape_text()` is only for text escaping (<>).

What about `''` escaping in attributes (here and in the generic xml formatter too)?

tests/light/functional_tests/filterx/test_filterx.py
```diff
@@ -2863,7 +2864,8 @@ def test_format_xml_valid_input(config, syslog_ng):
         r""""multiple_root":"<a>b</a><a>c</a>","""
         r""""integer_leaf":"<a>100</a>","""
         r""""float_leaf":"<a>100.0</a>","""
-        r""""datetime_leaf":"<a>946677600.000000</a>"}"""
+        r""""datetime_leaf":"<a>946677600.000000</a>","""
+        r""""escaped":"<a>&lt;b&gt;</a>"}"""
```
MrAnno, line 2868: Please write a bugfix entry for this.

modules/xml/filterx-func-format-windows-eventlog-xml.c
sodomelle: keys are validated in: `_is_valid_xml_tag_name`

modules/xml/filterx-func-format-windows-eventlog-xml.c
sodomelle: Thanks, added freeing the escaped string.

modules/xml/filterx-func-format-xml.c
sodomelle: I moved it there. I think we should escape attributes too, as they may contain special charactres.

tests/light/functional_tests/filterx/test_filterx.py
sodomelle: Added bugfix entry.

modules/xml/filterx-func-format-xml.c
```diff
@@ -107,23 +107,27 @@ append_list(FilterXObject *key, FilterXObject *list, gpointer user_data)
...
+_append_attribute(const char *key_str, GString *value_buffer, GString *buffer)
 {
   if (buffer->str[buffer->len - 1] == '>')
     g_string_overwrite(buffer, buffer->len - 1, " ");
   else
     g_string_append_c(buffer, ' ');
 
-  g_string_append_printf(buffer, "%s='%s'>", &key_str[1], value_str);
+  gchar *escaped_value = g_markup_escape_text(value_buffer->str, value_buffer->len);
+  g_string_append_printf(buffer, "%s='%s'>", &key_str[1], escaped_value);
```
MrAnno, line 118: Everywhere we format attribute values, we should also escape `'`.

We have, for example, `str_repr_encode()` for that.

PR #746, opened 2025-07-31
`dpath()`: new function to set a potentially non-existing path within a dict

MrAnno:
For example,
```
dpath(dict.path.to["complex.n-v"].create) = {...};
```

### DIFF
--- a/lib/filterx/filterx-dpath.c
+++ b/lib/filterx/filterx-dpath.c
@@ -256,7 +256,7 @@ filterx_dpath_lvalue_optimize(FilterXExpr *s)
               if (!literal)
                 return NULL;
 
-              elem->data = filterx_dpath_elem_object_new(literal);;
+              elem->data = filterx_dpath_elem_object_new(literal);
               filterx_dpath_elem_free(dpath_elem);
             }
         }
@@ -278,7 +278,17 @@ filterx_dpath_lvalue_init(FilterXExpr *s, GlobalConfig *cfg)
       if (dpath_elem->type == FILTERX_DPATH_ELEMENT_EXPR)
         {
           if (!filterx_expr_init(dpath_elem->expr, cfg))
-            return FALSE;
+            {
+              for (GList *d = self->dpath_elements; d != elem; d = d->next)
+                {
+                  FilterXDPathElement *de = (FilterXDPathElement *) elem->data;
+                  if (de->type == FILTERX_DPATH_ELEMENT_EXPR)
+                    filterx_expr_deinit(de->expr, cfg);
+                }
+
+              filterx_expr_deinit(self->variable, cfg);
+              return FALSE;
+            }
         }
 
       self->last_dpath_element = dpath_elem;


bazsi:
there's a similar patch on the get side here: https://github.com/axoflow/axosyslog/pull/450 but the syntax is diffrent.

MrAnno:
#450 looks like an optimized getter method. ` dpath_set()` is meant for creating the path inside the dict.

My preferred syntax would have been this:
```
dpath(a.b.c.d) += {};
```

This wouldn't be a real function, just a magic syntax, in order to avoid introducing references into the FilterX language.
But I ended up implementing this simple function instead.

sodomelle (dismissed):

MrAnno:
I think I would rather go with the syntax discussed in #450. @jszigetvari Is that okay?

MrAnno:
@sodomelle Sorry, I had to reimplement half of it.

lib/filterx/filterx-dpath.c
```diff
@@ -0,0 +1,319 @@
...
+
+  gboolean result;
+  if (self->append_mode)
+    {
+      if (!filterx_object_is_type_or_ref(*new_value, &FILTERX_TYPE_NAME(dict)))
+        {
+          filterx_eval_push_error("Right hand side of dpath()+= must be a dict", s, NULL);
+          filterx_object_unref(last_element);
+          return FALSE;
+        }
```
alltilla: We could handle any type here with their `add()` method.

alltilla (dismissed):

lib/filterx/filterx-dpath.c
MrAnno: I'll fix this.

lib/filterx/filterx-dpath.c
```diff
@@ -0,0 +1,319 @@
...
+  if (!filterx_expr_init(self->variable, cfg))
+    return FALSE;
+
+  for (GList *elem = self->dpath_elements; elem; elem = elem->next)
+    {
+      FilterXDPathElement *dpath_elem = (FilterXDPathElement *) elem->data;
+      if (dpath_elem->type == FILTERX_DPATH_ELEMENT_EXPR)
+        {
+          if (!filterx_expr_init(dpath_elem->expr, cfg))
+            return FALSE;
```
sodomelle: Shouldn't we `deinit` already initialized expressions, when an error occurs here? It is a loop, and when for example initializing the 3rd element fails, right now we don't `deinit` the previous 2. Or am I wrong? :D
Also `self-> variable` is not deinited in error case.

lib/filterx/filterx-dpath.c
MrAnno: Fixed, it was a nightmare to track all those new_value refs :)

lib/filterx/filterx-dpath.c
MrAnno: You are right. Fixed, thanks.

sodomelle (approved):
Looks good to me.

PR #753, opened 2025-08-08
logthrdest: fix slow destinations when flow-control is "misconfigured"

MrAnno:
log-iw-size()/max-connections() had to be greater than batch-lines()*workers() when flow-control was enabled, otherwise message delivery was slowed down artificially.

The given source stopped reading messages because of flow-control, but the destination didn't ack messages because it was waiting for the batch to be complete. After batch-timeout() milliseconds, the batch was finally sent out, and this repeated again and again.


sodomelle (approved):
Looks good to me.

PR #757, opened 2025-08-13
parallelize(): batch-size() and minor improvements

MrAnno:
Notes:
- `log-iw-size()` must be set to a greater value to make parallelization effective.
- `batch-size()` must be equal to `log-fetch-limit()` of the sources, otherwise not all processor cores will be utilized. This is due to the complicated mechanism of how worker jobs are scheduled into different ivykis threads (idle threads + fetch-limit trigger).


lib/logscheduler.c
```diff
@@ -385,13 +385,21 @@ log_scheduler_options_defaults(LogSchedulerOptions *options)
...
+
 gboolean
 log_scheduler_options_init(LogSchedulerOptions *options, GlobalConfig *cfg)
 {
   if (options->num_partitions == -1)
     options->num_partitions = 0;
   if (options->num_partitions > LOGSCHEDULER_MAX_PARTITIONS)
-    options->num_partitions = LOGSCHEDULER_MAX_PARTITIONS;
+    {
+      msg_warning("parallelize() currently supports up to " TOSTRING(LOGSCHEDULER_MAX_PARTITIONS) " workers",
```
bazsi: please add a "WARNING:" prefix and also that we limited the workers to the precompiled value.

lib/logscheduler.c
```diff
@@ -229,21 +240,25 @@ _queue_thread(LogScheduler *self, LogSchedulerThreadState *thread_state, LogMess
...
+  guint partition_index = _get_partition_index(self, thread_state, msg, &should_flush);
 
   LogMessageQueueNode *node;
   node = log_msg_alloc_queue_node(msg, path_options);
   iv_list_add_tail(&node->list, &thread_state->batch_by_partition[partition_index]);
   thread_state->num_messages++;
   log_msg_unref(msg);
+
+  if (should_flush)
+    _move_batch_to_partition(self, thread_state, partition_index);
```
MrAnno: This actually degrades performance. I can't explain why (maybe lock contention?), but if we assume that  input threads are an order of magnitude faster than the processing threads, this extra flush is not important.

MrAnno: Reverted

PR #761, opened 2025-08-15
Clickhouse destination json mode

bshifter:
This PR introduces an alternative way to send data into the ClickHouse database.
By setting the new `json-var()` option, the ClickHouse destination ingests data in a schemaless way (FORMAT JSONEachRow) over Protobuf.

The following directives are mutually exclusive with `json-var()`:
- schema()
- protobuf-schema()
- server-side-schema()

The `json-var()` option’s argument must be either a template name or a template containing a JSON string with a JSON type hint.

example;
```
    destination {
      clickhouse (
      ...
      json-var("$json_data");
or
      json-var(json("{\"ingest_time\":1755248921000000000, \"body\": \"test template\"}"))      
      ...
      };
   };
```

### DIFF
--- a/news/feature-761.md
+++ b/news/feature-761.md
@@ -1,0 +1,18 @@
+`clickhouse-destination()`: new `json-var()` directive
+
+The new `json-var()` option accepts either a JSON template or a variable containing a JSON string, and sends it to the ClickHouse server in Protobuf/JSON mixed mode (JSONEachRow format). In this mode, type validation is performed by the ClickHouse server itself, so no Protobuf schema is required for communication.
+
+This option is mutually exclusive with `proto-var()`, `server-side-schema()`, `schema()`, and `protobuf-schema()` directives.
+
+example:
+```
+   destination {
+      clickhouse (
+      ...
+      json-var("$json_data");
+or
+      json-var(json("{\"ingest_time\":1755248921000000000, \"body\": \"test template\"}"))
+      ...
+      };
+   };
+```


modules/grpc/clickhouse/clickhouse-dest-worker.cpp
```diff
@@ -73,6 +73,16 @@ DestWorker::insert(LogMessage *msg)
...
+
+    }
+  else if (owner_->json_mode())
+    {
+      ssize_t len;
+      const gchar *json_str = owner_->format_json_var(msg, &len);
+      if (!json_str)
+        goto drop;
+      this->query_data.write(json_str, len);
+      this->query_data.put('\n');
```
MrAnno: We could extract these 3 modes to separate functions to make insert() shorter (next time).

modules/grpc/clickhouse/clickhouse-dest.cpp
```diff
@@ -61,27 +61,29 @@ DestDriver::init()
...
+    if (this->proto_var && !this->log_message_protobuf_formatter.empty())
+      {
+        msg_error("Error initializing ClickHouse destination: 'proto-var()' cannot be used together with 'schema()' "
+                  "or 'protobuf-schema()'. Please use either 'proto-var()', or 'schema()'/'protobuf-schema()', "
+                  "but not both.",
+                  log_pipe_location_tag(&this->super->super.super.super.super));
+        return false;
+      }
+
+    if ((!this->proto_var && this->log_message_protobuf_formatter.empty()))
```
MrAnno: This logic seems to be organized incorrectly.

We have these:
- `protobuf-var()`
- `protobuf-schema()`
- `schema()`
- `json-var()`

Each option is  _mutually exclusive_, but one of them needs to be used.

The current logic checks this in a complicated way, as a result, it misses some scenarios, for example, you can configure json-var() and everything else as well.

We need exactly one out of the listed 4 options to be present.

MrAnno (commented):
Please write a feature news entry.

modules/grpc/clickhouse/clickhouse-dest.cpp
MrAnno: The next patch addresses the missing cases, but the validation code looks too long. Can't this be done more cleanly? For example, counting the used options and saying that only one of them can be used when `!= 1`?

modules/grpc/clickhouse/clickhouse-dest-worker.cpp
```diff
@@ -53,46 +53,68 @@ DestWorker::should_initiate_flush()
...
+
+  google::protobuf::io::OstreamOutputStream zero_copy_output(&this->query_data);
+  google::protobuf::io::CodedOutputStream coded_output(&zero_copy_output);
+  coded_output.WriteVarint32(len);
+  coded_output.WriteRaw(serialized, len);
+  return true;
+}
+
+bool
+DestWorker::create_protobuf_message_from_jsonvar(LogMessage *msg)
```
MrAnno: I think we agreed to call this `insert_query_data_from_protovar/schema/jsonvar`, as this is not creating a message but appends the query data batch.

MrAnno: This looks like a change that can be a separate refactor/extract function commit, no need to squash it.

modules/grpc/clickhouse/clickhouse-dest.cpp
```diff
@@ -44,6 +44,21 @@ DestDriver::DestDriver(GrpcDestDriver *s)
   this->enable_dynamic_headers();
 }
 
+bool
+DestDriver::check_options()
```
MrAnno: check_schema_options?
Or anything that names the category of conflicting options we check.

modules/grpc/clickhouse/clickhouse-dest.cpp
```diff
@@ -44,6 +44,21 @@ DestDriver::DestDriver(GrpcDestDriver *s)
...
+  int count_opts = 0;
+  count_opts += int(this->proto_var != nullptr);
+  count_opts += int(this->json_mode());
+  count_opts += int(!this->log_message_protobuf_formatter.empty() && this->log_message_protobuf_formatter.init());
+
+  if (count_opts != 1)
+    return false;
+  if ((this->proto_var != nullptr) ^ this->server_side_schema.empty() == 0)
+    return false;
+  return true;
```
MrAnno, line 57: Optional: `== 0` and ` > 1` might be worth separating to produce a more understandable error message.

modules/grpc/clickhouse/clickhouse-dest.cpp
```diff
@@ -44,6 +44,21 @@ DestDriver::DestDriver(GrpcDestDriver *s)
...
+DestDriver::check_options()
+{
+  int count_opts = 0;
+  count_opts += int(this->proto_var != nullptr);
+  count_opts += int(this->json_mode());
+  count_opts += int(!this->log_message_protobuf_formatter.empty() && this->log_message_protobuf_formatter.init());
+
+  if (count_opts != 1)
+    return false;
+  if ((this->proto_var != nullptr) ^ this->server_side_schema.empty() == 0)
```
MrAnno: Shouldn't it be `this->proto_var && this->server_side_schema.empty()`?
We don't want an error when non of them is set.

(Feel free to keep a separate error message for this case.)

bshifter: The use of proto_var and server_side_schema is mandatory as a pair: either both must be set or neither should be. The XOR condition enforces this requirement

MrAnno: XOR is 0 for 0^0.

```
if ((this->proto_var != nullptr) ^ this->server_side_schema.empty() == 0)
if(FALSE ^ FALSE == 0)
  return false;
```
You don't want this. That's probably the same reason why tests are failing.

As we discussed it yesterday, the counter-based checks make sure that either proto_var is set, or something else is set, so the only thing left to validate is:
```
if (this->proto_var && this->server_side_schema.empty())
  return false;
```

bshifter: silly me, thanks for noticing!

MrAnno (approved):

PR #772, opened 2025-08-25
lib/filterx/func-str: fix lifetime of expr evaluated strings

OverOrion:
This solves an improper read in `_expr_format`:
```c
  result = _obj_format(obj, str, str_len, ignore_case, expr); // it calls _format_str_obj, which will call filterx_object_extract_string_ref, str will need the backing obj though
  filterx_object_unref(obj); // frees the backing memory too soon here
  return result;
```

TODO: 
- [ ] add missing news file

### DIFF
--- a/lib/filterx/func-str.c
+++ b/lib/filterx/func-str.c
@@ -121,16 +121,15 @@ exit:
 }
 
 static gboolean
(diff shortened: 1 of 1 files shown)


MrAnno (approved):
Thanks.

(probably worth some refactoring in the future)

PR #775, opened 2025-08-26
Filterx location tracking fix

bazsi:
This fixes an ugly location tracking issue in filterx block ref processing.


MrAnno (approved):
🎉

MrAnno:
I tried the new ECHO macro with a huge internal configuration (appdb), no warnings were emitted.

MrAnno:
Hm, but we don't seem to use the ECHO action at all. Is this for debugging?

PR #778, opened 2025-08-27
Cfg lexer source extraction opt

bazsi:
This is an improvement on top of #777 @MrAnno 

I think this improves readability a bit and reduces complexity a lot in cfg-source, which was a mess already.


bazsi:
I rebased against master.

PR #779, opened 2025-08-28
Fix invalid access of freed log-writer cfg

MrAnno:
Cherry-picked from syslog-ng/syslog-ng#5445

This is an excellent catch and fix from @beni-atlnz.

It doesn't cause invalid reads or a crash in case of AxoSyslog (we don't have `cfg_tree_register_initialized_pipe()`), but we shouldn't keep a pointer to a freed configuration anyway.

### DIFF
--- a/modules/afprog/afprog.c
+++ b/modules/afprog/afprog.c
@@ -526,7 +526,6 @@ afprogram_dd_restore_reload_store_item(AFProgramDestDriver *self, GlobalConfig *
     {
       self->process_info.pid = restored_info->pid;
       self->writer = restored_info->writer;
-      ((LogPipe *) self->writer)->cfg = cfg;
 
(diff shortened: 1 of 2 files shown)


modules/afprog/afprog.c
```diff
@@ -526,6 +526,7 @@ afprogram_dd_restore_reload_store_item(AFProgramDestDriver *self, GlobalConfig *
     {
       self->process_info.pid = restored_info->pid;
       self->writer = restored_info->writer;
+      ((LogPipe *) self->writer)->cfg = cfg;
```
MrAnno: We could call `log_pipe_set_config()` in `init()`, right before `log_writer_set_options()`, to make sure we use the appropriate config instance no matter which code path we are coming from (this is what we do in other modules).

bazsi:
Great catch @beni-atlnz thanks for the fix.

PR #780, opened 2025-08-29
filterx: support negative string slicing indexes

MrAnno:
For example,
```
filterx {
  str = "example";
  str[..-2] == "examp";
  str[-3..] == "ple";
};
```


jszigetvari (approved):
I love it! Thanks!
Just curious what happens if: `string[..]`

MrAnno:
That's not a valid syntax.

alltilla (approved):

PR #781, opened 2025-09-02
filterx/metrics-labels: dedup automatically

alltilla:
It is a hassle to call it manually and it is really hard to guarantee that there are no duplicate keys in case of a complex config.

It is probably the best to call it implicitly, when it is needed, with a relatively good performance implementation.


alltilla:
It is better to use a dict in the first place.

PR #783, opened 2025-09-03
light: followup clickhouse changes regarding integer quoting

OverOrion:
A followup is needed because clickhouse no longer quotes 64 bit integer values by default in JSON formats:
https://github.com/ClickHouse/ClickHouse/pull/74079


mitzkia (approved):
Thanks, approved.

PR #789, opened 2025-09-08
fix Clickhouse destination error handling

bshifter:
- Consolidate common code in the gRPC destination to eliminate duplication across its subclasses.
- Introduce ClickHouseErrorCode enum to represent different ClickHouse exception codes.
- Classify ClickHouse errors to determine the appropriate retry strategy.
- Handle ClickHouse log errors using the new error classification system.
- Add patch for stringstream reset

### DIFF
--- a/modules/grpc/clickhouse/clickhouse-exception-codes.c
+++ b/modules/grpc/clickhouse/clickhouse-exception-codes.c
@@ -1,0 +1,137 @@
+/*
+ * Copyright (c) 2024 Axoflow
+ * Copyright (c) 2024 shifter <shifter@axoflow.com>
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "clickhouse-exception-codes.h"
+
+// Retry forever (transient infra issues)
+static const gboolean CLICKHOUSE_ERRORS_RETRY_FOREVER[] =
+{
+  [CANNOT_READ_FROM_SOCKET] = TRUE,
+  [CANNOT_WRITE_TO_SOCKET] = TRUE,
+  [TIMEOUT_EXCEEDED] = TRUE,
+  [TOO_SLOW] = TRUE,
+  [TOO_MANY_SIMULTANEOUS_QUERIES] = TRUE,
+  [NO_FREE_CONNECTION] = TRUE,
+  [SOCKET_TIMEOUT] = TRUE,
+  [NETWORK_ERROR] = TRUE,
+  [QUERY_WITH_SAME_ID_IS_ALREADY_RUNNING] = TRUE,
+  [REPLICA_IS_ALREADY_ACTIVE] = TRUE,
+  [NO_ZOOKEEPER] = TRUE,
+  [UNEXPECTED_ZOOKEEPER_ERROR] = TRUE,
+  [NO_ACTIVE_REPLICAS] = TRUE,
+  [NO_AVAILABLE_REPLICA] = TRUE,
+  [ALL_CONNECTION_TRIES_FAILED] = TRUE,
+  [TOO_FEW_LIVE_REPLICAS] = TRUE,
+  [UNSATISFIED_QUORUM_FOR_PREVIOUS_WRITE] = TRUE,
+  [REPLICA_IS_NOT_IN_QUORUM] = TRUE,
+  [RECEIVED_ERROR_TOO_MANY_REQUESTS] = TRUE,
+  [ALL_REPLICAS_ARE_STALE] = TRUE,
+  [PART_IS_TEMPORARILY_LOCKED] = TRUE,
+  [REPLICA_STATUS_CHANGED] = TRUE,
+  [ALL_REPLICAS_LOST] = TRUE,
+  [NOT_A_LEADER] = TRUE,
+  [DISTRIBUTED_TOO_MANY_PENDING_BYTES] = TRUE,
+  [ZERO_COPY_REPLICATION_ERROR] = TRUE,
+  [CANNOT_CONNECT_NATS] = TRUE,
+  [AWS_ERROR] = TRUE,
+  [GCP_ERROR] = TRUE,
+  [GOOGLE_CLOUD_ERROR] = TRUE,
+  [SERVER_OVERLOADED] = TRUE,
+  [KEEPER_EXCEPTION] = TRUE,
+  [INCORRECT_QUERY] = TRUE,
+  [UNKNOWN_IDENTIFIER] = TRUE,
+  [UNKNOWN_TABLE] = TRUE,
+  [UNKNOWN_DATABASE] = TRUE
+};
+
+// Drop immediately (permanent data/query issues)
+static const gboolean CLICKHOUSE_ERRORS_DROP_IMMEDIATE[] =
+{
+  [UNSUPPORTED_METHOD] = TRUE,
+  [UNSUPPORTED_PARAMETER] = TRUE,
+  [CANNOT_PARSE_ESCAPE_SEQUENCE] = TRUE,
+  [CANNOT_PARSE_QUOTED_STRING] = TRUE,
+  [BAD_ARGUMENTS] = TRUE,
+  [SYNTAX_ERROR] = TRUE,
+  [UNKNOWN_FUNCTION] = TRUE,
+  [CANNOT_PARSE_TEXT] = TRUE,
+  [CANNOT_PARSE_DATE] = TRUE,
+  [CANNOT_PARSE_DATETIME] = TRUE,
+  [CANNOT_PARSE_UUID] = TRUE,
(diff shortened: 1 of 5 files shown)


modules/grpc/bigquery/bigquery-worker.cpp
```diff
@@ -109,22 +116,7 @@ DestinationWorker::disconnect()
...
-
-  status = this->stub->FinalizeWriteStream(&ctx, finalize_request, &finalize_response);
-  if (!status.ok())
-    {
-      msg_warning("Error finalizing BigQuery write stream", evt_tag_str("error", status.error_message().c_str()),
-                  evt_tag_str("details", status.error_details().c_str()),
-                  log_pipe_location_tag((LogPipe *) this->super->super.owner));
-    }
-
-  this->connected = false;
```
MrAnno: Why did you remove this?

modules/grpc/common/grpc-dest-worker.hpp
```diff
@@ -44,20 +44,23 @@ class DestWorker
 
   virtual bool init();
   virtual void deinit();
-  virtual bool connect();
-  virtual void disconnect();
   virtual LogThreadedResult insert(LogMessage *msg) = 0;
   virtual LogThreadedResult flush(LogThreadedFlushMode mode) = 0;
+  virtual bool connect();
+  virtual void disconnect();
```
MrAnno, line 50: These should be protected, but I guess we can't do that because of the C binding: the threaded destination can't be a real subclass of this.

modules/grpc/loki/loki-worker.hpp
```diff
@@ -49,12 +49,14 @@ class DestinationWorker final: public syslogng::grpc::DestWorker
...
-  bool init();
-  bool connect();
-  void disconnect();
   LogThreadedResult insert(LogMessage *msg);
   LogThreadedResult flush(LogThreadedFlushMode mode);
 
+protected:
+  bool init();
+  void deinit();
+  bool connect();
```
MrAnno: Optional: `override`

modules/grpc/common/grpc-dest-worker.cpp
```diff
@@ -74,27 +74,67 @@ DestWorker::create_channel_args()
...
 
 bool
 DestWorker::connect()
 {
+  std::chrono::system_clock::time_point connect_timeout =
+    std::chrono::system_clock::now() + std::chrono::seconds(10);
+
+  if (!this->channel->WaitForConnected(connect_timeout))
+    {
+      msg_error("gRPC connection timed out", log_pipe_location_tag((LogPipe *) this->super->super.owner));
```
MrAnno: This would produce 2 error message every connect failure (+ a third one because of threaded destination).
I would remove remove this and keep the driver-specific error messages.

modules/grpc/common/grpc-dest-worker.cpp
```diff
@@ -74,27 +74,67 @@ DestWorker::create_channel_args()
...
+      return false;
+    }
+
+  this->connected = true;
```
MrAnno, line 125: We can't be sure the connection was established because `DestWorker::connect()` is called from derived classes which may do additional stuff to make the connection complete. If we set this to true too early, the symmetry between init/deinit may break.

I think we can't move this flag to the the gRPC-level, or if we do, we need child classes to set this explicitly.

modules/grpc/clickhouse/clickhouse-exception-codes.h
```diff
@@ -0,0 +1,233 @@
...
+ *
+ */
+
+#include <stddef.h>
+#include "logthrdest/logthrdestdrv.h"
+
+typedef enum
+{
+  ACTION_RETRY_FOREVER,
+  ACTION_RETRY_LIMITED,
```
MrAnno: Optional: `RETRY_THEN_DROP`?

modules/grpc/clickhouse/clickhouse-exception-codes.h
```diff
@@ -0,0 +1,233 @@
...
+  PROTOBUF_BAD_CAST,
+  MULTIPLE_COLUMNS_SERIALIZED_TO_SAME_PROTOBUF_FIELD,
+  DATA_TYPE_INCOMPATIBLE_WITH_PROTOBUF_FIELD,
+  CORRUPTED_DATA,
+  BAD_DATA_PART_NAME,
+  POTENTIALLY_BROKEN_DATA_PART,
+  LOGICAL_ERROR,
+};
+
+#define NUM_CLICKHOUSE_ERRORS_RETRY_FOREVER \
```
MrAnno: Optional: You can use `G_N_ELEMENTS()` instead.

modules/grpc/clickhouse/clickhouse-exception-codes.h
```diff
@@ -0,0 +1,233 @@
...
+    (sizeof(CLICKHOUSE_ERRORS_RETRY_FOREVER) / sizeof(ClickHouseErrorCode))
+
+#define NUM_CLICKHOUSE_ERRORS_DROP_IMMEDIATE \
+    (sizeof(CLICKHOUSE_ERRORS_DROP_IMMEDIATE) / sizeof(ClickHouseErrorCode))
+
+static inline bool is_in_set(ClickHouseErrorCode code,
+                             const ClickHouseErrorCode *set,
+                             size_t set_size)
+{
+  for (size_t i = 0; i < set_size; i++)
```
MrAnno: Optional: you can use a sparse array to imitate a set, that way we won't need to search linearly in expense for some static storage.

For example:
```
static const gboolean CLICKHOUSE_ERRORS_DROP_IMMEDIATE[] =
{
  [CANNOT_PARSE_TEXT] = TRUE,
};
```

modules/grpc/clickhouse/clickhouse-exception-codes.h
```diff
@@ -0,0 +1,233 @@
...
+  AWS_ERROR,
+  GCP_ERROR,
+  GOOGLE_CLOUD_ERROR,
+  SERVER_OVERLOADED,
+  KEEPER_EXCEPTION,
+};
+
+// Drop immediately (permanent data/query issues)
+static const ClickHouseErrorCode CLICKHOUSE_ERRORS_DROP_IMMEDIATE[] =
+{
```
MrAnno: UNSUPPORTED_METHOD, UNSUPPORTED_PARAMETER?

CANNOT_PARSE_ESCAPE_SEQUENCE,  CANNOT_PARSE_QUOTED_STRING, 

INCORRECT_DATA, NO_DATA_TO_INSERT?

modules/grpc/clickhouse/clickhouse-exception-codes.h
```diff
@@ -0,0 +1,233 @@
...
+  UNKNOWN_FUNCTION = 46,
+  UNKNOWN_IDENTIFIER = 47,
+  LOGICAL_ERROR = 49,
+  TYPE_MISMATCH = 53,
+  UNKNOWN_TABLE = 60,
+  SYNTAX_ERROR = 62,
+  CANNOT_PARSE_NUMBER = 72,
+  UNKNOWN_FORMAT = 73,
+  INCORRECT_QUERY = 80,
+  UNKNOWN_DATABASE = 81,
```
MrAnno, line 67: Might be better as retry forever. I don't think we want to lose messages when the table doesn't exist or mistyped. The user can fix the problem, and inject the messages accumulated in disk-buffer.

modules/grpc/clickhouse/clickhouse-exception-codes.h
```diff
@@ -0,0 +1,233 @@
...
+  SIZE_OF_FIXED_STRING_DOESNT_MATCH = 19,
+  NUMBER_OF_COLUMNS_DOESNT_MATCH = 20,
+  BAD_ARGUMENTS = 36,
+  CANNOT_PARSE_DATE = 38,
+  CANNOT_PARSE_DATETIME = 41,
+  UNKNOWN_FUNCTION = 46,
+  UNKNOWN_IDENTIFIER = 47,
+  LOGICAL_ERROR = 49,
+  TYPE_MISMATCH = 53,
+  UNKNOWN_TABLE = 60,
```
MrAnno, line 62: Might be better as retry forever. I don't think we want to lose messages when the table doesn't exist or mistyped. The user can fix the problem, and inject the messages accumulated in disk-buffer.

modules/grpc/common/grpc-dest-worker.cpp
bshifter: In the current implementation, connected reflects the gRPC channel state, which is managed by grpc-dest-worker. Any additional logic specific to child classes should be implemented in their connect method overrides and tracked using the child’s own custom state.

modules/grpc/clickhouse/clickhouse-exception-codes.c
```diff
@@ -0,0 +1,137 @@
...
+  [SERVER_OVERLOADED] = TRUE,
+  [KEEPER_EXCEPTION] = TRUE,
+  [INCORRECT_QUERY] = TRUE,
+  [UNKNOWN_IDENTIFIER] = TRUE,
+  [UNKNOWN_TABLE] = TRUE,
+  [UNKNOWN_DATABASE] = TRUE,
+  [UNKNOWN_FORMAT] = TRUE,
+  [CANNOT_PARSE_NUMBER] = TRUE,
+  [TYPE_MISMATCH] = TRUE,
+  [LOGICAL_ERROR] = TRUE
```
MrAnno: Are we sure that the last 3 can be fixed and it's not about an undeliverable message?

MrAnno (approved):

modules/grpc/clickhouse/clickhouse-exception-codes.c
bshifter: you suggested them to move to retry forever 4 days ago :) it's hard to deceide which is the better pick, i agree. i would say it's better to drop these

MrAnno: I was only talking about UNKNOWN_TABLE and UNKNOWN_DATABASE.
I'm unsure about UNKNOWN_FORMAT CANNOT_PARSE_NUMBER and TYPE_MISMATCH.
If you are unsure too, I think they should not be in the retry forever column.

MrAnno (approved):

PR #793, opened 2025-09-10
Light: Refine checking clickhouse server start/stop

mitzkia:

### DIFF
--- a/tests/light/src/axosyslog_light/helpers/clickhouse/clickhouse_server_executor.py
+++ b/tests/light/src/axosyslog_light/helpers/clickhouse/clickhouse_server_executor.py
@@ -67,11 +67,11 @@ class ClickhouseServerExecutor():
         self.clickhouse_server_ports.append(clickhouse_ports.http_port)
         self.clickhouse_server_ports.append(clickhouse_ports.grpc_port)
 
-        with open("clickhouse_server_config.xml", "r") as file:
+        with open("clickhouse_server_config.xml", "r", encoding="utf-8") as file:
             config = file.read()
         config = config.replace("ALLOCATED_HTTP_PORT", str(clickhouse_ports.http_port))
         config = config.replace("ALLOCATED_INTERSERVER_GRPC_PORT", str(clickhouse_ports.grpc_port))
-        with open("clickhouse_server_config.xml", "w") as file:
+        with open("clickhouse_server_config.xml", "w", encoding="utf-8") as file:
             file.write(config)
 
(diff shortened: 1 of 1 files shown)


tests/light/src/axosyslog_light/helpers/clickhouse/clickhouse_server_executor.py
```diff
@@ -59,24 +63,26 @@ def start(self, clickhouse_ports) -> None:
...
 
-    def get_open_ports(self) -> list[int]:
-        if not self.process:
-            raise RuntimeError("Clickhouse server process is not running.")
+    @retry(wait=wait_fixed(0.1), reraise=True, stop=stop_after_delay(10))
+    def wait_for_server_start(self, port) -> None:
+        client = clickhouse_connect.get_client(
+            host="localhost",
+            username="default",
+            password="password",
```
MrAnno: Can we move `clickhouse_options` from tests into `clickhouse_server_executor.py` and store the default username/password as named constants?

mitzkia: yes, I think it could work. I will check it. Thanks for the comment

mitzkia: fixed

MrAnno: Thanks. I meant can we use those options here instead of hard-coding them?

mitzkia: Sorry, did not understand. Can you give an example?

mitzkia: Oh, I think finally I understand.

mitzkia: In that case, this part is still missing:
```
def wait_for_server_start(self, port) -> None:
        client = clickhouse_connect.get_client(
            host="localhost",
            username=CLICKHOUSE_OPTIONS["user"],
            password=CLICKHOUSE_OPTIONS["password"],
            port=port,
        )
        try:
            client.command("SHOW DATABASES;")
```

MrAnno: Yes, that's what I mean, so that we don't have hard-coded strings in the middle of the code (also avoiding repeating those hard-coded values).
We can add localhost as the default hostname as well.

PR #797, opened 2025-09-18
Add changes to S3 destination from syslog-ng

sodomelle:
This PR takes the commits from: 

- https://github.com/syslog-ng/syslog-ng/pull/5257
- https://github.com/syslog-ng/syslog-ng/pull/5286

Added license change.

### DIFF
--- a/modules/python-modules/syslogng/modules/s3/s3_session_handler.py
+++ b/modules/python-modules/syslogng/modules/s3/s3_session_handler.py
@@ -104,13 +104,19 @@ class S3SessionHandler:
         except ClientError as e:
             if e.response["Error"]["Code"] == "404":
                 self.logger.info(f"Bucket ({self.__bucket}) does not exist, trying to create it")
-            try:
-                self.__client.create_bucket(Bucket=self.__bucket)
-            except (ClientError, EndpointConnectionError) as e:
-                self.logger.error(f"Failed to create bucket ({self.__bucket}): {e}")
-                return False
-            self.logger.info(f"Bucket ({self.__bucket}) successfully created")
-            return True
+                try:
+                    self.__client.create_bucket(Bucket=self.__bucket)
+                except (ClientError, EndpointConnectionError) as e:
+                    self.logger.error(f"Failed to create bucket ({self.__bucket}): {e}")
+                    return False
+                self.logger.info(f"Bucket ({self.__bucket}) successfully created")
+                return True
+            else:
+                self.logger.info(f"Client error: {e}")
+                self.logger.debug(f"Client error debug: {e.response}")
+        except EndpointConnectionError as e:
+            self.logger.info(f"Failed to connect to S3 server: {e}")
+        return False
 
     def upload_object(self, s3_object: S3ObjectBuffer) -> bool:
         if not self.connection_open:


lib/messages.c
```diff
@@ -209,7 +209,8 @@ msg_event_send(EVTREC *e)
 void
 msg_event_suppress_recursions_and_send(EVTREC *e)
 {
-  msg_event_send_with_suppression(e, msg_limit_internal_message);
+  if(G_LIKELY(e != NULL))
+    msg_event_send_with_suppression(e, msg_limit_internal_message);
```
MrAnno: This looks like a workaround

MrAnno:
FYI #799, #805

MrAnno:
We might need syslog-ng/syslog-ng#5447 too.

lib/messages.c
MrAnno: We'll drop it.

modules/python-modules/syslogng/modules/s3/s3_destination.py
```diff
@@ -142,213 +162,191 @@ def __init_options(self, options: Dict[str, Any]) -> None:
...
-        try:
-            self.client.head_bucket(Bucket=self.bucket)
-            is_opened = True
-        except ClientError as e:
-            if e.response["Error"]["Code"] == "404":
-                self.logger.info(f"Bucket ({self.bucket}) does not exist, trying to create it")
-                is_opened = self.__create_bucket()
-            else:
-                self.logger.info(f"Client error: {e}")
-                self.logger.debug(f"Client error debug: {e.response}")
```
MrAnno, line 301: The logging I added seems to be missing.

sodomelle: Similar logging is present in the `ready_bucket` function.
Multipart related logging is not present, as it is implicitly used by `upload_file`.

modules/python-modules/syslogng/modules/s3/s3_session_handler.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def ready_bucket(self):
+        try:
+            self.__client.head_bucket(Bucket=self.__bucket)
+            return True
+        except ClientError as e:
+            if e.response["Error"]["Code"] == "404":
+                self.logger.info(f"Bucket ({self.__bucket}) does not exist, trying to create it")
+            try:
+                self.__client.create_bucket(Bucket=self.__bucket)
```
MrAnno: The bucket should be created only in the 404 case.

modules/python-modules/syslogng/modules/s3/s3_session_handler.py
```diff
@@ -0,0 +1,232 @@
...
+        with self.__connection_lock:
+            return self.__client is not None
+
+    def ready_bucket(self):
+        try:
+            self.__client.head_bucket(Bucket=self.__bucket)
+            return True
+        except ClientError as e:
+            if e.response["Error"]["Code"] == "404":
+                self.logger.info(f"Bucket ({self.__bucket}) does not exist, trying to create it")
```
MrAnno, line 106:
```
else:
                self.logger.info(f"Client error: {e}")
                self.logger.debug(f"Client error debug: {e.response}")
        except EndpointConnectionError as e:
            self.logger.info(f"Failed to connect to S3 server: {e}")
```

this is missing as I mentioned from #805. Please note that this belongs to the head_bucket() query, not to the create_bucket.

Without that, figuring out configuration and permission issues are very difficult.

modules/python-modules/syslogng/modules/s3/s3_session_handler.py
sodomelle: Done.

modules/python-modules/syslogng/modules/s3/s3_session_handler.py
sodomelle: Reintroduced logging enhancements.

PR #815, opened 2025-10-04
Filterx eliminate FilterXList virtual functions

bazsi:
This PR removes the entire FilterXList interface, improving parse_csv() performance by ~3%

```
bazsi@bzorp3:~/src/syslog-ng/syslog-ng$ time for i in `seq 1 200`; do cat dbld/install/pan_dump.txt ; done | nc -q0 localhost 2000

real	0m3.170s


bazsi@bzorp3:~/src/syslog-ng/syslog-ng$ time for i in `seq 1 200`; do cat dbld/install/pan_dump.txt ; done | nc -q0 localhost 2000

real	0m3.072s
```

Instead of creating an entire abstraction layer, this moves all the related logic into a single function, filterx_list_normalize_index() which can easily be invoked from the list implementations (object-otel-array and object-list).


alltilla:
rebased to main, fixed the conflict

alltilla (approved):

PR #856, opened 2025-11-24
Filterx: lighter type checks

bazsi:
This PR removes some of the unneeded type checks from filterx-mapping and filterx-sequence. These were
preparatory steps for some of my performance improvements, but can be merged separately.

Also, it has a refactor on filterx_mapping_keys() which had a very arcane behaviour:
* it freed its first argument (e.g. self)
* it expected the caller to construct the list object where the results are returned

Both of these were changed to make filterx_mapping_keys() to be more consistent with the rest of the filterx implementation.

### DIFF
--- a/lib/filterx/expr-membership.c
+++ b/lib/filterx/expr-membership.c
@@ -63,16 +63,7 @@ _eval_in(FilterXExpr *s)
 
   FilterXObject *list_obj = filterx_ref_unwrap_ro(rhs_obj);
 
-  if (!filterx_object_is_type(list_obj, &FILTERX_TYPE_NAME(sequence)))
-    {
-      filterx_eval_push_error_info_printf("Failed to evaluate 'in' operator", &self->super.super,
-                                          "Right hand side must be list type, got: %s",
-                                          filterx_object_get_type_name(list_obj));
-      goto error;
-    }
-
   guint64 size;
-
   if (!filterx_object_len(list_obj, &size))
     {
       filterx_eval_push_error_static_info("Failed to evaluate 'in' operator", s, "len() method failed");
--- a/lib/filterx/filterx-mapping.c
+++ b/lib/filterx/filterx-mapping.c
@@ -44,8 +44,6 @@ _add_elem(FilterXObject *key_obj, FilterXObject *value_obj, gpointer user_data)
 gboolean
 filterx_mapping_merge(FilterXObject *s, FilterXObject *other)
 {
-  other = filterx_ref_unwrap_ro(other);
-  g_assert(filterx_object_is_type(other, &FILTERX_TYPE_NAME(mapping)));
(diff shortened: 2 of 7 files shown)


lib/filterx/filterx-mapping.c
```diff
@@ -44,8 +44,6 @@ _add_elem(FilterXObject *key_obj, FilterXObject *value_obj, gpointer user_data)
 gboolean
 filterx_mapping_merge(FilterXObject *s, FilterXObject *other)
 {
-  other = filterx_ref_unwrap_ro(other);
-  g_assert(filterx_object_is_type(other, &FILTERX_TYPE_NAME(mapping)));
```
alltilla, line 48: I don't understand why these calls are unnecessary.

`filterx_object_iter()` only handles the error case where the given object does not have an `iter()` function.
If we were meant to operate on two `mapping` typed objects, but the `other` is a `sequence`, then `_sequence_iter()` will be executed here, which is a problem. Or are we okay with doing that?

Sure, currently all the call sites check for `other`'s type before calling `filterx_mapping_merge()`, but we can easily miss that, and `assert`s are supposed to save us from coding errors like those.

Am I missing something? Could you kindly help me understand?

lib/filterx/expr-membership.c
```diff
@@ -63,16 +63,7 @@ _eval_in(FilterXExpr *s)
...
   FilterXObject *list_obj = filterx_ref_unwrap_ro(rhs_obj);
 
-  if (!filterx_object_is_type(list_obj, &FILTERX_TYPE_NAME(sequence)))
-    {
-      filterx_eval_push_error_info_printf("Failed to evaluate 'in' operator", &self->super.super,
-                                          "Right hand side must be list type, got: %s",
-                                          filterx_object_get_type_name(list_obj));
-      goto error;
-    }
-
```
alltilla, line 73: If `rhs` is a variable, what guarantees, that it is a `sequence`?

lib/filterx/filterx-mapping.c
bazsi: if we pass a list, we will iterate on its elements and fail on set_subscript later. But if we had any other object that produces key/value pairs where "key" is a string, it will be merged just the same.

I was removing these checks, as we don't have to validate the type anymore just to see if we have an iter function. Checking the type explicitly will prevent interesting use-cases where we have a new type that behaves like a dict, so we can merge its contents too (for instance metrics_labels)

but this patch is up for debate. I can be convinced otherwise,

lib/filterx/expr-membership.c
bazsi: nothing? still we can iterate it, because it has an iter function, again the interface allows us to use the "in" operator on any sequence like objects not just lists.

PR #925, opened 2026-01-23
Filterx non atomic refcount

bazsi:
Another attempt at removing the atomic ref counter from FilterXObjects for a cca 15% performance improvement (single thread), but which can also improve scalability to more CPUs).

It builds on all of my pending PRs as a whole (over 100 patches), so merging this would only happen once the base is smaller.

The core idea:
* our thread specific allocator would free associated memory with FilterXObjects anyway, so using an atomic ref counter does not help much in preventing parallel use of objects from multiple threads
* I added a validator code, that checks if somehow an early, compile time allocation leaks into normal evaluation without being frozen or hibernated.
* I added the ability to freeze mutable objects
* removed readonly support (as we have freezing now)
* it is now possible to do copy-on-write on cache_json_file() objects, so no need to keep it readonly


lib/filterx/tests/test_object_dict.c
```diff
@@ -209,9 +209,7 @@ Test(filterx_dict, test_dict_dedup)
   FilterXObject *dict = filterx_object_from_json("{\"a\": \"a\", \"b\": \"b\", \"c\": \"a\"}", -1, NULL);
   FilterXObject *orig_dict = dict;
 
-  GHashTable *dedup_store = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, NULL);
-  filterx_object_dedup(&dict, dedup_store);
-  g_hash_table_unref(dedup_store);
+  filterx_config_freeze_object(configuration, &dict);
```
MrAnno, line 212: This seems to be undoing a lot of what has been done in #662 to make the design simpler (and to fix some crashes):

> - recursive/mutable types can't be frozen or hibernated anymore
> - the virtual freeze/unfreeze calls have been removed
> - deduplication of immutable values is now done explicitly, without hiding it inside the freeze() implementation  (this is what actually fixes the crashes I mentioned).

My biggest concern here is the deduplication part: it is now hidden and coupled with freeze() again. `FilterXEnvironment` couples them correctly this time (at least for now), so the crash hasn't been reintroduced, but I think it just reintroduces that same complication that caused the misunderstanding and the crash at the first place.

Deduplication and freezing should not imply each other in my opinion, we should let "users" deduplicate their objects regardless of whether they freeze them or not. I understand we don't need such functionality at the moment, but it logically makes sense to allow this and it would keep everything easier to understand and avoid introducing similar to the above lifetime bugs/crashes.

bazsi: I've split the freeze functionality into dedup and freeze, separately. I don't like the duplication of the iteration logic (e.g. both freeze and dedup iterating over the entire object hierarchy), but my simple attempt at merging those two iterations on all child objects failed and I did not want to spend much more time on that.

bazsi:
This should now be ready for one final review cycle, @MrAnno

bazsi:
I found a potential bug while working in a parallel thread. This change makes the stashed JSON object returned by cache_json_file() "preserved". And while an object is preserved, filterx_eval_retain_object() will not dup it:

```
static inline void
filterx_eval_retain_object(FilterXObject **pobject)
{
  if (!object || filterx_object_is_preserved(object) || !object->allocator_used)
    return;
...
}
```

This means that as soon as the JSON is freed (for instance because the JSON file is reloaded), retained copies may be freed too, as retain_object will not actually duplicate these. This may not be an acual bug, but is worth a check.

bazsi:
> I found a potential bug while working in a parallel thread. This change makes the stashed JSON object returned by cache_json_file() "preserved". And while an object is preserved, filterx_eval_retain_object() will not dup it:
> 
> ```
> static inline void
> filterx_eval_retain_object(FilterXObject **pobject)
> {
>   if (!object || filterx_object_is_preserved(object) || !object->allocator_used)
>     return;
> ...
> }
> ```
> 
> This means that as soon as the JSON is freed (for instance because the JSON file is reloaded), retained copies may be freed too, as retain_object will not actually duplicate these. This may not be an acual bug, but is worth a check.

This is not an issue here, but will be in the aggregate() branch, in which I am keeping around FilterXObject instances for longer than the current input thread. So this is not an issue here.

bazsi:
Another issue that I identified and which I have now fixed as well, was that we need to keep a reference to FilterXStashedObject until the very end of the current evaluation context. I fixed that by adding a weakref to a dummy FilterXObject that keeps a ref to FilterXStashedObject until it is freed.

I also pushed a refactor to the StashedObject API, which I think makes it easer to use and read on the caller side.

lib/filterx/filterx-stashed-object.c
```diff
@@ -0,0 +1,136 @@
...
+{
+  if (!stash || !*stash)
+    return NULL;
+  /* load-acquire the atomic stash variable */
+
+  /* g_atomic_pointer_get(): load */
+  FilterXStashedObject *stashed_object = g_atomic_pointer_get(stash);
+
+  /* filterx_stashed_object_ref(): acquire */
+  FilterXObject *stashed_reference = filterx_stash_reference_new(filterx_stashed_object_ref(stashed_object));
```
alltilla: I think there is a race between these two operations:
1. We get a `stashed_object` here.
2. In another thread `filterx_stash_store()` pulls from the `stash` the same `stashed_object` that is held here, and `unref()`s it, potentially freeing it.
3. `filterx_stashed_object_ref()` is called on a freed memory.

lib/filterx/filterx-object.h
```diff
@@ -291,13 +289,27 @@ filterx_free_object(FilterXObject *object)
...
+{
+#if SYSLOG_NG_ENABLE_DEBUG
+  if (G_UNLIKELY(!self->early_allocation_checked))
+    {
+      _filterx_object_assert_object_is_not_shared(self);
+      self->early_allocation_checked = TRUE;
+    }
+#endif
+}
+
```
alltilla, line 309: I think this misses one failure scenario:
1. Create an object in compile time -> `early_allocation` is `TRUE`.
2. Do not freeze it.
3. Call some operation that `ref/unref`s it still during compile.
4. The `ref/unref` will run this check, which runs no-op as we are not in production.
5. `filterx_object_check_early_allocation()` flips `early_allocation_checked` to `TRUE`.
6. Put the object somewhere available, but forget to freeze it.
7. In eval something `ref/unref`s it.
8. The check should run and catch it, but it won't do as in step 5 we marked it checked.

lib/filterx/filterx-stashed-object.c
```diff
@@ -0,0 +1,136 @@
...
+  filterx_eval_store_weak_ref(stashed_reference);
+  filterx_object_unref(stashed_reference);
+
+  /* this filterx_object_ref() is normally a noop as the object itself is
+   * frozen, but is here as a matter of correctness. */
+  return filterx_object_ref(stashed_object->object);
+}
+
+gboolean
+filterx_stash_store(FilterXStashedObject **stash, FilterXObject *object, FilterXEnvironment *env)
```
alltilla, line 136: [optional]

Can we add a comment that this steals the ref?

lib/filterx/object-string.c
```diff
@@ -419,6 +415,14 @@ _filterx_string_new_slice_from_non_string(FilterXObject *object, gsize start, gs
...
 }
 
+FilterXObject *
+filterx_string_new_frozen(const gchar *str)
+{
+  FilterXObject *self = filterx_string_new(str, -1);
+  filterx_eval_freeze_object(&self);
+  return self;
+}
+
```
alltilla, line 434: [optional]

Can we add a comment to only use this in compile time? Or an assert?

lib/filterx/filterx-stashed-object.c
bazsi: This is a load-acquire and store-release pattern which ensures that it is race free. The important part is that due to ordering this works without locking.

bazsi: @MrAnno I have second thoughts here, when I did my research last time, I saw that the same technique was used previously (with the FilterXObject refcount), but I now tried to redo my chats with  Claude and ChatGPT and they both are now in disagreement if this pattern is safe.

bazsi: I think this is an issue on current main as well. I am trying to create a quick poor man's RCU to solve this.

bazsi: I've now added a real mutex to protect this critical section.

lib/filterx/filterx-object.h
bazsi: fixed this in the latest iteration. During initialization time, we will not cancel out the check, only during
production.

lib/filterx/object-string.c
bazsi: I added one to filterx_eval_freeze_object() to check if the context is a production context or a compile-time one.

alltilla (dismissed):

lib/filterx/filterx-stashed-object.c
alltilla: Thanks!
Approve from my side. I would wait for @MrAnno 's opinion about this before merging.

MrAnno: This is incorrect on main. Shifi's code was race-free, I introduced a possible use-after-free race. Nice catch.

Bazsi told me he'll add an RWlock (a regular mutex could block parallel readers) instead, that should be enough for now.

bazsi: I retained the use of a simple mutex, end-to-end tests did not show improvements, was slightly slower but that can be just noise. Claude seems to agree that using rwlocks will be a bit slower.

MrAnno: If we can't measure the difference, an RWLock promises not to lock parallel reads, so I think that is the safe choice.

lib/filterx/filterx-stashed-object.c
```diff
@@ -0,0 +1,147 @@
...
+  FilterXStashedObject *self = g_new0(FilterXStashedObject, 1);
+  g_atomic_counter_set(&self->ref_cnt, 1);
+
+  filterx_env_move(&self->env, env);
+  self->object = object;
+  filterx_env_freeze_object(&self->env, &self->object);
+  return self;
+}
+
+static GRWLock stash_rwl;
```
MrAnno, line 105: We should have a separate lock for each protected object, so it should be somewhere in cache-json-file, but preferable inside the stashed object API. I'll add a comment in a separate PR not to forget about this..

MrAnno (approved):

PR #994, opened 2026-03-12
breaking(chart): overhaul

csatib02:
## Summary

- Standardize labels/selectors: drop legacy app/chart/heritage/release labels, adopt `app.kubernetes.io/*` standard, introduce component-aware selector helpers for both components.
- Parity between collector ds and aggregator sts fields were met.   
- Fix default template usage: correct argument order (A | default B → default A B) so component-level values properly override globals.
- Fix `metricsExporter.image.pullPolicy`: no longer emits `imagePullPolicy: ""` when set to `~`
- Disable `rfc5424-tls port` when no TLS is configured: `default-network-drivers()` was opening port 6514 without a cert/key pair, causing a startup warning.
- Bootstrap helm-docs: add `README.md.gotmpl`, annotate values, add CI check to ensure docs are regenerated

## Breaking changes

- Values key syslog: renamed to aggregator, existing overrides must be updated
- Resource names change from *-syslog to *-aggregator, existing sts definitions must be deleted before upgrade (selectors are immutable).
- Selector labels changed from app/release to app.kubernetes.io/name+instance+component, same constraint applies to the ds resource too.

Manually tested:

```sh
# Default render
helm template test ./charts/axosyslog
```

```sh
# Global resources fallback
helm template test ./charts/axosyslog \
--set 'resources.requests.memory=256Mi,resources.limits.memory=512Mi' \
| grep -A6 resources
```

```sh
# Component-level resources override
helm template test ./charts/axosyslog \
--set 'resources.requests.memory=256Mi' \
--set 'syslog.resources.requests.memory=512Mi,syslog.resources.limits.memory=1Gi' \
| grep -A6 resources
```

### DIFF
--- a/charts/axosyslog/templates/aggregator-config.yaml
+++ b/charts/axosyslog/templates/aggregator-config.yaml
@@ -3,10 +3,10 @@ kind: ConfigMap
 metadata:
   labels:
     {{- include "axosyslog.labels" . | nindent 4 }}
-  name: {{ include "axosyslog.fullname" . }}-syslog
+  name: {{ include "axosyslog.fullname" . }}-aggregator
 data:
-{{- if .Values.syslog.config.raw }}
-  syslog-ng.conf: {{ tpl (toYaml .Values.syslog.config.raw) . | indent 2 }}
+{{- if .Values.aggregator.config.raw }}
+  syslog-ng.conf: {{ tpl (toYaml .Values.aggregator.config.raw) . | indent 2 }}
 {{- else }}
   syslog-ng.conf: |
     @version: {{ regexFind "^[0-9]+\\.[0-9]+" .Chart.AppVersion | default "current" }}
@@ -14,13 +14,13 @@ data:
 
     options {
       stats(
-        level({{ .Values.syslog.config.stats.level }})
+        level({{ .Values.aggregator.config.stats.level }})
       );
       ts-format(iso);
       frac-digits(6);
     };
 
-{{- with .Values.syslog.config.sources.syslog }}
+{{- with .Values.aggregator.config.sources.syslog }}
   {{- if .enabled }}
     source s_syslog {
       default-network-drivers(
@@ -56,7 +56,7 @@ data:
   {{- end }}
 {{- end }}
 
-{{- with .Values.syslog.config.sources.axosyslogOtlp }}
+{{- with .Values.aggregator.config.sources.axosyslogOtlp }}
   {{- if .enabled }}
     source s_syslogng_otlp {
       axosyslog-otlp();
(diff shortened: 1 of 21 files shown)


charts/axosyslog/templates/syslog-statefulset.yaml
```diff
@@ -20,36 +20,68 @@ metadata:
...
     spec:
+      {{- if .Values.serviceAccount.create }}
+      serviceAccountName: {{ template "axosyslog.serviceAccountName" . }}
+      {{- end }}
       {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-
+      tolerations: {{ toYaml ( .Values.tolerations | default .Values.syslog.tolerations ) | nindent 8 }}
```
orymate: shouldn't these all be the other way around?

```suggestion
      tolerations: {{ toYaml ( .Values.syslog.tolerations | default .Values.tolerations ) | nindent 8 }}
```

csatib02: Yes, but on the collector side, the pattern was like this way.

we can change the order on both sides, but that would count as breaking.

csatib02: Included a lot of breaking changes to make this happen :P

charts/axosyslog/values.yaml
```diff
@@ -88,14 +88,14 @@ collector:
   # -- Use the host network namespace for collector pods
   hostNetworking: false
 
-# -- Syslog StatefulSet: receives incoming syslog/OTLP traffic (including from the collector) and routes it to destinations.
+# -- Aggregator StatefulSet: receives incoming syslog/OTLP traffic (including from the collector) and routes it to destinations.
 # @default -- See values.yaml
-syslog:
+aggregator:
```
MrAnno, line 93: I'm not sure this name is meaningful enough in our context,  but it's definitely better than `syslog`.

charts/axosyslog/templates/aggregator-config.yaml
```diff
@@ -26,7 +26,7 @@ data:
       default-network-drivers(
         udp-port(1514)
         tcp-port(1514)
-        rfc5424-tls-port(6514)
+        rfc5424-tls-port({{ if .tls }}6514{{ else }}0{{ end }})
```
MrAnno: How did you find out about this?

MrAnno (commented):
Thanks for sorting out this mess, it's much better now.

This `helm-docs` stuff is pretty good, finally we can stop the double bookkeeping.

charts/axosyslog/templates/aggregator-config.yaml
csatib02: If you deploy the chart from main, without setting anything, the syslog component will report a warning about this on startup.

charts/axosyslog/templates/syslog-statefulset.yaml
orymate: shouldn't we maybe template replicas count too? it looks like this should work just fine with multiple replicas

charts/axosyslog/templates/aggregator-config.yaml
MrAnno: I mean the 0 value. It disables network listeners, but I don't think it was ever documented (or at least I never knew :))

csatib02: Should we change this to something?

MrAnno: It works correctly

orymate (approved):

MrAnno:
PRs are piling up around the Helm chart.

I'm merging this and start reviewing the other ones. The chart is released separately from AxoSyslog, so we don't have any time pressure on fixing minor issues (if we find any).

Thanks for the overhaul!

PR #1001, opened 2026-03-17
add source message to http batching errors

bazsi:
This patch adds the _original_ message details to the internal HTTP batching error messages. This should dramatically improve the time needed to diagnose rejected batches by stuff like Splunk.


modules/http-adapters/splunk-adapter.h
```diff
@@ -0,0 +1,30 @@
...
+ * COPYING for details.
+ *
+ */
+
+#ifndef SPLUNK_ADAPTER_H_INCLUDED
+#define SPLUNK_ADAPTER_H_INCLUDED 1
+
+#include "http-adapter.h"
+
+HttpAdapter *splunk_adapter_new(void);
```
OverOrion, line 28: Type mismatch here:
```suggestion
LogDriverPlugin *splunk_adapter_new(void);
```

modules/http-adapters/http-adapter.c
```diff
@@ -0,0 +1,68 @@
...
+  HttpAdapter *self = (HttpAdapter *) s;
+  SignalSlotConnector *ssc = driver->signal_slot_connector;
+
+  DISCONNECT(ssc, signal_http_response, _on_http_response_received, self);
+}
+
+static void
+_free(LogDriverPlugin *s)
+{
+//  HttpAdapter *self = (HttpAdapter *) s;
```
OverOrion: [nit]
Commented out code

modules/http/http-worker.c
```diff
@@ -91,14 +92,64 @@ _curl_debug_function(CURL *handle, curl_infotype type,
...
+
+  GString *message_details = scratch_buffers_alloc();
+
+  g_string_append_c(message_details, '@');
+  append_format_unix_time(&msg->timestamps[LM_TS_RECVD], message_details, TS_FMT_UNIX, -1, 6);
+  g_string_append_c(message_details, ' ');
+
+  gchar source[MAX_SOCKADDR_STRING], dest[MAX_SOCKADDR_STRING];
+  g_sockaddr_format(msg->saddr, source, sizeof(source), GSA_ADDRESS_PORT);
+  g_sockaddr_format(msg->saddr, dest, sizeof(dest), GSA_ADDRESS_PORT);
```
OverOrion: typo
```suggestion
  g_sockaddr_format(msg->daddr, dest, sizeof(dest), GSA_ADDRESS_PORT);

```

Also these fields could be empty, so it crashes with example-msg-generator() at the moment.

OverOrion (commented):
Overall looks great, I am giving it a test run! 🚀 

I had to apply the following patch to compile with cmake:
```diff
diff --git a/modules/CMakeLists.txt b/modules/CMakeLists.txt
index c87d2bb42..1c610a73a 100644
--- a/modules/CMakeLists.txt
+++ b/modules/CMakeLists.txt
@@ -26,6 +26,7 @@ add_subdirectory(getent)
 add_subdirectory(graphite)
 add_subdirectory(hook-commands)
 add_subdirectory(http)
+add_subdirectory(http-adapters)
 add_subdirectory(java)
 add_subdirectory(java-modules)
 add_subdirectory(json)
diff --git a/modules/http-adapters/CMakeLists.txt b/modules/http-adapters/CMakeLists.txt
index c33a50efa..7ae2349b7 100644
--- a/modules/http-adapters/CMakeLists.txt
+++ b/modules/http-adapters/CMakeLists.txt
@@ -18,6 +18,8 @@ set(HTTP_ADAPTERS_SOURCES
 add_module(
   TARGET http-adapters
   GRAMMAR http-adapters-grammar
+  INCLUDES ${PROJECT_SOURCE_DIR}
+  DEPENDS http
   SOURCES ${HTTP_ADAPTERS_SOURCES}
 )

diff --git a/modules/http-adapters/tests/CMakeLists.txt b/modules/http-adapters/tests/CMakeLists.txt
new file mode 100644
index 000000000..f996acfbd
--- /dev/null
+++ b/modules/http-adapters/tests/CMakeLists.txt
@@ -0,0 +1 @@
+add_unit_test(CRITERION TARGET test_splunk_adapter INCLUDES ${PROJECT_SOURCE_DIR} DEPENDS http-adapters http)
```

modules/http-adapters/http-adapter.c
bazsi: thanks.

modules/http-adapters/splunk-adapter.h
bazsi: I retained the HttpAdapter return value and changed the .c file instead.

modules/http/http-worker.c
bazsi: thanks for catching this. I've now added handling of NULLs

modules/http/http-worker.c
```diff
@@ -91,14 +92,71 @@ _curl_debug_function(CURL *handle, curl_infotype type,
...
+  gchar source[MAX_SOCKADDR_STRING], dest[MAX_SOCKADDR_STRING];
+  if (msg->saddr)
+    g_sockaddr_format(msg->saddr, source, sizeof(source), GSA_ADDRESS_PORT);
+  else
+    g_strlcpy(source, "0.0.0.0:0", sizeof(source));
+
+  if (msg->daddr)
+    g_sockaddr_format(msg->daddr, dest, sizeof(dest), GSA_ADDRESS_PORT);
+  else
+    g_strlcpy(source, "0.0.0.0:0", sizeof(source));
```
OverOrion, line 135: [nit]
```suggestion
    g_strlcpy(source, "0.0.0.0:0", sizeof(dest));
```

modules/http-adapters/http-adapters-parser.c
```diff
@@ -0,0 +1,48 @@
...
+#if SYSLOG_NG_ENABLE_DEBUG
+  .debug_flag = &http_adapters_debug,
+#endif
+  .name = "http-adapters",
+  .keywords = http_adapters_keywords,
+  .parse = (gint (*)(CfgLexer *, gpointer *, gpointer)) http_adapters_parse,
+  .cleanup = (void (*)(gpointer)) log_pipe_unref,
+};
+
+CFG_PARSER_IMPLEMENT_LEXER_BINDING(http_adapters_, HTTP_, LogDriver **)
```
OverOrion, line 48:
```suggestion
CFG_PARSER_IMPLEMENT_LEXER_BINDING(http_adapters_, HTTPADAPTERS_, LogDriver **)
```

OverOrion (commented):
Looks good to me, two minor notes to address!

lib/logqueue.h
```diff
@@ -87,7 +87,7 @@ struct _LogQueue
   void (*push_tail)(LogQueue *self, LogMessage *msg, const LogPathOptions *path_options);
   LogMessage *(*pop_head)(LogQueue *self, LogPathOptions *path_options);
   LogMessage *(*peek_head)(LogQueue *self);
-  void (*ack_backlog)(LogQueue *self, gint n);
+  void (*ack_backlog)(LogQueue *self, guint n);
```
MrAnno, line 90: DiskQ used `-1` for some rewind_all/ack mechanism, but I can't find it anymore.

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -968,6 +992,7 @@ _set_logqueue_virtual_functions(LogQueue *s)
 {
   s->get_length = _get_length;
   s->ack_backlog = _ack_backlog;
+  s->peek_backlog = _peek_backlog;
```
MrAnno, line 995: The reliable disk-buffer is missing this method (Claude found it).

modules/http/http-worker.c
```diff
@@ -712,7 +712,9 @@ _flush_on_target(HTTPDestinationWorker *self, const gchar *url)
   HttpResponseReceivedSignalData signal_data =
   {
     .result = HTTP_SLOT_SUCCESS,
-    .http_code = http_code
+    .http_code = http_code,
+    .response_body = self->response_buffer,
+    .offending_message = 0,
```
MrAnno, line 717: Do we want to show the first message when no adapter is available?

modules/http/http-worker.c
```diff
@@ -91,14 +92,71 @@ _curl_debug_function(CURL *handle, curl_infotype type,
...
 
-#define HTTP_RESPONSE_MAX_LENGTH 1024
-
 static inline EVTTAG *
 _tag_request(HTTPDestinationWorker *self)
 {
-  return evt_tag_printf("request", "%.512s%s", self->request_body->str, self->request_body->len > 512 ? "..." : "");
+  if (self->response_signal.offending_request_len)
+    return evt_tag_mem("request-slice",
+                       &self->request_body->str[self->response_signal.offending_request_start],
```
MrAnno, line 100: These are coming from separate plugins, I guess we'll have a few more of these implementations in the future, which may accidentally trust input coming from outside of axosyslog.

Can we add a boundary check for `offending_request_start` or `offending_request_len` to avoid future vulnerabilities if someone implements an adapter unsafely?

modules/diskq/logqueue-disk-non-reliable.c
bazsi: I know, but in that case we will simply fail the peek_backlog() call and will not be able to display the offending message.

With the reliable queue, getting the message out is not trivial as we may need to go to disk and deserialize messages one after the other. The solution I chose was to say in the log that the offending message is not available. See https://github.com/axoflow/axosyslog/blob/e1524002cc48dc08a0c95621fe849525204d609e/modules/http/http-worker.c#L153-L156

modules/http/http-worker.c
bazsi: I would, yes. That's what we display currently, and I think it may give us some context even in that case.

modules/http/http-worker.c
bazsi: good idea.

bazsi:
rebased, fixed most comments, except two which are still open, but I gave answers to both of them. @MrAnno please resolve them if my answers are satisfactory.

modules/diskq/logqueue-disk-non-reliable.c
MrAnno: Okay, thank you.

MrAnno (approved):

PR #1016, opened 2026-03-28
Autoconf monolithic build

bazsi:
This adds a new linking mode "monolithic" that links all modules into the main syslog-ng binary. It also removes the old and now broken "mixed" mode, which cannot be fixed (Debian does not support static glib anymore).

The reason: the number of cycles required to run our callgrind tests go down

without the patch: `I   refs:      2,158,813`

with the patch: `I   refs:      1,944,897`


docker/apkbuild/axoflow/axosyslog/APKBUILD
```diff
@@ -143,34 +106,6 @@ dev() {
 		usr/share/syslog-ng/xsd
 }
 
-_python3() {
-	pkgdesc="$pkgdesc (python3 module)"
-	depends="$pkgname=$pkgver-r$pkgrel python3"
-	install="$subpkgname.post-install"
```
MrAnno, line 150: The post-install script should probably go to `package()`.

docker/apkbuild/axoflow/axosyslog/APKBUILD
```diff
@@ -125,14 +96,6 @@ package() {
 	install -d -m 700 "$pkgdir"/var/lib/syslog-ng
 }
 
-scl() {
-	pkgdesc="$pkgdesc (configuration library)"
-	depends="$pkgname=$pkgver-r$pkgrel"
-
-	_submv usr/share/syslog-ng/include/scl
-	_submv usr/share/syslog-ng/include/scl.conf
```
MrAnno, line 134: This should be fine to delete, but for some reason, the smoke test can't find scl.conf anymore.

docker/apkbuild/axoflow/axosyslog/APKBUILD
bazsi: good catch, thank you

docker/apkbuild/axoflow/axosyslog/APKBUILD
bazsi: I did not fight it, but restored the scl package instead.

bazsi:
As discussed, this is always going to fail the callgrind CI job until we merge this once, simply because the base does not have the `--with-linking-mode=monolithic` configure option. The fact that callgrind runs on the PR and does not run on the base is to be expected.

@MrAnno can you approve/merge this with that constraint?

bazsi:
I changed my mind and implemented that callgrind(BASE) failures are ignored. This would come up any time we are introducing incompatibilities in the autoconf scripts.

MrAnno:
Do I understand correctly that this PR keeps `libsyslog-ng` as a dynamic library?

A dynamic core lib is necessary for the JIT compiler I'm working on, otherwise we'd have to generate a lot of boilerplate for each filterx symbol (I'm using `llvm::orc::DynamicLibrarySearchGenerator`).

configure.ac
MrAnno: This path is untouched, which means we override SYSLOGNG_LINK to use CCLD in case of the new monolithic build. But we need the C++ linker in the monolithic case (and also when the sanitizer is enabled).

lib/plugin.c
```diff
@@ -529,15 +529,24 @@ plugin_extract_candidate_plugins(PluginContext *context, const gchar *module_nam
  * The called function can call plugin_extract_candidate_plugins() to
  * register any module_info struct that is built in.
  */
+#if SYSLOG_NG_ENABLE_BUILTIN_MODULES
```
MrAnno, line 532: `__attribute__((weak))` is no longer neded this way.

configure.ac
MrAnno: In a later patch: now it's correct for the monolithic build, but the sanitizer won't work in `mixed` mode. That's probably fine.

configure.ac
```diff
@@ -2208,15 +2208,22 @@ if test "x$linking_mode" = "xdynamic"; then
...
 	SYSLOGNG_DEPS_LIBS="$LIBS $BASE_LIBS $RESOLV_LIBS $EVTLOG_NO_LIBTOOL_LIBS $SECRETSTORAGE_NO_LIBTOOL_LIBS $LD_START_STATIC -Wl,${WHOLE_ARCHIVE_OPT} $GLIB_LIBS $PCRE2_LIBS $OPENSSL_LIBS $REGEX_LIBS $JSON_LIBS $LIBUNWIND_LIBS -Wl,${NO_WHOLE_ARCHIVE_OPT} $IVYKIS_NO_LIBTOOL_LIBS $LD_END_STATIC $LIBCAP_LIBS $DL_LIBS"
 	TOOL_DEPS_LIBS="$LIBS $BASE_LIBS $GLIB_LIBS $EVTLOG_LIBS $SECRETSTORAGE_LIBS $RESOLV_LIBS $LIBCAP_LIBS $PCRE2_LIBS $REGEX_LIBS $IVYKIS_LIBS $DL_LIBS $OPENSSL_LIBS $JSON_LIBS"
 	CORE_DEPS_LIBS=""
 
 	# bypass libtool in case we want to do mixed linking because it
 	# doesn't support -Wl,-Bstatic -Wl,-Bdynamic on a per-library basis.
 	SYSLOGNG_LINK='$(CCLD) $(AM_CFLAGS) $(CFLAGS) $(AM_LDFLAGS) $(LDFLAGS) -o $@'
+elif test "x$linking_mode" = "xmonolithic"; then
+	enable_shared=no
+	enable_static=yes
```
MrAnno: I don't understand how this works.

LT_INIT is called with disable-static:
`LT_INIT([dlopen disable-static])`

Setting `enable_static=yes` have effects on libtool if we set it later?

It seems to be working, but we install all the static libraries:
```
ls -al lib/syslog-ng/ 
total 138408
drwxrwxr-x 5 redacted redacted     4096 ápr   28 14:54 ./
drwxrwxr-x 4 redacted redacted     4096 ápr   28 14:53 ../
-rw-r--r-- 1 redacted redacted   443308 ápr   28 14:53 libadd-contextual-data.a
-rwxr-xr-x 1 redacted redacted     1132 ápr   28 14:53 libadd-contextual-data.la*
-rw-r--r-- 1 redacted redacted   364610 ápr   28 14:53 libafamqp.a
-rwxr-xr-x 1 redacted redacted     1104 ápr   28 14:53 libafamqp.la*
-rw-r--r-- 1 redacted redacted  1420740 ápr   28 14:53 libaffile.a
-rwxr-xr-x 1 redacted redacted     1182 ápr   28 14:53 libaffile.la*
-rw-r--r-- 1 redacted redacted   392268 ápr   28 14:53 libafmongodb.a
-rwxr-xr-x 1 redacted redacted     1131 ápr   28 14:53 libafmongodb.la*
-rw-r--r-- 1 redacted redacted   359704 ápr   28 14:53 libafprog.a
-rwxr-xr-x 1 redacted redacted     1093 ápr   28 14:53 libafprog.la*
-rw-r--r-- 1 redacted redacted   296580 ápr   28 14:53 libafsmtp.a
-rwxr-xr-x 1 redacted redacted     1101 ápr   28 14:53 libafsmtp.la*
-rw-r--r-- 1 redacted redacted   441272 ápr   28 14:53 libafsnmp.a
-rwxr-xr-x 1 redacted redacted     1150 ápr   28 14:53 libafsnmp.la*
-rw-r--r-- 1 redacted redacted  1282810 ápr   28 14:53 libafsocket.a
-rwxr-xr-x 1 redacted redacted     1131 ápr   28 14:53 libafsocket.la*
-rw-r--r-- 1 redacted redacted   397876 ápr   28 14:53 libafsql.a
-rwxr-xr-x 1 redacted redacted     1111 ápr   28 14:53 libafsql.la*
```

syslog-ng/Makefile.am
```diff
@@ -26,5 +29,13 @@ endif
 
 INSTALL_EXEC_HOOKS			+= syslog-ng-install-exec-hook
 
+BUILT_SOURCES += syslog-ng/builtin-modules.c
```
MrAnno: This is generated even when the default dynamic mode is used.

configure.ac
MrAnno: Now mixed mode is gone :D

docker/apkbuild/axoflow/axosyslog/APKBUILD
```diff
@@ -124,6 +96,7 @@ package() {
 	rm -f usr/lib/libsyslog-ng-native-connector.a
 
 	install -d -m 700 "$pkgdir"/var/lib/syslog-ng
+	install="$pkgname.post-install"
```
MrAnno: This needs to be set on the top of the file (not in build).

lib/plugin.c
```diff
@@ -390,49 +390,48 @@ plugin_load_module(PluginContext *context, const gchar *module_name, CfgArgs *ar
...
+      if (!mod)
+        return FALSE;
+      g_module_make_resident(mod);
+
+      module_info = _get_module_info(mod, module_name);
+      if (!module_info)
+        {
+          msg_error("No ModuleInfo structure found in plugin",
+                    evt_tag_str("module", module_name));
+          return FALSE;
```
MrAnno, line 414: We no longer call the module's init method here when `ModuleInfo` is missing.

63/63 modules in this repo have both of them, and I don't know any third-party users of such weirdness.

configure.ac
bazsi: yes, setting enable_static/disable_static overrides the default that LT_INIT() initializes.

I've now changed these assignments to AC_ENABLE_SHARED, AC_DISABLE_SHARED

```
 -- Macro: AC_DISABLE_SHARED
 -- Macro: AM_DISABLE_SHARED
     Change the default behaviour for ‘LT_INIT’ to disable shared
     libraries.  The user may still override this default by specifying
     ‘--enable-shared’.  The option ‘disable-shared’ to ‘LT_INIT’ is a
     shorthand for this.  ‘AM_DISABLE_SHARED’ is a deprecated alias for
     ‘AC_DISABLE_SHARED’.

 -- Macro: AC_ENABLE_SHARED
 -- Macro: AM_ENABLE_SHARED
     Change the default behaviour for ‘LT_INIT’ to enable shared
     libraries.  This is the default on all systems where Libtool knows
     how to create shared libraries.  The user may still override this
     default by specifying ‘--disable-shared’.  The option ‘shared’ to
     ‘LT_INIT’ is a shorthand for this.  ‘AM_ENABLE_SHARED’ is a
     deprecated alias for ‘AC_ENABLE_SHARED’.

 -- Macro: AC_DISABLE_STATIC
 -- Macro: AM_DISABLE_STATIC
     Change the default behaviour for ‘LT_INIT’ to disable static
     libraries.  The user may still override this default by specifying
     ‘--enable-static’.  The option ‘disable-static’ to ‘LT_INIT’ is a
     shorthand for this.  ‘AM_DISABLE_STATIC’ is a deprecated alias for
     ‘AC_DISABLE_STATIC’.

 -- Macro: AC_ENABLE_STATIC
 -- Macro: AM_ENABLE_STATIC
     Change the default behaviour for ‘LT_INIT’ to enable static
     libraries.  This is the default on all systems where shared
     libraries have been disabled for some reason, and on most systems
     where shared libraries have been enabled.  If shared libraries are
     enabled, the user may still override this default by specifying
     ‘--disable-static’.  The option ‘static’ to ‘LT_INIT’ is a
     shorthand for this.  ‘AM_ENABLE_STATIC’ is a deprecated alias for
     ‘AC_ENABLE_STATIC’.

```

lib/plugin.c
bazsi: this does not compile, unless all small executables (pdbtool and dqtool for instance) provides its own version of register_builtin_modules()

unfortunately weak does not work the same way on macos, so this might need a separate solution.

configure.ac
bazsi: I removed my previous comment on AC_ENABLED_SHARED and AC_DISABLE_STATIC as they did not override the settings set in LT_INIT.

I have now reworked this part and after a few iterations, I ended up specifying "-static" for modules and the core library.

```
+elif test "x$linking_mode" = "xmonolithic"; then
+       MODULE_LDFLAGS="-static"
+       CORE_LDFLAGS="-static"
```

this option is processed by libtool and will disable the shared object for specific libraries. This will link syslog-ng properly. I needed to add solutions to pdbtool to load its module dependencies (as we will not have them as shared objects, e.g. it needs to link them too).

Once we have static libraries, they will be installed. Could not find a way to override that behaviour. They don't need to be installed by the package however.

syslog-ng/Makefile.am
bazsi: fixed.

docker/apkbuild/axoflow/axosyslog/APKBUILD
bazsi: fixed and confirmed that the postinstall script now runs when I build the production image.

lib/plugin.c
bazsi: I don't think this is an issue, requirement of an .so being a plugin is to have a module_info struct, we are still running the init function, but only once moduleinfo is there.

MrAnno (approved):
Works with JIT too.

PR #1020, opened 2026-03-30
.github/workflows/devshell: set light worker processes to 16

bazsi:


MrAnno (approved):
We have a mysterious heap corruption crash that this parallel execution PR triggered somehow, but it's flaky. The best would be if we could reproduce it with ASAN enabled.

bazsi:
This should also fix the failure on main.

PR #1026, opened 2026-04-02
lib/metrics: add new metric input_transport_errors_total

OverOrion:
Example:
```
syslogng_input_transport_errors_total{address="127.0.0.1:5514",driver="syslog",peer_address="127.0.0.1",reason="invalid-frame-header",transport="tls"} 0
syslogng_input_transport_errors_total{address="127.0.0.1:5514",driver="syslog",peer_address="127.0.0.1",reason="tls-handshake",transport="tls"} 2
```

### DIFF
--- a/lib/logproto/logproto-auto-server.c
+++ b/lib/logproto/logproto-auto-server.c
@@ -32,6 +32,7 @@ G_STATIC_ASSERT(LOG_TRANSPORT_RA_BUFFER_SIZE >= RFC6587_MAX_FRAME_LEN_DIGITS + 1
 typedef struct _LogProtoAutoServer
 {
   LogProtoServer super;
+  StatsClusterKeyBuilder *kb;
 } LogProtoAutoServer;
 
 static inline gboolean
@@ -68,7 +69,7 @@ _construct_detected_proto(LogProtoAutoServer *self, const gchar *detect_buffer, 
 
       msg_debug("Auto-detected octet-counted-framing, using framed protocol",
                 evt_tag_int("fd", fd));
-      return log_proto_framed_server_new(NULL, self->super.options);
+      return log_proto_framed_server_new(NULL, self->super.options, self->kb);
     }
 
   if (detect_buffer[0] == '<')
@@ -125,7 +126,8 @@ log_proto_auto_handshake(LogProtoServer *s, gboolean *handshake_finished, LogPro
 }
 
 LogProtoServer *
-log_proto_auto_server_new(LogTransport *transport, const LogProtoServerOptions *options)
+log_proto_auto_server_new(LogTransport *transport, const LogProtoServerOptions *options,
+                          StatsClusterKeyBuilder *kb)
 {
   LogProtoAutoServer *self = g_new0(LogProtoAutoServer, 1);
 
@@ -135,5 +137,6 @@ log_proto_auto_server_new(LogTransport *transport, const LogProtoServerOptions *
   log_proto_server_init(&self->super, transport, options);
   self->super.handshake = log_proto_auto_handshake;
   self->super.poll_prepare = log_proto_auto_server_poll_prepare;
+  self->kb = kb;
   return &self->super;
 }
--- a/lib/logproto/logproto-framed-server.c
+++ b/lib/logproto/logproto-framed-server.c
@@ -425,7 +425,8 @@ log_proto_framed_server_free(LogProtoServer *s)
 }
 
 LogProtoServer *
-log_proto_framed_server_new(LogTransport *transport, const LogProtoServerOptions *options)
+log_proto_framed_server_new(LogTransport *transport, const LogProtoServerOptions *options,
+                            StatsClusterKeyBuilder *kb)
 {
   LogProtoFramedServer *self = g_new0(LogProtoFramedServer, 1);
 
--- a/lib/logproto/logproto-framed-server.c
+++ b/lib/logproto/logproto-framed-server.c
@@ -24,6 +24,9 @@
 #include "logproto.h"
 #include "messages.h"
 #include "str-utils.h"
+#include "gsocket.h"
+#include "stats/stats-registry.h"
+#include "stats/stats-cluster-single.h"
 
 #include <errno.h>
 #include <ctype.h>
@@ -63,6 +66,12 @@ typedef struct _LogProtoFramedServer
   /* auxiliary data (e.g. GSockAddr, other transport related meta
    * data) associated with the already buffered data */
   LogTransportAuxData buffer_aux;
+
+  struct
+  {
+    StatsClusterKey *invalid_frame_header_key;
+    StatsCounterItem *invalid_frame_header_counter;
+  } metrics;
 } LogProtoFramedServer;
 
 static LogProtoPrepareAction
@@ -143,10 +152,40 @@ log_proto_framed_server_fetch_data(LogProtoFramedServer *self, gboolean *may_rea
   return TRUE;
 }
 
+static void
+_handle_invalid_frame_header(LogProtoFramedServer *self, guint32 frame_start)
+{
+  gchar peer_buf[MAX_SOCKADDR_STRING] = "unknown";
+  gchar local_buf[MAX_SOCKADDR_STRING] = "unknown";
+
+  GSockAddr *peer_addr = g_socket_get_peer_name(self->super.transport_stack.fd);
+  if (peer_addr)
+    {
+      g_sockaddr_format(peer_addr, peer_buf, sizeof(peer_buf), GSA_FULL);
+      g_sockaddr_unref(peer_addr);
+    }
+
+  GSockAddr *local_addr = g_socket_get_local_name(self->super.transport_stack.fd);
+  if (local_addr)
+    {
+      g_sockaddr_format(local_addr, local_buf, sizeof(local_buf), GSA_FULL);
+      g_sockaddr_unref(local_addr);
+    }
+
+  msg_error("Invalid frame header",
+            evt_tag_mem("header", &self->buffer[frame_start],
+                        MIN(self->buffer_end - frame_start, RFC6587_MAX_FRAME_LEN_DIGITS)),
+            evt_tag_str("peer_addr", peer_buf),
+            evt_tag_str("local_addr", local_buf));
+
+  stats_counter_inc(self->metrics.invalid_frame_header_counter);
+}
+
 static gboolean
 log_proto_framed_server_extract_frame_length(LogProtoFramedServer *self, gboolean *need_more_data)
 {
   gint i;
+  guint32 frame_start = self->buffer_pos;
 
   *need_more_data = TRUE;
   self->frame_len = 0;
@@ -164,8 +203,7 @@ log_proto_framed_server_extract_frame_length(LogProtoFramedServer *self, gboolea
         }
       else
         {
-          msg_error("Invalid frame header",
-                    evt_tag_mem("header", &self->buffer[self->buffer_pos], (i - self->buffer_pos)));
+          _handle_invalid_frame_header(self, frame_start);
           return FALSE;
         }
     }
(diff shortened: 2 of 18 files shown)


lib/logproto/logproto-framed-server.c
```diff
@@ -143,6 +152,33 @@ log_proto_framed_server_fetch_data(LogProtoFramedServer *self, gboolean *may_rea
...
+  guint16 port = 0;
+  GSockAddr *local_addr = g_socket_get_local_name(self->super.transport_stack.fd);
+  if (local_addr)
+    {
+      port = g_sockaddr_get_port(local_addr);
+      g_sockaddr_unref(local_addr);
+    }
+
+  msg_error("Invalid frame header",
+            evt_tag_mem("header", &self->buffer[self->buffer_pos], MIN(self->buffer_end, RFC6587_MAX_FRAME_LEN_DIGITS)),
```
alltilla: Can we print from the beginning of the expected header? A lot of the times the interesting part is there, like the year or the beginning of an IP address.

I think we could store the beginning position in `log_proto_framed_server_extract_frame_length()` and pass it here.

Of course it won't work for messages split at the octet count, but most of the time it would.

lib/logproto/logproto-auto-server.c
```diff
@@ -68,7 +68,7 @@ _construct_detected_proto(LogProtoAutoServer *self, const gchar *detect_buffer,
 
       msg_debug("Auto-detected octet-counted-framing, using framed protocol",
                 evt_tag_int("fd", fd));
-      return log_proto_framed_server_new(NULL, self->super.options);
+      return log_proto_framed_server_new(NULL, self->super.options, NULL);
```
alltilla: We use the `auto` proto most of the time. If it is not super hard to implement the `kb` here, I think we should.

tests/light/functional_tests/source_drivers/network_source/test_transport_error_metrics.py
```diff
@@ -0,0 +1,88 @@
...
+
+    syslog_ng.start(config)
+
+    # Send raw data that is not a valid RFC6587 frame header (starts with a letter, not a digit)
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sock.connect(("localhost", source.options["port"]))
+    sock.sendall(b"INVALID_FRAME_HEADER\n")
+    sock.close()
+
+    assert wait_until_true(lambda: _get_metric_value(config, "invalid-frame-header") == 1)
```
alltilla, line 64: Can/Should we grab the error log and assert for it?

fernya (commented):
I focused on manual triggering of the added metric and check if it increments when a fault condition appears.
I used the following syslog-ng.conf:

```
@version: current

options {
  stats (level(3));
};

log {
  source {
    # framed proto
    syslog(port(4444));

    #new-line separated proto
    network(port(5555));

    # tls
    network(port(6666) transport(tls) tls(peer-verify(yes) pkcs12-file("server.p12")));
  };

  destination { stdout(); };
```

## Frame error-related tests
Used netcat against bare 'syslog' source: `nc localhost 4444`
  - happy path - sent `3<space>a<lf>`, log entry created, 'input_transport_errors_total' **was not** incremented
  - wrong path - sent `4<space>a<lf>`, fails as expected, 'input_transport_errors_total' **was** incremented

## TLS issue-related tests
Generated two server and client keysets 
Used openssl s_client against tls wrapped 'network' source: `echo <logentry> | openssl s_client -connect localhost:6666 -CAfile ca.crt -cert client.crt -key client.key`

same client & server CA|expected|result|passed
-----------------------------------|-------------|--------|----------
yes|OK|OK|✅
not|NOK|NOK|✅

fernya:
I reran the TLS tests and the error cause code/strings are added to the log metrics and their count seems to be properly maintained per unique category.

lib/logproto/logproto-framed-server.c
```diff
@@ -396,6 +433,27 @@ _step_state_machine(LogProtoFramedServer *self, const guchar **msg, gsize *msg_l
     }
 }
 
+static void
+_allocate_stats_keys(LogProtoFramedServer *self, StatsClusterKeyBuilder *kb)
```
bazsi: the name `_allocate_stats_keys()` tells me it allocates only the key for various metrics, but it also allocates the metric.

I know we sometimes we split the allocation of keys vs metrics into separate functions, but in this case I like the fact it is in the same function, but I think we should name it like `_register_stats()` or something.

lib/logproto/logproto-framed-server.c
```diff
@@ -414,10 +472,31 @@ log_proto_framed_server_fetch(LogProtoServer *s, const guchar **msg, gsize *msg_
   return status;
 }
 
+
+
+static void
+_deallocate_stats_keys(LogProtoFramedServer *self)
```
bazsi: similar nit to the allocation

lib/logproto/logproto-framed-server.c
```diff
@@ -143,10 +152,39 @@ log_proto_framed_server_fetch_data(LogProtoFramedServer *self, gboolean *may_rea
...
+  if (local_addr)
+    {
+      port = g_sockaddr_get_port(local_addr);
+      g_sockaddr_unref(local_addr);
+    }
+
+  msg_error("Invalid frame header",
+            evt_tag_mem("header", &self->buffer[frame_start],
+                        MIN(self->buffer_end - frame_start, RFC6587_MAX_FRAME_LEN_DIGITS)),
+            evt_tag_str("peer_addr", peer_addr ? addr_buf : "unknown"),
```
bazsi: I would format the peer and local addresses both using GSA_ADDRESS_PORT, the port number is less important for the source, but the IP is definitely important for the destination. We might be listening on different interfaces.

lib/transport/transport-tls.c
```diff
@@ -386,10 +443,28 @@ log_transport_tls_free_method(LogTransport *s)
 {
   LogTransportTLS *self = (LogTransportTLS *) s;
 
+  if (self->metrics.error_cluster)
```
bazsi: I'd rather see this extracted into a separate function (_unregister_stats?)

lib/transport/transport-tls.c
```diff
@@ -202,6 +211,48 @@ log_transport_tls_aux_data_add_peer_info(LogTransportTLS *self, LogTransportAuxD
...
+
+  StatsCounterItem *counter;
+  stats_lock();
+  {
+    if (self->metrics.error_cluster)
+      {
+        StatsCounterItem *old_counter = stats_cluster_single_get_counter(self->metrics.error_cluster);
+        stats_unregister_counter(&self->metrics.error_cluster->key, SC_TYPE_SINGLE_VALUE, &old_counter);
+      }
+    self->metrics.error_cluster = stats_register_counter(STATS_LEVEL1, key, SC_TYPE_SINGLE_VALUE, &counter);
```
bazsi: why do we keep error_cluster member around? we can register the counter and then immediately release it.

stats_register_dynamic_counter/stats_unregister_dynamic_counter

lib/logproto/logproto-framed-server.c
bazsi: I saw this implemented, so resolving this one.

lib/logproto/logproto-auto-server.c
bazsi: this was done too.

lib/transport/transport-tls.c
OverOrion: Good point, no need for that!

lib/logproto/logproto-framed-server.c
OverOrion: Good point!

I changed it and it looked like this with `GSA_ADDRESS_PORT`
```
[2026-04-23T08:42:38.687426] Syslog connection accepted; fd='16', client='AF_INET(127.0.0.1:46136)', local='AF_INET(127.0.0.1:5513)'
[2026-04-23T08:42:38.687790] Invalid frame header; header='ASD\x0a', peer_addr='127.0.0.1:46136', local_addr='127.0.0.1:5513'
[2026-04-23T08:42:38.687823] Syslog connection closed; fd='16', client='AF_INET(127.0.0.1:46136)', local='AF_INET(127.0.0.1:5513)'
```

I think `GSA_FULL` fits better here, so I went for it. Let me know if `GSA_ADDRESS_PORT` should be used instead.

tests/light/functional_tests/source_drivers/network_source/test_transport_error_metrics.py
```diff
@@ -0,0 +1,148 @@
...
+            "peer-verify": "optional-untrusted",
+        },
+    )
+    file_destination = config.create_file_destination(file_name="output.log")
+    config.create_logpath(statements=[source, file_destination])
+
+    syslog_ng.start(config)
+
+    # Connect with plain TCP (no TLS handshake) to trigger a TLS handshake error
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
OverOrion: rename it to reflect its TCP-only nature

tests/light/functional_tests/source_drivers/network_source/test_transport_error_metrics.py
```diff
@@ -0,0 +1,148 @@
...
+    sock.connect(("localhost", source.options["port"]))
+    sock.sendall(b"THIS IS NOT A TLS HANDSHAKE\n")
+    sock.close()
+
+    assert wait_until_true(lambda: _get_metric_value(config, "tls-handshake") == 1)
+    assert syslog_ng.wait_for_message_in_console_log("SSL error while reading stream") != []
```
OverOrion: Comment explanation

tests/light/functional_tests/source_drivers/network_source/test_transport_error_metrics.py
```diff
@@ -0,0 +1,148 @@
...
+#############################################################################
+import socket
+import ssl
+
+from axosyslog_light.common.blocking import wait_until_true
+from axosyslog_light.common.file import copy_shared_file
+from axosyslog_light.syslog_ng_ctl.prometheus_stats_handler import MetricFilter
+
+
+METRIC_NAME = "syslogng_input_transport_errors_total"
```
OverOrion, line 33: Mention why UDP is missing in a comment

tests/light/functional_tests/source_drivers/network_source/test_transport_error_metrics.py
```diff
@@ -0,0 +1,148 @@
...
+            "ca-file": trusted_ca_path,
+            "peer-verify": "required-trusted",
+        },
+    )
+    file_destination = config.create_file_destination(file_name="output.log")
+    config.create_logpath(statements=[source, file_destination])
+
+    syslog_ng.start(config)
+
+    # TLS connection without  client cert, server requires one and will reject
```
OverOrion: typo

tests/light/functional_tests/source_drivers/network_source/test_transport_error_metrics.py
```diff
@@ -0,0 +1,148 @@
...
+    client_ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+    client_ctx.check_hostname = False
+    client_ctx.verify_mode = ssl.CERT_NONE
+
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sock.connect(("localhost", source.options["port"]))
+    try:
+        client_ctx.wrap_socket(sock, server_hostname="localhost")
```
OverOrion: switch to `do_handshake`
https://docs.python.org/3/library/ssl.html#ssl.SSLSocket.do_handshake

lib/logproto/logproto-framed-server.c
bazsi: you are right sorry. we used that format only metrics and this is a log message.

bazsi:
fyi, the compile errors should go away once this is rebased against current main

lib/logproto/logproto-framed-server.c
bazsi: one tiny nit is that one of the calls is register_stats() the other unregister_stat(), note the differences in purality. But I don't mind that much, it's ok this way. If you do another round, it's better to fix it.

lib/transport/transport-tls.c
bazsi: looks good, the stats API is terrible though. :(

tests/light/functional_tests/source_drivers/network_source/test_transport_error_metrics.py
OverOrion: I think dropping it is better though, just causes confusion.

alltilla (approved):

PR #1029, opened 2026-04-07
light/filter: Add timing diagnosis to rate limit filter test assertion

mitzkia:
- without this change it is hard to identify the rare case when the CI host is slower than expected, causing the token bucket to partially refill mid-test and producing more matched messages than expected
  - the fix captures the syslog-ng console log timestamps of the first and last incoming message, calculates the elapsed processing time, and compares it against token_refill_period_ms (1000 / rate)
  - if elapsed > token_refill_period_ms the assertion failure message explicitly names the root cause: CI host slowness caused a token bucket refill, not a filter logic bug


mitzkia:
This will be fixed in #1031

PR #1030, opened 2026-04-10
Disk-buffer: don't save memory segments on reload

sodomelle:
If the queue is present in the persist only drop it if it's a different type of queue, or it is not compatible with the new buffer. Otherwise use it.

### DIFF
--- a/modules/diskq/diskq.c
+++ b/modules/diskq/diskq.c
@@ -140,23 +140,34 @@ _acquire_queue(LogDestDriver *dd, const gchar *persist_name, gint stats_level,
 {
   DiskQDestPlugin *self = log_driver_get_plugin(&dd->super, DiskQDestPlugin, DISKQ_PLUGIN_NAME);
   GlobalConfig *cfg = log_pipe_get_config(&dd->super.super);
-  LogQueue *queue;
+  LogQueue *queue = NULL;
   gchar *persist_qfile_name, *new_qfile_name = NULL;
 
   if (persist_name)
-    log_queue_unref(cfg_persist_config_fetch(cfg, persist_name));
+    queue = cfg_persist_config_fetch(cfg, persist_name);
+
+  if (queue && (!log_queue_has_type(queue, log_queue_disk_get_type()) ||
+      !log_queue_disk_has_compatible_options((LogQueueDisk *)queue, &self->options)))
+    {
+      log_queue_unref(queue);
+      queue = NULL;
+    }
 
   persist_qfile_name = persist_state_lookup_string(cfg->state, persist_name, NULL, NULL);
-  queue = _create_and_start_disk_queue_with_filename_from_persist(
-            self, persist_qfile_name, persist_name, stats_level, driver_sck_builder, queue_sck_builder);
+
   if (queue)
-    goto exit;
+    log_queue_disk_set_options((LogQueueDisk *)queue, &self->options);
+  else
+    queue = _create_and_start_disk_queue_with_filename_from_persist(
+              self, persist_qfile_name, persist_name, stats_level, driver_sck_builder, queue_sck_builder);
 
-  new_qfile_name = qdisk_get_next_filename(self->options.dir, self->options.reliable);
-  queue = _create_and_start_disk_queue_with_new_filename(self, new_qfile_name, persist_name, stats_level,
-                                                         driver_sck_builder, queue_sck_builder);
+  if (!queue)
+    {
+      new_qfile_name = qdisk_get_next_filename(self->options.dir, self->options.reliable);
+      queue = _create_and_start_disk_queue_with_new_filename(self, new_qfile_name, persist_name, stats_level,
+                                                             driver_sck_builder, queue_sck_builder);
+    }
 
-exit:
   if (queue)
     {
       log_queue_set_throttle(queue, dd->throttle);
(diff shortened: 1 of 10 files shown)


modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -868,6 +868,9 @@ _free(LogQueue *s)
 {
   LogQueueDiskNonReliable *self = (LogQueueDiskNonReliable *)s;
 
+  gboolean persistent;
+  log_queue_disk_stop(&self->super.super, &persistent);
```
MrAnno, line 891: log_queue_disk_stop() originally called `log_queue_queued_messages_sub()` on reload. I remember some trickery around this in `logqueue.c` so we should check whether the queued stats are still correct after multiple reloads.

MrAnno (dismissed):

MrAnno (dismissed):

alltilla:
I'll review this.

mitzkia (dismissed):
Light changes are approved from my side

modules/diskq/logqueue-disk-reliable.c
```diff
@@ -351,7 +351,7 @@ _pop_head(LogQueue *s, LogPathOptions *path_options)
 static inline gboolean
 _is_reserved_buffer_size_reached(LogQueueDiskReliable *self)
 {
-  return qdisk_get_empty_space(self->super.qdisk) < qdisk_get_flow_control_window_bytes(self->super.qdisk);
+  return qdisk_get_empty_space(self->super.qdisk) < self->flow_control_window_bytes;
```
alltilla: There is one more use in line 397. We could use this there and remove the `QDisk::` getter altogether.

modules/diskq/diskq.c
```diff
@@ -140,23 +140,33 @@ _acquire_queue(LogDestDriver *dd, const gchar *persist_name, gint stats_level,
...
+      queue = NULL;
+    }
 
   persist_qfile_name = persist_state_lookup_string(cfg->state, persist_name, NULL, NULL);
-  queue = _create_and_start_disk_queue_with_filename_from_persist(
-            self, persist_qfile_name, persist_name, stats_level, driver_sck_builder, queue_sck_builder);
+
   if (queue)
-    goto exit;
+    log_queue_disk_set_options((LogQueueDisk *)queue, &self->options);
```
alltilla, line 159: This is very risky, we could change the `Options` under an already created `QDisk`, where `reliable` or `capacity_bytes` can be changed to different values, which can cause all kinds of erroneous behavior.

E.g. `reliable(no)` -> `reliable(yes)` will keep it `LogQueueDiskNonReliable`, but `QDisk` will be `reliable=TRUE`, etc.

alltilla:
```
Breakpoint 1.1, log_queue_push_tail (self=0x51f000062e80, msg=0x517000158900, path_options=0x7ffff4ae0c40) at ../lib/logqueue.h:136
136       log_queue_update_processing_latency(self, msg);
(gdb) p self->push_tail
$6 = (void (*)(LogQueue *, LogMessage *, const LogPathOptions *)) 0x7ffff1b6edb0 <_non_reliable_push_tail>
(gdb) p ((LogQueueDisk *)self)->qdisk->options->reliable
$7 = 1
(gdb) 
```

MrAnno: We checked all callsites of the options fields, and we already store and use all the important fields in/from the QDisk header or other cached data structures, so this should not cause problems.

We realized this "user configured the buffer to be smaller, changed other fields too" problem years ago, so we started storing these important fields in the header.

The reliable->non-reliable transition can be a problem though, but I have some memories that this transition never worked properly, because the QDisk file has different magic values: `SLQF` and `SLRQ` (user changed the queue type, but we errored out with `Error reading disk-queue file header. Invalid magic"`).

MrAnno: To rephrase it: we have been changing the `options` field under the disk-buffer for years, we just did it implicitly: we wrote everything to disk, destroyed memory objects, created new memory objects, loaded the disk buffer back.

But yes, we no longer destroy the memory objects, so there may be some new bugs (for example, that's why Tomi introduced the `LogQueueDiskReliable::flow_control_window_bytes` cached value).

MrAnno:
We've checked the reliable->nonreliable transition, it works because of the retry mechanism, but we should fix this in the future (this is the behavior BEFORE this PR):

```
[2026-04-17T13:30:30.176660] Reliable disk-buffer state; operation='save', filename='/home/anno/dev/x/syslog/install/var/syslog-ng-00001.rqf', number_of_messages='0'
[2026-04-17T13:30:30.176660] Error reading disk-queue file header. Invalid magic; filename='/home/anno/dev/x/syslog/install/var/syslog-ng-00001.rqf'
[2026-04-17T13:30:30.176660] Error reading disk-queue file header. Invalid magic; filename='/home/anno/dev/x/syslog/install/var/syslog-ng-00001.rqf'
[2026-04-17T13:30:30.176660] Error initializing log queue;
[2026-04-17T13:30:30.176660] Non-reliable disk-buffer state; operation='load', filename='/home/anno/dev/x/syslog/install/var/syslog-ng-00002.qf', number_of_messages='0'
```

modules/diskq/logqueue-disk-reliable.c
sodomelle: We no longer cache this.

modules/diskq/diskq.c
sodomelle: If the options used outside of QDisk change, drop the queue and create a new one.

modules/diskq/logqueue-disk.h
```diff
@@ -78,5 +78,6 @@ void log_queue_disk_drop_msg_and_suspend_src(LogQueueDisk *self, LogMessage *msg
 gboolean log_queue_disk_serialize_msg(LogQueueDisk *self, LogMessage *msg, GString *serialized);
 gboolean log_queue_disk_deserialize_msg(LogQueueDisk *self, GString *serialized, LogMessage **msg);
 void log_queue_disk_set_options(LogQueueDisk *self, DiskQueueOptions *options);
+gboolean is_queue_compatible(LogQueueDisk *self, DiskQueueOptions *options);
```
MrAnno, line 81: This needs a prefix as it is a public function.

log_queue_has_compatible_options(self, options) may be a good name choice to avoid the double "queue" in name (or the original name you chose was also good).

MrAnno (approved):

modules/diskq/logqueue-disk.h
sodomelle: Renamed it.

MrAnno (approved):

alltilla (approved):

modules/diskq/diskq.c
alltilla: thanks!

PR #1032, opened 2026-04-10
otel source: `mode(filterx-dict)`

alltilla:
Adds a `mode()` option to the `opentelemetry()` source:

* `mode(logmessage)`: The old behavior, creating `${.otel_raw.<...>}` NVs.
* `mode(filterx-dict)`: Creates declared `resource`, `scope` and `log`
  FilterX variables, each holding a plain FilterX dict converted from the
  incoming protobuf message. The values are fully copied, nothing borrows
  from the gRPC request.

Only log records consult the mode, spans and metrics are out of scope for now.

---

Performance, measured on 4 cores (AWS EC2 m5n.2xlarge — 8 vCPU, 32 GB RAM, Ubuntu 24.04):

| FilterX work per record | before: OTel objects | after: mode(filterx-dict) | speedup |
|---|---|---|---|
| a few field reads | 5,199 EPS/core | 6,459 EPS/core | +24% |
| heavy field mapping (dozens of reads) | 888 EPS/core | 3,106 EPS/core | +250% |

---

Also fixes the source address extraction for percent-encoded gRPC peer URIs (newer gRPC versions encode IPv6 peers as `ipv6:%5B::1%5D:<port>`), which affected `mode(logmessage)` too.


OverOrion (dismissed):
Looks great! 🚀

alltilla:
I have found a crash during formatting sometimes, I'll check it.

modules/grpc/otel/filterx/otel-field-converter.cpp
```diff
@@ -422,3 +425,77 @@ syslogng::grpc::otel::iter_on_otel_protobuf_message_fields(google::protobuf::Mes
...
 }
+
+static FilterXObject *_convert_field_value_to_plain(FilterXObject *value);
+
+static gboolean
+_add_field_to_dict(FilterXObject *key, FilterXObject *value, gpointer user_data)
+{
+  FilterXObject *dict = (FilterXObject *) user_data;
+
+  FilterXObject *plain_value = _convert_field_value_to_plain(value);
```
fernya, line 436: Do we have a limit on this recursion depth?
If grpc framework already enforces, then I'd put a comment here to signal the lack of it here.

modules/grpc/otel/otel-source-services.hpp
```diff
@@ -209,27 +237,48 @@ syslogng::grpc::otel::LogsServiceCall::Proceed(bool ok)
...
-                      ProtobufParser::store_raw_metadata(msg, ctx.peer(), resource, resource_logs_schema_url, scope,
-                                                         scope_logs_schema_url);
-                      ProtobufParser::store_raw(msg, log_record);
-                    }
-                  else
+                  ProtobufParser::store_peer_address(msg, ctx.peer());
+
+                  gboolean converted = FALSE;
+                  FilterXEvalContext eval_context;
+                  FILTERX_EVAL_BEGIN_CONTEXT(eval_context, NULL, msg, filterx_scope_var_layout)
```
fernya: This is in a loop starting in L220.
The old usage `log_pipe_queue_slow_path` -> virtual call to `log_filterx_pipe_queue` freed the `scope`/`fx_scope` on return as it is allocated by `g_alloca`.
This block should be put into a function in its current form, or the macro should have a variant where a buffer is provided for the scope.

Confirmed and triggered with Claude.

fernya (changes requested):
The work looks robust, but needs some tweaks to become teh bestest. ;)

I found a non-clear recursion depth issue, also confirmed a smell around `FILTERX_EVAL_BEGIN_CONTEXT`.

modules/grpc/otel/filterx/otel-field-converter.cpp
alltilla: Comment added.

modules/grpc/otel/otel-source-services.hpp
alltilla: Nice catch, thanks! Fixed.

fernya (approved):
Let's 🚢 it!

PR #1033, opened 2026-04-12
Filterx object make repr type specific

bazsi:
This PR changes the repr() function (and associated object method) to return a more specific format and thereby separate
the parallel str() and repr().

I needed this when I started introducing the tuple() type. When a tuple is serialized to JSON, it is basically a list (as JSON
does not have a tuple data type), but this also means that we need a representation of objects that we can use in log messages generated by the filterx interpreter.

While it seems simple in the surface, it cascaded into a lot of changes, so this PR is not small, but mostly trivial changes.


alltilla:
TODO: Followup `repr()` changes.

sodomelle (approved):
Looks go to me.

tests/light/functional_tests/filterx/test_filterx_types.py
```diff
@@ -0,0 +1,358 @@
...
+log {{
+    source(genmsg);
+    {render_filterx_exprs(expr)};
+    destination(dest_final);
+}};
+"""
+    config.set_raw_config(preamble)
+    return (file_final,)
+
+
```
mitzkia, line 74: I have found that these filterx types are missing from the covered testcases:
- I am not sure if these types are related to this PR or not
```
filterx_type_init(&FILTERX_TYPE_NAME(ip));
filterx_type_init(&FILTERX_TYPE_NAME(mapping));
filterx_type_init(&FILTERX_TYPE_NAME(message_value));
filterx_type_init(&FILTERX_TYPE_NAME(null));
filterx_type_init(&FILTERX_TYPE_NAME(sequence));
filterx_type_init(&FILTERX_TYPE_NAME(string));
filterx_type_init(&FILTERX_TYPE_NAME(subnet));
```

mitzkia:
Other than that review finding Light part is approved from my side. 
Maybe later I would reduce the test code here in this file: tests/light/functional_tests/filterx/test_filterx_types.py
If we replace tests with 1 parameterized testcase the file will be more readable and about 100 lines shorter

PR #1034, opened 2026-04-13
Filterx cache json file add default value

sodomelle:
When the file given to `cache_json_file()` function doesn't exist, or an error occurs during loading it, instead of exiting with config error, if the `default_value` parameter is provided, then the `default_value` will be used.

Example: `$MSG = cache_json_file("./test.json", default_value={"key": "value"});`

### DIFF
--- a/lib/filterx/func-cache-json-file.c
+++ b/lib/filterx/func-cache-json-file.c
@@ -26,6 +26,7 @@
 #include "filterx/filterx-sequence.h"
 #include "filterx/filterx-mapping.h"
 #include "filterx/expr-literal.h"
+#include "filterx/expr-literal-container.h"
 #include "filterx/filterx-eval.h"
 #include "scratch-buffers.h"
 #include "file-monitor.h"
@@ -38,7 +39,7 @@
 #include <errno.h>
 #include <glib.h>
 
-#define FILTERX_FUNC_CACHE_JSON_FILE_USAGE "Usage: cache_json_file(\"/path/to/file.json\")"
+#define FILTERX_FUNC_CACHE_JSON_FILE_USAGE "Usage: cache_json_file(\"/path/to/file.json\", default_value=[dict])"
 
 #define CACHE_JSON_FILE_ERROR cache_json_file_error_quark()
 
@@ -64,6 +65,7 @@ typedef struct FilterXFunctionCacheJsonFile_
   gchar *filepath;
   gpointer cached_json;
   FileMonitor *file_monitor;
+  FilterXExpr *default_value;
 } FilterXFunctionCacheJsonFile;
 
 static gchar *
@@ -124,12 +126,34 @@ _load_json_file(const gchar *filepath, GError **error)
   return result;
 }
 
+static gboolean
+_cache_json_file_init(FilterXExpr *s, GlobalConfig *cfg)
+{
+  FilterXFunctionCacheJsonFile *self = (FilterXFunctionCacheJsonFile *) s;
+
+  if (self->default_value)
+    {
+      if (!filterx_expr_is_literal(self->default_value))
+        {
+          msg_error("The default_value argument for cache_json_file() has to be a literal",
+                    filterx_expr_format_location_tag(s));
+          return FALSE;
+        }
+    }
+  return filterx_function_init_method(&self->super, cfg);
+}
+
 static FilterXObject *
 _eval(FilterXExpr *s)
 {
   FilterXFunctionCacheJsonFile *self = (FilterXFunctionCacheJsonFile *) s;
+
   FilterXObject *cached_json = g_atomic_pointer_get(&self->cached_json);
-  return filterx_object_ref(cached_json);
+
+  if (cached_json)
+    return filterx_object_ref(cached_json);
+
+  return filterx_expr_eval_typed(self->default_value);
 }
 
 static void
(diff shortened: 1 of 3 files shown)


lib/filterx/func-cache-json-file.c
```diff
@@ -207,34 +230,83 @@ _file_monitor_callback(const FileMonitorEvent *event, gpointer user_data)
...
+        }
+      else
+        goto error;
+    }
+  else
+    {
+      filterx_expr_unref(self->default_value);
+      self->default_value = NULL;
+    }
+
```
OverOrion: Could we move the default value handling logic into a helper? I think that would help with readability, what do you think?
```c
  static gboolean
  _try_load_from_file_with_default(FilterXFunctionCacheJsonFile *self, GError **error)
  {
    if (self->filepath && _load_json_file_version(self, error))
      return TRUE;

    if (!self->default_value)
      return FALSE;

    g_clear_error(error);
    return TRUE;
  }

  Then the constructor is just early exits:

  FilterXExpr *
  filterx_function_cache_json_file_new(FilterXFunctionArgs *args, GError **error)
  {
    FilterXFunctionCacheJsonFile *self = g_new0(FilterXFunctionCacheJsonFile, 1);
    filterx_function_init_instance(&self->super, "cache_json_file", FXE_WORLD);

    self->super.super.eval = _eval;
    self->super.super.init = _cache_json_file_init;
    self->super.super.walk_children = _cache_json_file_walk;
    self->super.super.free_fn = _free;

    self->default_value = _extract_cache_json_file_default_value_dict(args, error);
    self->filepath = _extract_filepath(args, error);

    if (!filterx_function_args_check(args, error))
      goto error;

    if (!_try_load_from_file_with_default(self, error))
      goto error;

    if (self->cached_json)
      {
        filterx_expr_unref(self->default_value);
        self->default_value = NULL;

        self->file_monitor = file_monitor_new(self->filepath);
        file_monitor_add_watch(self->file_monitor, _file_monitor_callback, self);
        file_monitor_start(self->file_monitor);
      }

    filterx_function_args_free(args);
    return &self->super.super;

  error:
    filterx_function_args_free(args);
    filterx_expr_unref(&self->super.super);
    return NULL;
  }

```

sodomelle: Looks good. Changed it.

mitzkia (dismissed):
Light testcases are approved from my side

lib/filterx/func-cache-json-file.c
```diff
@@ -207,7 +230,47 @@ _file_monitor_callback(const FileMonitorEvent *event, gpointer user_data)
...
+  FilterXExpr **exprs[] = { &self->default_value };
+
+  for (gsize i = 0; i < G_N_ELEMENTS(exprs); i++)
+    {
+      if (!filterx_expr_visit(s, exprs[i], f, user_data))
+        return FALSE;
+    }
+
+  return TRUE;
```
MrAnno: this could be a one-liner, just `return filterx_expr_visit(s, &self->default_value, f, user_data);`

lib/filterx/func-cache-json-file.c
```diff
@@ -124,10 +126,30 @@ _load_json_file(const gchar *filepath, GError **error)
...
+    }
+  return filterx_function_init_method(&self->super, cfg);
+}
+
 static FilterXObject *
 _eval(FilterXExpr *s)
 {
   FilterXFunctionCacheJsonFile *self = (FilterXFunctionCacheJsonFile *) s;
+  if (self->default_value)
+    return filterx_expr_eval_typed(self->default_value);
```
MrAnno: Something seems off here.

If the file doesn't originally exist (or it's empty or erroneous), but it appears (or gets fixed) later, the file monitor will find it and load it.
The default value should only be returned until we receive the first correct JSON from a file.

(To make make this expectation clear in code, I'd always check whether we have a cached_json, and fall back to the default if we don't.)

lib/filterx/func-cache-json-file.c
sodomelle: Right, done.

lib/filterx/func-cache-json-file.c
sodomelle: I changed it to work like this.

alltilla (dismissed):

lib/filterx/func-cache-json-file.c
```diff
@@ -124,12 +126,35 @@ _load_json_file(const gchar *filepath, GError **error)
...
+}
+
 static FilterXObject *
 _eval(FilterXExpr *s)
 {
   FilterXFunctionCacheJsonFile *self = (FilterXFunctionCacheJsonFile *) s;
-  FilterXObject *cached_json = g_atomic_pointer_get(&self->cached_json);
-  return filterx_object_ref(cached_json);
+
+  if (self->cached_json)
```
MrAnno: This should be acquired with `g_atomic_pointer_get()` to be atomic (now we access the racy field twice). We can store it in a local variable and do the check+return, that way there will be no logical conflict.

MrAnno (commented):
LGTM otherwise.

MrAnno (approved):

lib/filterx/func-cache-json-file.c
sodomelle: Done.

PR #1038, opened 2026-04-16
helm: make syslog Service type configurable

avk999:
The Service type was hardcoded to NodePort, making it impossible to use LoadBalancer or ClusterIP without post-render patching.

- Add service.type value (default: NodePort for backwards compatibility)
- Add service.loadBalancerIP value for LoadBalancer deployments
- Only emit nodePort fields when explicitly set in values, avoiding conflicts when service type is LoadBalancer or ClusterIP

### DIFF
--- a/charts/axosyslog/templates/service.yaml
+++ b/charts/axosyslog/templates/service.yaml
@@ -6,7 +6,10 @@ metadata:
 spec:
   selector:
     app: {{ template "axosyslog.fullname" . }}-syslog
-  type: NodePort
+  type: {{ .Values.service.type | default "NodePort" }}
+  {{- with .Values.service.loadBalancerIP }}
+  loadBalancerIP: {{ . }}
+  {{- end }}
   ports:
 {{ if .Values.service.extraPorts }}
 {{ toYaml ( .Values.service.extraPorts ) | nindent 4 }}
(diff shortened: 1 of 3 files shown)


bazsi:
@csatib02 I am not sure if this is something that #994 resolves, can you please comment?

bazsi:
As I see this is complementary to that PR, so can be merged separately. 

This PR uses spec.loadBalancerIP  which is getting deprecated (as of k8s 1.24)

https://stackoverflow.com/questions/73750700/what-is-the-replacement-for-the-deprecated-loadbalancerip-attribute-in-services

GKE does support an annotation instead of loadBalancerIP already:

https://docs.cloud.google.com/kubernetes-engine/docs/concepts/service-load-balancer-parameters#spd-static-ip-parameters

AWS

In AWS no static IP allocation is possible, so it does not have a replacement.

Considering that all implementations that support static IP allocation (GKE and MetalLB) still support loadBalancerIP I am fine with merging this as is, the alternative would be to add a helm template for the new annotations and expose a provider independent setting in the values file. But I am not that much versed in the Kubernetes world, so I am open for inputs.

With that said, if nothing comes up and the signed-off-by line is added, we can merge this.

bazsi:
@avk999 I see you did add a sign-off-by to the documentation patch, but each of the commits need to be signed off.

Can you rebase your commits against the latest master and add the signed-off-by to the first commit as well? Thank you.

charts/axosyslog/templates/service.yaml
```diff
@@ -6,7 +6,10 @@ metadata:
 spec:
   selector:
     app: {{ template "axosyslog.fullname" . }}-syslog
-  type: NodePort
+  type: {{ .Values.service.type | default "NodePort" }}
+  {{- with .Values.service.loadBalancerIP }}
+  loadBalancerIP: {{ . }}
```
csatib02, line 11:
```suggestion
  loadBalancerIP: {{ . | quote }}
```

charts/axosyslog/values.yaml
```diff
@@ -169,6 +169,8 @@ openShift:
 
 service:
   create: true
+  type: NodePort  # NodePort, LoadBalancer, or ClusterIP
+  # loadBalancerIP: ""  # set when type is LoadBalancer
```
csatib02, line 173:
```suggestion
  # loadBalancerIP: ""  # DEPRECATED in k8s 1.24+. Use cloud-provider annotations.
```

charts/axosyslog/templates/service.yaml
```diff
@@ -6,7 +6,10 @@ metadata:
 spec:
   selector:
     app: {{ template "axosyslog.fullname" . }}-syslog
-  type: NodePort
+  type: {{ .Values.service.type | default "NodePort" }}
```
csatib02, line 9: It's already the default.

```suggestion
  type: {{ .Values.service.type }}
```

charts/axosyslog/values.yaml
```diff
@@ -169,6 +169,8 @@ openShift:
 
 service:
   create: true
+  type: NodePort  # NodePort, LoadBalancer, or ClusterIP
```
csatib02, line 172:
```suggestion
  type: NodePort  # NodePort, LoadBalancer, ClusterIP, or ExternalName
```

charts/axosyslog/templates/service.yaml
```diff
@@ -16,29 +19,39 @@ spec:
 {{- if .syslog.enabled }}
     - name: rfc3164-udp
       protocol: UDP
-      nodePort: {{ .syslog.rfc3164UdpPort | default 30514 }}
+      {{- if .syslog.rfc3164UdpPort }}
+      nodePort: {{ .syslog.rfc3164UdpPort }}
+      {{- end }}
```
csatib02, line 24: _nodePort may only be used when the service type is NodePort or LoadBalancer_

Please add this to the other nodeport cases as well.

```suggestion
	  {{- if and .syslog.rfc3164UdpPort (or (eq $.Values.service.type "NodePort") (eq
	  $.Values.service.type "LoadBalancer")) }}
	  nodePort: {{ .syslog.rfc3164UdpPort }}
	  {{- end }}
```

EngZaz:
@bazsi this pull request https://github.com/axoflow/axosyslog/pull/1066 covers the changes you requested here

MrAnno:
Closing in favor of #1066.

PR #1041, opened 2026-04-17
modules/xml: fix double unref

OverOrion:
The exit tag already calls filterx_object_unref() on the object, no need to do it in the helper method

- [x] NEWS file


MrAnno (approved):

MrAnno:
For example:
```
log.body = {"a": 3};
log.body = parse_windows_eventlog_xml(log.body);
```

PR #1046, opened 2026-04-24
Filterx generic dicts

bazsi:
This patch prepares to support generic types as dictionary keys. My plan is to add support for the upcoming tuple type.


sodomelle (approved):

alltilla:
Talked about this IRL: Merge after the 4.25 release.

sodomelle:
I think this should be mentioned in a NEWS entry.

MrAnno:
I'll write the NEWS entry, thanks for noticing.

PR #1047, opened 2026-04-24
Filterx: add tuple datatype

bazsi:
Intended to be very similar to Python's own implementation.
    
Examples:
```
      t = ();  # empty tuple
      t = (1,); # singleton
      t = (1,2); # pair
```

*  The tuple can contain mutable items, but the tuple itself can't be changed. 
* A tuple is hashable (as long as all its items are hashable), so with the combination of #1046, we can use tuples as hash keys, which was my intended use-case.

NOTE: I used AI to get a review of the production code (that I wrote myself) and to generate 75% of the unit tests.
NOTE/2: this depends on a lot of previous PRs, hence the long history. These are the dependent PRs: #925, #1046, #1033

### DIFF
--- a/lib/filterx/filterx-grammar.ym
+++ b/lib/filterx/filterx-grammar.ym
@@ -166,6 +166,10 @@ _assign_location(FilterXExpr *expr, CfgLexer *lexer, CFG_LTYPE *lloc)
 %type <ptr> list_literal
 %type <ptr> list_element
 %type <ptr> list_elements
+%type <ptr> tuple_literal
+%type <ptr> tuple_elements
+%type <ptr> tuple_list
+%type <ptr> tuple_element
 %type <ptr> dpath_element
 %type <ptr> dpath_elements
 %type <ptr> regexp_match
@@ -379,6 +383,7 @@ literal
 	: literal_object			{ FilterXObject *literal = $1; $$ = filterx_literal_new(literal); }
 	| dict_literal
 	| list_literal
+	| tuple_literal
 	;
 
 literal_object
@@ -472,6 +477,28 @@ list_elements
 	;
 
 list_element
+	: expr					{ $$ = filterx_literal_element_new(NULL, $1); }
+	;
+
+tuple_literal
+	: '(' tuple_elements ')'
+          {
+            $$ = filterx_literal_tuple_new($2);
+          }
+	;
+
+tuple_elements
+	: tuple_list				{ $$ = $1; }
+	| 					{ $$ = NULL; }
+	;
+
+tuple_list
+	: tuple_element ',' tuple_list		{ $$ = g_list_prepend($3, $1); }
+	| tuple_element ',' tuple_element	{ $$ = g_list_prepend(g_list_append(NULL, $3), $1); }
+	| tuple_element ',' 			{ $$ = g_list_prepend(NULL, $1); }
+	;
+
+tuple_element
 	: expr					{ $$ = filterx_literal_element_new(NULL, $1); }
 	;
 
--- a/lib/filterx/object-tuple.c
+++ b/lib/filterx/object-tuple.c
@@ -1,0 +1,406 @@
+/*
+ * Copyright (c) 2026 Axoflow
+ * Copyright (c) 2026 Balazs Scheidler <balazs.scheidler@axoflow.com>
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "filterx/object-tuple.h"
+#include "filterx/object-string.h"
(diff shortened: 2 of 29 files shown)


lib/filterx/object-tuple.c
```diff
@@ -0,0 +1,338 @@
...
+{
+  FilterXTuple *self = (FilterXTuple *) s;
+
+  guint acc = _TUPLE_HASH_XXPRIME_5;
+  gsize len = self->array->len;
+
+  for (gsize i = 0; i < len; i++)
+    {
+      FilterXObject *value = g_ptr_array_index(self->array, i);
+      guint hash = filterx_object_hash(value);
```
alltilla, line 172: What ensures that an element has a `hash()` function? Shouldn't we check `filterx_object_hashable()`?

lib/filterx/filterx-grammar.ym
```diff
@@ -476,6 +480,22 @@ list_element
...
 
+tuple_literal
+	: '(' tuple_elements ')'
+          {
+            $$ = filterx_literal_tuple_new($2);
+          }
+	;
+
+tuple_elements
+	: tuple_element ',' tuple_elements	{ $$ = g_list_prepend($3, $1); }
```
alltilla: This forces us to always write a closing colon, which I think is not your original intention.

E.g.:
* this is invalid: `t = (1,2);`
* this is valid: `t = (1,2,);`.

lib/filterx/object-tuple.h
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright (c) 2025 Axoflow
+ * Copyright (c) 2025 Balazs Scheidler <balazs.scheidler@axoflow.com>
```
alltilla: [optional]

Wrong year?

lib/filterx/object-tuple.c
```diff
@@ -0,0 +1,338 @@
...
+      return result;
+    }
+
+  filterx_eval_push_error_info_printf("Failed to create tuple", s,
+                                      "Argument must be a sequence or a string, got: %s",
+                                      filterx_object_get_type_name(arg));
+  return NULL;
+}
+
+FILTERX_DEFINE_TYPE(tuple, FILTERX_TYPE_NAME(sequence),
```
alltilla, line 391: We inherit `add()` and `add_inplace()` from `sequence`. Is this intentional?

lib/filterx/object-tuple.c
```diff
@@ -0,0 +1,338 @@
...
+{
+  guint64 len;
+
+  g_assert(filterx_object_len(sequence, &len));
+
+  FilterXTuple *self = (FilterXTuple *) filterx_tuple_new(len);
+  for (guint64 i = 0; i < len; i++)
+    {
+      FilterXObject *elt = filterx_sequence_get_subscript(sequence, i);
+      g_ptr_array_add(self->array, filterx_object_copy(elt));
```
alltilla: Shouldn't we also check `hashable` in the ctors?

alltilla: Nevermind, we can store unhashable types, that should just make the actual tuple instance unhashable.

alltilla:
> The tuple can contain mutable items, but the tuple itself can't be changed.
> A tuple is hashable (as long as all its items are hashable), so with the combination of https://github.com/axoflow/axosyslog/pull/1046, we can use tuples as hash keys, which was my intended use-case.

As I understand, this means not all tuple instances are hashable, and it is decided by its elements. If it only has hashable elements, the tuple instance is hashable. However, it can have unhashable elements (mapping), but in this case the tuple instance is not hashable and cannot be used as a key.

Is this the intention?

lib/filterx/object-tuple.c
bazsi: Good catch. I fixed this by using the "is_hashable" member if FilterXObject, which is set to FALSE when the first unhashable element is added to the tuple.

lib/filterx/filterx-grammar.ym
bazsi: I've fixed this in the grammar.

lib/filterx/object-tuple.h
bazsi: fixed

lib/filterx/object-tuple.c
bazsi: this still needs to be fixed.

bazsi: I changed add/add_inplace methods to be opt-in for sequences and then implemented the add() function for
the tuple separately. It supports both other tuples and sequences.

lib/filterx/object-tuple.h
```diff
@@ -0,0 +1,74 @@
...
+static inline void
+filterx_tuple_set_subscript(FilterXObject *s, gint64 index, FilterXObject *new_value)
+{
+  FilterXTuple *self = (FilterXTuple *) s;
+
+  if (self->array->len == index)
+    g_ptr_array_set_size(self->array, index + 1);
+  FilterXObject **slot = (FilterXObject **) &g_ptr_array_index(self->array, index);
+  filterx_object_unref(*slot);
+  *slot = filterx_object_ref(new_value);
```
alltilla, line 72: Shouldn't we check `is_hashable` of `new_value`, and follow it if it is not?

lib/filterx/object-tuple.c
```diff
@@ -0,0 +1,399 @@
...
+  return &clone->super.super;
+}
+
+FilterXObject *
+_filterx_tuple_add(FilterXObject *s, FilterXObject *o)
+{
+  FilterXTuple *self = (FilterXTuple *) s;
+  if (o->type != s->type)
+    {
+      return _filterx_tuple_add_sequence(self, o);
```
alltilla, line 275: This crashes with:
```
x = (1,2) + 3;
```
or with:
```
x = (1,2) + "abc";
```

bazsi: fixed. thanks!

lib/filterx/object-tuple.h
bazsi: fixed, thanks

alltilla (approved):

PR #1048, opened 2026-04-27
FilterX JIT compiler - core

MrAnno:
This PR introduces a LLVM IR code generation and a JIT compilation and runtime for FilterX.

### FilterXJIT

`FilterXJIT` is a Facade pattern, encapsulating everything needed for FilterX to be a (JIT) compiled rather than an interpreted language:

- methods to generate LLVM intermediate representation (IR) code with as little boilerplate as possible
- compilation targeting the host platform, running a comprehensive set of optimization passes
- minimal interface to execute the compiled FilterX blocks during pipeline execution
- methods to debug and trace IR generation, JIT execution (TODO)

The current implementation is not "just-in-time compilation" in the classical sense, as all FilterX blocks are compiled and optimized during config initialization, not at runtime.
This can easily be changed in the future.

For now, FilterXJIT is called "JIT" because it uses LLVM JIT APIs to produce in-memory code.

The `filterx-jit(yes/no)` global option can be used to enable/disable the feature.
It is turned on by default, but falls back to interpreted code on hardened platforms.

### Incremental development

Expressions must implement the `compile()` method, which is responsible for generating the corresponding IR code for that expression. You can find 10 examples in #1049.

To allow incremental development of the JIT feature, the `compile()` method is currently optional.
In order to keep the architecture simple, JIT-compiled code can call into interpreted code, but interpreted code can't call back into JIT.

The de-inlined boilerplate in `ffi.c` is temporary.


lib/filterx/filterx-jit.c
```diff
@@ -0,0 +1,359 @@
...
+}
+
+static inline void
+_debug_print_llvm_ir(FilterXJIT *self)
+{
+  gchar *ir_code = LLVMPrintModuleToString(self->mod);
+  printf("%s", ir_code);
+  LLVMDisposeMessage(ir_code);
+}
+#endif
```
alltilla: Is this something we intend to keep?

lib/filterx/jit/jit.c
```diff
@@ -0,0 +1,357 @@
...
+
+  LLVMDisposePassBuilderOptions(options);
+  return err;
+}
+
+LLVMErrorRef
+_optimize_transform(gpointer s, LLVMOrcThreadSafeModuleRef *thr_mod, LLVMOrcMaterializationResponsibilityRef mr)
+{
+  return LLVMOrcThreadSafeModuleWithModuleDo(*thr_mod, _optimize_module, s);
+}
```
alltilla: [optional]

These could be `static`.

stats-sorted.txt
alltilla, line 1: This got added accidentally.

alltilla (commented):
amazing!

stats-sorted.txt
MrAnno: Jesus.. thanks. I've started using `jujutsu`, it seems I'm now pushing weird things, hopefully not my private keys...

lib/filterx/filterx-jit.c
MrAnno: It was useful for debugging, but with #1055, we can probably remove it.

alltilla: I was unsure because they are in an `#if 0` block. I am okay with whatever you decide to do with them.

MrAnno:
The CI won't be green on this PR as it requires new dbld dependencies.
The last patch should be dropped before merging. A green CI is possible if we merge the packaging commit separately. Let me know if you want to proceed that way.

lib/filterx/jit/ffi.c
```diff
@@ -0,0 +1,272 @@
...
+static inline FilterXIRValue
+_emit_call(FilterXJIT *jit, FilterXJITFFICall c, FilterXIRValue *args, unsigned param_count)
+{
+  return LLVMBuildCall2(jit->ir, c.ty, c.fn, args, param_count, "");
+}
+
+FilterXIRValue
+fx_jit_emit_const_ptr(FilterXJIT *jit, gconstpointer p)
+{
+  return LLVMConstIntToPtr(LLVMConstInt(LLVMInt64TypeInContext(jit->ctx), (guintptr) p, FALSE),
```
fernya: Do we support 32 bit architectures?

MrAnno: Yes, we have arm7 images. Nice catch.

bazsi (dismissed):
This is great work. I think we can merge this with the dependency issues, no need to open another pull request. Kudos to @fernya for finding the pointer size mismatch, I would not have found it, but I did a review of the interfaces and have  a pretty good understanding how this works.

I would merge this as soon as any outstanding issues are addressed, we will start using this code in the followups, so any potential issues will only be discovered then.

Also, this whole thing can be disabled both at compile and at runtime.

Approving, but not merging, to see if @MrAnno still wants to fix stuff here.

MrAnno:
Thanks.

@fernya found more things to fix, let me go through them.

lib/filterx/jit/ffi.c
MrAnno: Fixed.

PR #1052, opened 2026-04-28
func-failure-info: restore original intent

MrAnno:
failure_info_meta() was operating on literals from the very beginning, so the patches that changed this behavior back and forth just complicated the original implementation and we lost other literal support (list, scalar), and introduced some error handling leak.

### DIFF
--- a/lib/filterx/func-failure-info.c
+++ b/lib/filterx/func-failure-info.c
@@ -48,6 +48,7 @@ typedef struct FilterXFunctionFailureInfoMeta_
 {
   FilterXFunction super;
   FilterXExpr *metadata_expr;
+  FilterXObject *metadata;
 } FilterXFunctionFailureInfoMeta;
 
 static inline void
@@ -144,9 +145,7 @@ _failure_info_meta_eval(FilterXExpr *s)
   if (!filterx_eval_get_failure_info(context))
     goto exit;
 
-  FilterXObject *metadata_object = filterx_expr_eval_typed(self->metadata_expr);
-  filterx_eval_set_current_frame_meta(context, metadata_object);
-  filterx_object_unref(metadata_object);
+  filterx_eval_set_current_frame_meta(context, self->metadata);
 
 exit:
   return filterx_boolean_new(TRUE);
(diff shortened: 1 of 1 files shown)


lib/filterx/func-failure-info.c
```diff
@@ -311,25 +303,9 @@ _failure_info_meta_init(FilterXExpr *s, GlobalConfig *cfg)
...
-   * I am delaying the retain_object() implementation for mutable objects,
-   * as I have a lot of  PRs lined up that touch the same code.
-   */
-
-  if (!filterx_expr_is_literal(self->metadata_expr))
-    {
-      msg_error("The metadata argument for failure_info_meta() has to be a literal",
-                filterx_expr_format_location_tag(s));
-      return FALSE;
-    }
```
alltilla, line 332: I think we still need this check. Literal containers might not optimize to real literal objects, if there is an expression in the keys or values.

Example config:
```
destination d_stdout {
    stdout(template("failure_info: $failure_info\n"));
};

source s_msg_gen {
    example_msg_generator(num(1));
};

log {
    source(s_msg_gen);
    filterx {
        failure_info_enable(collect_falsy=true);
    };
    if {
        filterx {
            bar = 4;
            failure_info_meta({"foo": bar});
            baz;
        };
    };

    log {
        filterx {
            $failure_info = failure_info();
        };
        destination(d_stdout);
    };
};
```

```
ERROR:../lib/filterx/expr-literal.c:33:filterx_literal_get_value: assertion failed: (filterx_expr_is_literal(s))
Bail out! ERROR:../lib/filterx/expr-literal.c:33:filterx_literal_get_value: assertion failed: (filterx_expr_is_literal(s))
```

sodomelle (dismissed):
If I'm correct, this was changed in https://github.com/axoflow/axosyslog/pull/742/changes/698c74aa6c447e0b869fa918eeac6f7bdda50168#diff-e3269fc27150fef2e2c2758bb2089f7c33929854e037102b9b58cba424766171 and https://github.com/axoflow/axosyslog/pull/874/changes/ec6bde8513f51043581a18a1b1e4b9ad39417185. First one was by mistake, and the second one, is not really an intention.

alltilla (commented):
I found an assert fail scenario from a possible user config.

Otherwise approve for the PR, I am okay with extracting and storing the obj, though I don't find it necessarily a must have.

lib/filterx/func-failure-info.c
MrAnno: Thanks. I misunderstood what `filterx_expr_is_literal_dict()` and `filterx_expr_is_literal_list()` mean.

MrAnno:
The object extraction is not a functional change, but the literal (list, scalar) support and the disappeared `filterx_expr_unref()` can be useful.

The comment about `filterx_eval_retain_object()` is not relevant in my opinion because of the misunderstanding that happened in https://github.com/axoflow/axosyslog/commit/698c74aa6c447e0b869fa918eeac6f7bdda50168#diff-e3269fc27150fef2e2c2758bb2089f7c33929854e037102b9b58cba424766171.

alltilla (approved):

PR #1056, opened 2026-05-04
[WIP] Vectorize CSV parsing

fernya:
This PR adds vectorization to some CSV parsing cases.

You may opt-out with the usage of `-DENABLE_MANUAL_SIMD=Off` (CMake) or `--enable-manual-simd=no` (autotools) flags.
The build system then checks the available and implemented SIMD instructions sets and compiles the appropriate code.

At runtime the SIMD blocks may be disabled if the CPU doesn't support them.
You may test this with the following command: `EXTRA_ARGS="--run-under=qemu --qemu-cpu=qemu64 -k panos" make light-check`
(qemu-user package needs to be installed on Ubuntu)

- [x] Remove `have_simd` branching from hot path
- [x] Move all `#ifdefs` out from `csv-scanner.c`


mitzkia:
Thanks for the Light related update. I can approve that.

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,427 @@
...
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "csv-scanner-simd.h"
+#include "compat/glib.h"
+
+/* Per-function target attribute lets GCC and clang emit AVX2 code */
```
MrAnno, line 27: The best comment is the one that is not needed. Clean, self-explanatory code can spare a lot of comments; which has the advantage of not requiring constant updates as the code changes (it won't mislead the reader once it goes stale).

"Comments should not duplicate the code" - based on this "rule", I'll mark the comments which I found unnecessary with `*c`.

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,427 @@
...
+#define AVX2_FN __attribute__((target("avx2")))
+
+/* Parser-level chunk: two AVX2 ymm loads (32 B each) → one 64-bit hit
+ * bitmap that we iterate with __builtin_ctzll.  Not a SIMD-imposed
+ * constant; the 64 falls out of "two movemasks fit in a uint64_t". */
```
MrAnno, line 32: *c

I don't really understand what this means. Looks like a generated comment, with a unicode `→`  symbol. I'm sure we can explain this more clearly with code or with a simpler comment near to the code that uses this constant.

I'd be less confused without this comment.

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,427 @@
...
+#define SIMD_PARSER_CHUNK_BYTES 64
+
+/* Per-chunk hit bitmaps: bit i is set iff chunk[i] matches the class.
+ * `quotes` collapses '"' and '\'' into a single mask; `spaces` does the
+ * same for ' ' and '\t'. */
```
MrAnno, line 37: I think the non-redundant valuable information here would be stating that these are bitmaps for finding the important characters within a given chunk.

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,427 @@
...
+  const __m256i comma  = _mm256_set1_epi8(',');
+  const __m256i quote  = _mm256_set1_epi8('"');
+  const __m256i squote = _mm256_set1_epi8('\'');
+  const __m256i space  = _mm256_set1_epi8(' ');
+  const __m256i tab    = _mm256_set1_epi8('\t');
+
+  __m256i lo = _mm256_loadu_si256((const __m256i *) chunk);
+  __m256i hi = _mm256_loadu_si256((const __m256i *) (chunk + SIMD_PARSER_CHUNK_BYTES / 2));
+
+  /* commas: single character */
```
MrAnno, line 70: *c 

Move the declaration of `comma` next to this call instead.

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,427 @@
...
+  self->commas = _combine_mm(_mm256_cmpeq_epi8(lo, comma),
+                             _mm256_cmpeq_epi8(hi, comma));
+
+  /* quotes: '"' | '\''  - fuse the OR in vector domain, movemask once per lane */
```
MrAnno, line 74: *c 

Move the declaration of `(s)quote` next to this call instead.

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,427 @@
...
+  __m256i q_lo = _mm256_or_si256(_mm256_cmpeq_epi8(lo, quote),
+                                 _mm256_cmpeq_epi8(lo, squote));
+  __m256i q_hi = _mm256_or_si256(_mm256_cmpeq_epi8(hi, quote),
+                                 _mm256_cmpeq_epi8(hi, squote));
+  self->quotes = _combine_mm(q_lo, q_hi);
+
+  /* spaces: ' ' | '\t' - same trick */
```
MrAnno, line 81: *c

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,427 @@
...
+
+      if (!_simd_parse_consume_hits(input, len, i * SIMD_PARSER_CHUNK_BYTES, interesting,
+                                    &quote_state, &field_start, fields_array))
+        {
+          g_array_set_size(fields_array, 0);
+          return FALSE;
+        }
+    }
+
+  /* Phase 2: tail (< 64 bytes) */
```
MrAnno, line 169: *c 

This is typically the comment that suggests the below code should be extracted into a static inline function and given a meaningful name.

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,427 @@
...
+gint
+csv_simd_find_delimiter_or_quote(const char *start, gsize max_len, gchar delimiter, gchar quote_char)
+{
+  CSVSimdFindResult result;
+  csv_simd_find_either(start, max_len, delimiter, quote_char, &result);
+  return result.offset;
+}
+
+/* Trim leading whitespace using SIMD scan (32 bytes at a time).
+ * Returns the new (inclusive) start offset. */
```
MrAnno, line 374: *c

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,427 @@
...
+  while (start_ofs < end_ofs && (input[start_ofs] == ' ' || input[start_ofs] == '\t'))
+    start_ofs++;
+  return start_ofs;
+}
+
+/* Trim trailing whitespace using SIMD scan (32 bytes at a time).
+ * Returns the new (exclusive) end offset.
+ *
+ * Operates entirely on offsets - never forms `input + start_ofs - 1`,
+ * which would be UB when start_ofs == 0. */
```
MrAnno, line 403: I don't understand the second sentence, probably `*c`.

lib/scanner/csv-scanner/csv-scanner-simd.h
```diff
@@ -0,0 +1,81 @@
...
+#ifndef CSVSCANNER_SIMD_H_INCLUDED
+#define CSVSCANNER_SIMD_H_INCLUDED
+
+#include "syslog-ng.h"
+#include <stdint.h>
+#include <immintrin.h>
+
+/************************************************************************
+ * SIMD primitives
+ ************************************************************************/
```
MrAnno, line 33: *c

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,427 @@
...
+static inline gboolean
+_find_in_chunk(const char *start, gsize offset, const __m256i *needles, int n,
+               CSVSimdFindResult *result)
+{
+  __m256i lo = _mm256_loadu_si256((const __m256i *) (start + offset));
+  __m256i hi = _mm256_loadu_si256((const __m256i *) (start + offset + SIMD_PARSER_CHUNK_BYTES / 2));
+
+  /* Per-needle cmpeq, kept around so we can produce per-needle bitmaps
+   * cheaply if there's a hit.  ORing in vector domain plus testz keeps
+   * the hot no-hit path off port 5 (movemask). */
```
MrAnno, line 276: port 5?

lib/scanner/csv-scanner/csv-scanner-simd.h
```diff
@@ -0,0 +1,81 @@
...
+} CSVFieldOfs;
+
+gboolean csv_simd_parse(const gchar *input, gsize len, GArray *fields_array);
+
+/* Find the first occurrence of either c1, c2, c3, or c4 in a string.
+ * Returns a structure with offset and which char was found or -1 if not found. */
+typedef struct
+{
+  gint offset;  /* offset of first match, or -1 if not found */
+  gint which;   /* 0 if c1 found first, 1 if c2, 2 if c3, 3 if c4 */
```
MrAnno, line 49: 3 * `*c`

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+  if (offset < max_len)
+    {
+      _scalar_search_impl(start, offset, max_len - offset, chars, n, result);
+      return;
+    }
+
+  result->offset = -1;
+  result->which = 0;
```
MrAnno, line 323: `which` should probably be `-1` on all implementations' error path.

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,427 @@
...
+          if (ch == chars[j])
+            {
+              result->offset = offset + i;
+              result->which = j;
+              return;
+            }
+        }
+    }
+  result->offset = -1;
+  result->which = 0;
```
MrAnno, line 240: `which` should probably be `-1` on all implementations' error path.

lib/scanner/csv-scanner/csv-scanner-simd.h
```diff
@@ -0,0 +1,81 @@
...
+ * This is optimized for quickly skipping normal text when scanning CSV fields. */
+gint csv_simd_find_delimiter_or_quote(const char *start, gsize max_len, gchar delimiter, gchar quote_char);
+
+/* Find delimiter, quote, or whitespace in a single pass.
+ * Returns: which type was found (0=delimiter, 1=quote, 2=space/tab) and offset.
+ * Optimizes whitespace-stripping configurations by doing all classification at once. */
+typedef struct
+{
+  gint offset;  /* offset of first match, or -1 if not found */
+  gint which;   /* 0=delimiter, 1=quote, 2=whitespace */
```
MrAnno, line 70: 3 * `*c`

`0=delimiter, 1=quote, 2=whitespace` sounds like an enum to me.

lib/scanner/csv-scanner/csv-scanner-simd.h
```diff
@@ -0,0 +1,95 @@
...
+} CSVSimdFindTripleResult;
+
+void csv_simd_find_delim_quote_space(const char *start, gsize max_len, gchar delimiter, gchar quote_char,
```
MrAnno, line 85: This function is never used, which means `CSVSimdFindTripleResult` is unused as well.

These APIs seem to be the clean ones made specifically for CSV, so it would be nice if we could use them instead of `csv_simd_find_either*`.

lib/scanner/csv-scanner/csv-scanner-simd.h
```diff
@@ -0,0 +1,81 @@
...
+gint csv_simd_find_delimiter_or_quote(const char *start, gsize max_len, gchar delimiter, gchar quote_char);
+
+/* Find delimiter, quote, or whitespace in a single pass.
+ * Returns: which type was found (0=delimiter, 1=quote, 2=space/tab) and offset.
+ * Optimizes whitespace-stripping configurations by doing all classification at once. */
+typedef struct
+{
+  gint offset;  /* offset of first match, or -1 if not found */
+  gint which;   /* 0=delimiter, 1=quote, 2=whitespace */
+} CSVSimdFindTripleResult;
```
MrAnno, line 71: What do you think about `CSVSeparatorLocation`?

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,427 @@
...
+ *   - precompute cmpeq vectors per needle for both 32-byte halves
+ *   - OR them all together and use _mm256_testz to skip blank chunks
+ *     in vector domain (no movemask), then only on a hit pay the
+ *     movemask cost to figure out which needle/where.
+ * --------------------------------------------------------------------- */
+
+/* Scalar tail search for regions smaller than SIMD_PARSER_CHUNK_BYTES. */
+AVX2_FN
+static inline void
+_scalar_search_impl(const char *start, gsize offset, gsize remaining, const gchar *chars, gsize num_chars,
```
MrAnno, line 223: *c

Let's name the function meaningfully. The comment above actually shows what the name should be: `_slow_tail_search()` or anything similar (without the word "impl").

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,427 @@
...
+csv_simd_find_either4(const char *start, gsize max_len, gchar c1, gchar c2, gchar c3, gchar c4,
+                      CSVSimdFindResult *result)
+{
+  const gchar chars[4] = { c1, c2, c3, c4 };
+  _find_n(start, max_len, chars, 4, result);
+}
+
+AVX2_FN
+void
+csv_simd_find_delim_quote_space(const char *start, gsize max_len, gchar delimiter, gchar quote_char,
```
MrAnno, line 353: dead code

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,427 @@
...
+  csv_simd_find_either4(start, max_len, delimiter, quote_char, ' ', '\t', &sr);
+
+  result->offset = sr.offset;
+  /* Normalize: which=0 (delim), 1 (quote), 2 (space), 3 (tab) -> 0, 1, 2, 2 */
+  result->which = (sr.which == 3) ? 2 : sr.which;
+}
+
+AVX2_FN
+gint
+csv_simd_find_delimiter_or_quote(const char *start, gsize max_len, gchar delimiter, gchar quote_char)
```
MrAnno, line 366: dead code

lib/scanner/csv-scanner/csv-scanner.c
```diff
@@ -265,9 +266,15 @@ static void
...
-  const gchar *nexthop = MIN(backslash, quote);
+  CSVSimdFindResult result;
+  gsize remaining = self->input_end - self->src;
+  csv_simd_find_either(self->src, remaining, '\\', self->current_quote, &result);
+  const gchar *nexthop;
+
+  if (result.offset >= 0)
+    nexthop = self->src + result.offset;
+  else
+    nexthop = self->src + remaining;  /* no match: scan to end of input */
```
MrAnno, line 277: This breaks compatibility on platforms where SIMD operations are not available.

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,614 @@
...
+  return -1;
+}
+
+gint32
+csv_simd_trim_trailing_whitespace(const gchar *input, gint32 start_ofs, gint32 end_ofs)
+{
+  return -1;
+}
+
+#else
```
MrAnno, line 528: Optional:

I think it's possible to refactor the SIMD code's entry points so that we use only 3-4 short calls that either perform the optimization or signal an error, letting the scanner fall back to the slow path when calling into the SIMD methods from the CSV scanner side. That way, we would no longer need some many NOOP functions, and bugs like the one I found in the previous comment would be easier to catch.

For example:
- Extracting the SIMD-related logic from `_parse_value_with_whitespace_and_delimiter()` into a separate function and moving it to `csv-scanner-simd.c`.
- `_scan_next_from_fast_path`, `_trim_fast_path_fields_whitespace`, and `_can_use_simd_fast_path` should also go into `csv-scanner-simd.c`.

The only code remaining inside `csv-scanner.c` would be the check for SIMD availability and short one-line calls into the SIMD-optimized paths. Passing `CSVScanner *self` to these functions is fine. If we want to separate the generic SIMD code that doesn't operate on `CSVScanner`, we can add a separate `simd.c` unit, which can be built conditionally as a whole.

lib/scanner/csv-scanner/csv-scanner-simd.h
```diff
@@ -0,0 +1,95 @@
...
+
+#ifdef SYSLOG_NG_MANUAL_AVX2
+
+#include <immintrin.h>
+
+#endif /* SYSLOG_NG_MANUAL_AVX2 */
+
+#ifdef SYSLOG_NG_MANUAL_NEON
+
+#include ???
```
MrAnno, line 38: `#error "NEON instructions are currently not supported"`

Or even better, we should get rid of all `SYSLOG_NG_MANUAL_NEON` branches in this PR so that we can compile and use the slow path until we have an actual working implementation.

A TODO comment should be enough.

MrAnno (commented):
This PR is not completely ready for review, as it contains a huge amount of probably LLM-generated noise, confusion, and dead code. Let's make some cleanup first.

My first review round is about this cleanup, I'll do a separate review round which will be much more meaningful and I'll concentrate on the actual implementation and perf aspects there.

Sorry for the short and dry comments. They may sound harsh, but I just wanted to get through the LLM-generated stuff as fast as possible and concentrate on the important things afterwards (where I'll be much more thorough and polite, of course).

lib/scanner/csv-scanner/csv-scanner.c
```diff
@@ -489,6 +498,40 @@ _parse_value_with_whitespace_and_delimiter(CSVScanner *self)
   gboolean nonliteral_input = FALSE;
 
   self->current_value_start_ofs = self->src - self->input;
+
+  /* SIMD fast path for simple unquoted values with default delimiter */
+  if (csv_is_simd_available() && !self->current_quote && !self->options->string_delimiters && !self->options->delimiters &&
+      self->options->dialect == CSV_SCANNER_ESCAPE_NONE && !(self->options->flags & CSV_SCANNER_STRIP_WHITESPACE))
```
MrAnno, line 504: This can be hidden inside a `csv_scanner_can_use_simd_fast_path()` or anything similar.

lib/scanner/csv-scanner/csv-scanner.c
```diff
@@ -489,6 +498,40 @@ _parse_value_with_whitespace_and_delimiter(CSVScanner *self)
...
+          self->src += result.offset;
+          if (*self->src == DEFAULT_DELIM_CHAR)
+            {
+              self->src++;
+              return;
+            }
+          /* Fall through to handle quote */
+          self->current_quote = *self->src;
+          self->src++;
+          goto continuation;
```
MrAnno, line 524: `_parse_unquoted_literal_characters()` detects escaping on `DEFAULT_DELIM_CHAR`, which seems to be missing from here.

----

What happens with `foo"bar,baz`? (Klaudia found this one)
The old code didn't allow starting a quote in the middle of the value, the new one does, so it can mess things up.

lib/scanner/csv-scanner/csv-scanner.c
```diff
@@ -489,6 +498,40 @@ _parse_value_with_whitespace_and_delimiter(CSVScanner *self)
...
   self->current_value_start_ofs = self->src - self->input;
+
+  /* SIMD fast path for simple unquoted values with default delimiter */
+  if (csv_is_simd_available() && !self->current_quote && !self->options->string_delimiters && !self->options->delimiters &&
+      self->options->dialect == CSV_SCANNER_ESCAPE_NONE && !(self->options->flags & CSV_SCANNER_STRIP_WHITESPACE))
+    {
+      gsize remaining = self->input_end - self->src;
+      CSVSimdFindResult result;
+      /* Find delimiter, double-quote, or single-quote in a single SIMD pass */
+      csv_simd_find_either3(self->src, remaining, DEFAULT_DELIM_CHAR, '"', '\'', &result);
```
MrAnno, line 509: `quotes_start` and `quotes_end` are configurable. Shouldn't we add this as a condition to running the fast path?

lib/scanner/csv-scanner/csv-scanner.c
```diff
@@ -489,6 +498,40 @@ _parse_value_with_whitespace_and_delimiter(CSVScanner *self)
...
+            }
+          /* Fall through to handle quote */
+          self->current_quote = *self->src;
+          self->src++;
+          goto continuation;
+        }
+      else if (result.offset == 0)
+        {
+          /* Delimiter at start */
+          self->src++;
```
MrAnno, line 529: We stop at quotes, not just delimiters. I think we shouldn't skip this character unconditionally.

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,614 @@
...
+  return TRUE;
+}
+
+/* Unified n-needle linear scan; n in [1, 4]. */
+AVX2_FN
+static inline void
+_find_n(const char *start, gsize max_len, const gchar *chars, int n,
+        CSVSimdFindResult *result)
+{
+  if (G_LIKELY(cpu_supports_simd))
```
MrAnno, line 324: I think the runtime SIMD detection should be done only once outside of this unit and we should fall back to the original CSV scanner implementation, not to this partial fallback (_scalar_search_impl).

To summarize it:
We shouldn't see multiple `cpu_supports_simd` and `csv_is_simd_available()` checks, we should disable the whole thing in the scanner.

lib/scanner/csv-scanner/csv-scanner.c
```diff
@@ -651,21 +832,38 @@ csv_scanner_init(CSVScanner *scanner, CSVScannerOptions *options, const gchar *i
...
   scanner->input = scanner->src = input;
+  scanner->input_end = input + strlen(input);
   scanner->current_value = scratch_buffers_alloc();
   scanner->current_value_start_pos = NULL;
   scanner->current_column = 0;
   scanner->options = options;
+
+  /* Pre-allocate fast_path_fields for at least 16 columns to avoid early reallocations */
+  scanner->fast_path_fields = g_array_sized_new(FALSE, FALSE, sizeof(CSVFieldOfs), 16);
+  csv_detect_cpu_features();
```
MrAnno, line 843: Optional: You started caching the CPU feature check, which is a good idea.
This means we could move the detection call into a `csv_scanner_global_init()` function, which you could call from `app_startup()`. That runs only once, `csv_scanner_init()` is called whenever we want to scan a new CSV string.

lib/scanner/csv-scanner/csv-scanner-simd.c
```diff
@@ -0,0 +1,614 @@
...
+    }
+
+  result->offset = -1;
+  result->which = 0;
+
+}
+
+AVX2_FN
+void
+csv_simd_find_either(const char *start, gsize max_len, gchar c1, gchar c2, CSVSimdFindResult *result)
```
MrAnno, line 358: if `max_len` is `gsize`, then all offsets returned should also be `gsize`.

lib/scanner/csv-scanner/csv-scanner.c
```diff
@@ -617,12 +660,150 @@ csv_scanner_append_rest(CSVScanner *self)
...
+ *   - no `null_value` translation, no escape dialect,
+ *     no GREEDY
+ * Additionally bails if the record contains `'` anywhere, because the
+ * classifier only tracks `"` in its quotes mask.
+ *
+ * On success, one SIMD pass builds an array of (start, end, quoted) field
+ * offsets; subsequent scan_next calls dequeue from that array.  Bails out
+ * to the generic path on cases the classifier's view of quoting can't
+ * express (unterminated quoted field, literal content after closing quote
+ * like "foo"bar). */
```
MrAnno, line 679: This is not true.

PR #1059, opened 2026-05-07
Permissive mode for app-parser/app-transform and a bugfix

MrAnno:
When one wants to use the app-parser() but allow all messages to pass
through even with 0 matches, the parser has to be wrapped in an if
block.

```
if {
  parser {
    app-parser(topic(t) filterx-app-variable(meta.app));
  };
};
```

With the new permissive(yes) mode, this is no longer necessary,
and the generated config is simpler (we basically added an if block
to remove the effect of the inner else block), therefore the built
pipeline is also less complex.

```diff
-if {
 parser {
   channel {
     if {
         filterx {
           1;
         };
     }
     elif {
         filterx {
           2;
         };
-      }
-      else {
-          filterx { false; };
-      };
     };
   };
 };
```

`permissive(yes)` also allows FilterX errors to happen inside the topic
codes, so there will be no message loss.

The default for app-parser is `permissive(no)`.

----

Transformations are by default permissive, meaning that a missing `filterx-app-variable`, an error, or accidental filtering rule would not cause messages being dropped.

This was true for transformations containing at least one non-filterx block. For filterx-only configs, this permissive behavior was violated, so I fixed it.

`permissive(yes/no)` can be set for transformations too, but the default here is `yes`.

### DIFF
--- a/modules/appmodel/app-parser-generator.c
+++ b/modules/appmodel/app-parser-generator.c
@@ -53,7 +53,7 @@ _parse_topic_arg(AppParserGenerator *self, CfgArgs *args, const gchar *reference
   return TRUE;
 }
 
-static gboolean
+static void
 _parse_allow_overlaps(AppParserGenerator *self, CfgArgs *args, const gchar *reference)
 {
   const gchar *v = cfg_args_get(args, "allow-overlaps");
@@ -61,10 +61,9 @@ _parse_allow_overlaps(AppParserGenerator *self, CfgArgs *args, const gchar *refe
     self->allow_overlaps = cfg_process_yesno(v);
   else
     self->allow_overlaps = FALSE;
-  return TRUE;
 }
 
-static gboolean
+static void
 _parse_permissive(AppParserGenerator *self, CfgArgs *args, const gchar *reference)
 {
   const gchar *v = cfg_args_get(args, "permissive");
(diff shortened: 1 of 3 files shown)


modules/appmodel/app-parser-generator.c
```diff
@@ -80,6 +94,9 @@ app_parser_generator_parse_arguments(AppObjectGenerator *s, CfgArgs *args, const
   if (!_parse_allow_overlaps(self, args, reference))
     return FALSE;
 
+  if (!_parse_permissive(self, args, reference))
```
sodomelle: This can't happen, as `_parse_permissive` doesn't return `FALSE` no matter what.

modules/appmodel/app-transform-generator.c
```diff
@@ -83,6 +97,9 @@ app_transform_generator_parse_arguments(AppObjectGenerator *s, CfgArgs *args, co
   if (!_parse_transforms_arg(self, args, reference))
     return FALSE;
 
+  if (!_parse_permissive(self, args, reference))
```
sodomelle: Same.

modules/appmodel/app-transform-generator.c
```diff
@@ -313,11 +330,22 @@ app_transform_generate_config(AppObjectGenerator *s, GlobalConfig *cfg, GString
...
                              "            };\n"
                              "        };\n"
-                             "    };\n",
+                             "    }",
                              self->filterx_app_variable,
                              block->str);
     }
 
+  if (self->permissive)
+    g_string_append_printf(result, ";\n");
```
sodomelle: `g_string_append` is enough.

sodomelle (commented):
Some minor finidngs

modules/appmodel/app-parser-generator.c
MrAnno: It's an unnecessary pattern that is followed in the appmodel module.

modules/appmodel/app-transform-generator.c
MrAnno: Fixed.

sodomelle (dismissed):
Looks good to me. I'll merge it this afternoon if nobody else comments on it.

sodomelle (approved):

PR #1060, opened 2026-05-07
Autoconf monolithic build should not use PIC compilation mode

bazsi:
The whole idea of linking syslog-ng statically in #1016 was to avoid the calls to _tls_get_addr() when using thread local storage. Unfortunately, in one of the last rebase iteration of that PR, this was lost.

This PR now explicitly tells libtool to build objects in non-pic mode, when monolithic build was selected. This should get back roughly 10% of the cycles as shown by callgrind. 3% is about the overhead of tls_get_addr() the rest is just dynamic linker overhead, we also got rid of (this is just a one-of cost)


MrAnno (dismissed):
A rebase will be needed.

modules/grpc/protos/Makefile.am
```diff
@@ -156,8 +160,13 @@ if ENABLE_EXTRA_WARNINGS
 modules_grpc_protos_libgrpc_protos_la_CXXFLAGS += -Wno-switch-default
 endif
 
-modules_grpc_protos_libgrpc_protos_la_LIBADD = $(MODULE_DEPS_LIBS) $(PROTOBUF_LIBS) $(GRPCPP_LIBS)
+if ENABLE_BUILTIN_MODULES
+modules_grpc_protos_libgrpc_protos_la_LDFLAGS = -static
```
MrAnno, line 164: I think we are just missing `modules_grpc_protos_libgrpc_protos_la_LDFLAGS = $(MODULE_LDFLAGS)` in general.

Maybe we can skip `noinst_LTLIBRARIES` once it's fixed?

MrAnno (approved):

modules/grpc/protos/Makefile.am
bazsi: this lib is not a module and using MODULE_LDFLAGS would break it in case of a dynamic build.

PR #1061, opened 2026-05-08
FilterX JIT compiler - core

MrAnno:
This is #1048 reopened.

I accidentally deleted the branch using `jj`, GitHub doesn't let me reopen the PR.

----

This PR introduces a LLVM IR code generation and a JIT compilation and runtime for FilterX.

### FilterXJIT

`FilterXJIT` is a Facade pattern, encapsulating everything needed for FilterX to be a (JIT) compiled rather than an interpreted language:

- methods to generate LLVM intermediate representation (IR) code with as little boilerplate as possible
- compilation targeting the host platform, running a comprehensive set of optimization passes
- minimal interface to execute the compiled FilterX blocks during pipeline execution
- methods to debug and trace IR generation, JIT execution (TODO)

The current implementation is not "just-in-time compilation" in the classical sense, as all FilterX blocks are compiled and optimized during config initialization, not at runtime.
This can easily be changed in the future.

For now, FilterXJIT is called "JIT" because it uses LLVM JIT APIs to produce in-memory code.

The `filterx-jit(yes/no)` global option can be used to enable/disable the feature.
It is turned on by default, but falls back to interpreted code on hardened platforms.

### Incremental development

Expressions must implement the `compile()` method, which is responsible for generating the corresponding IR code for that expression. You can find 10 examples in #1049.

To allow incremental development of the JIT feature, the `compile()` method is currently optional.
In order to keep the architecture simple, JIT-compiled code can call into interpreted code, but interpreted code can't call back into JIT.

The de-inlined boilerplate in `ffi.c` is temporary.


MrAnno:
The CI is red because the LLVM dependency will only be installed in `devshell` after merging the packaging change.

The last state of #1048 shows that all tests have passed.

lib/cfg.c
```diff
@@ -128,6 +128,13 @@ _invoke_module_init(gchar *key, ModuleConfig *mc, gpointer *args)
     *result = FALSE;
 }
 
+static void
+_invoke_module_post_cfg_init(gchar *key, ModuleConfig *mc, gpointer *c)
```
fernya: It should be `_invoke_module_post_cfg_init(gchar *key, ModuleConfig *mc, gpointer c)`
Or at least I guess based on the cast int line 134

MrAnno: Indeed, thanks.

fernya (approved):
LGTM, and depart for new voyages! 🚢

PR #1064, opened 2026-05-11
Logproto spoof source

bazsi:
This PR moves the implementation of the spoof-source() feature so it becomes a LogProtoClient, e.g. it stops bypassing the LogWriter machinery. This has a number of benefits:

1) receiving and sending of message become separate threads, allowing more scale
2) the mutex protecting the libnet context is eliminated (another scalability improvement)
3) the queue metrics light up (until now, spoofed messages are not counted in stats)
4) the code is much nicer

The majority of the PR has been  generated by claude as indicated in the commits. I reviewed it and tested it with ipv4. I also compared the ipv4/ipv6 packet construction against the old implementation and they match up.

### DIFF
--- a/modules/afsocket/afinet-dest.c
+++ b/modules/afsocket/afinet-dest.c
@@ -30,6 +30,9 @@
 #include "afsocket-signals.h"
 #include "transport/transport-tls.h"
 #include "transport/transport-stack.h"
+#if SYSLOG_NG_ENABLE_SPOOF_SOURCE
+#include "logproto-spoof-source-client.h"
+#endif
 
 #include <sys/types.h>
 #include <sys/socket.h>
@@ -38,23 +41,6 @@
 #include <netinet/in.h>
 #include <stdlib.h>
 #include <string.h>
-
-#ifdef _GNU_SOURCE
-#  define _GNU_SOURCE_DEFINED 1
-#  undef _GNU_SOURCE
-#endif
-
-#if SYSLOG_NG_ENABLE_SPOOF_SOURCE
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wundef"
-#include <libnet.h>
-#pragma GCC diagnostic pop
-#endif
-
-#if _GNU_SOURCE_DEFINED
-#  undef _GNU_SOURCE
-#  define _GNU_SOURCE 1
-#endif
 
 static const gint MAX_UDP_PAYLOAD_SIZE = 65507;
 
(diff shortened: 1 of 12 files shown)


bazsi:
This is flaky because #1179

modules/afsocket/logproto-spoof-source-client.c
```diff
@@ -0,0 +1,264 @@
...
+  *consumed = FALSE;
+
+  gsize effective_len = MIN(msg_len, (gsize) self->max_msglen);
+  gboolean success = _construct_packet(self, logmsg, msg, effective_len)
+                     && (libnet_write(self->lnet_ctx) >= 0);
+
+  if (!success)
+    {
+      msg_error("Error sending raw frame, falling back to normal UDP",
+                evt_tag_str("error", libnet_geterror(self->lnet_ctx)));
```
alltilla: I'm afraid this can flood if there is a permanent error.

modules/afsocket/logproto-spoof-source-client.c
```diff
@@ -0,0 +1,264 @@
...
+      return log_proto_text_client_post_method(s, logmsg, msg, msg_len, consumed);
+    }
+
+  *consumed = TRUE;
+  log_proto_client_msg_ack(s, 1);
```
alltilla, line 216: Don't we need to free `msg` here?

modules/afsocket/afinet-dest.c
```diff
@@ -474,198 +428,34 @@ afinet_dd_init(LogPipe *s)
...
-  gboolean success = afinet_dd_construct_ip_packet(self, msg, self->lnet_buffer);
-
-  if (!success)
-    goto finish;
-
-  success = libnet_write(self->lnet_ctx) >= 0;
+  AFInetDestDriver *self = (AFInetDestDriver *) s;
 
-  if (!success)
+  if (self->spoof_source)
```
alltilla: I think we lost one check here:
```c
if (self->super.transport_mapper->sock_type == SOCK_DGRAM)
```

Currently we allow spoof-source with transport TCP or even TLS, and send UDP packets on the happy path (`_is_message_spoofable()` -> `TRUE`).

modules/afsocket/logproto-spoof-source-client.c
```diff
@@ -0,0 +1,264 @@
+/*
```
alltilla, line 1: I think we are missing some `SYSLOG_NG_ENABLE_SPOOF_SOURCE` checks in this file.

modules/afsocket/logproto-spoof-source-client.c
alltilla: Talked with Bazsi IRL, I fixed this on his behalf.

modules/afsocket/logproto-spoof-source-client.c
alltilla: Talked with Bazsi IRL, I fixed this on his behalf.

modules/afsocket/afinet-dest.c
alltilla: Talked with Bazsi IRL, I fixed this on his behalf.

modules/afsocket/logproto-spoof-source-client.c
alltilla: Talked with Bazsi IRL, I fixed this on his behalf.

alltilla (approved):

PR #1065, opened 2026-05-12
Add FilterX JIT toggle to Light tests

fernya:
During the incubation phase of the FilterX JIT feature we should test interpreted and JIT code paths alongside each other.

The test framework implies that the appropriate keyword is known to the axosyslog instance (ie. FilterX JIT merged)
The test framework works well without compiled in JIT functionality, in that case the JIT is never enabled in the config.
The change adds the appropriate JIT enable/disable flags to the raw and rendered logs.

### DIFF
--- a/tests/light/src/axosyslog_light/syslog_ng_config/syslog_ng_config.py
+++ b/tests/light/src/axosyslog_light/syslog_ng_config/syslog_ng_config.py
@@ -93,6 +93,7 @@ class SyslogNgConfig(object):
         self._prometheus_stats_handler = prometheus_stats_handler
         self.__implicit_statement_groups: typing.Dict[int, StatementGroup] = {}
         self.teardown = teardown
+        self.filterx_jit_enabled = None
 
     stringify = staticmethod(stringify)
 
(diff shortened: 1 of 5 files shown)


MrAnno:
Sorry, I tried to make the `(push)` variant of the CI jobs disappear.

MrAnno (dismissed):
There is a style-check issue (just a newline), otherwise 👍🏻

mitzkia (dismissed):
Thanks for the PR. Approved

MrAnno (dismissed):

tests/light/src/axosyslog_light/syslog_ng_config/syslog_ng_config.py
```diff
@@ -78,6 +78,7 @@ def __init__(
         stats_handler: LegacyStatsHandler,
         prometheus_stats_handler: PrometheusStatsHandler,
         teardown,
+        filterx_jit_enabled: bool,
```
alltilla: I'd rather not break the ctor of `SyslogNgConfig` with this functionality. `light` is a public python library and it is used in other projects, this would introduce a breaking change.

We could keep the ctor as is and just do this instead:
```diff
@pytest.fixture
def config(request, syslog_ng, teardown) -> SyslogNgConfig:
-    jit_enabled = request.getfixturevalue("jit_available") and not request.config.getoption("--disable-filterx-jit")
-    return syslog_ng.create_config(request.getfixturevalue("config_version"), teardown, jit_enabled)
+    cfg = syslog_ng.create_config(request.getfixturevalue("config_version"))
+    cfg.filterx_jit_enabled = request.getfixturevalue("jit_available") and not request.config.getoption("--disable-filterx-jit")
+    return cfg
```

What do you think?

mitzkia: Thanks, I agreed with this.

fernya: Been there, done that, thanks!

MrAnno (approved):

PR #1068, opened 2026-05-12
Ubuntu 26.04 and Fedora 44 packages

MrAnno:
Depends on #1067


OverOrion (dismissed):
Thank you! Should we remove the old Docker image files from dbld as well?

MrAnno:
Good idea. I removed the EOL platforms, will remove the images too.

OverOrion (approved):

PR #1069, opened 2026-05-13
`arrow-flight()` destination

alltilla:
https://arrow.apache.org/docs/format/Flight.html

Options:
  * `url()`: Flight endpoint
  * `path()`: descriptor path, templatable so a single destination can route to many tables
  * `schema()`: one `"name" TYPE => template` entry per column

Column types supported in `schema()`:
  * `STRING`
  * `INT64` (alias `INTEGER`)
  * `DOUBLE` (alias `FLOAT64`)
  * `BOOL` (alias `BOOLEAN`)
  * `TIMESTAMP`
  * `MAP(STRING, STRING)`

Example configuration:

```
destination d_arrow {
  arrow-flight(
    url("grpc://flight.example.com:8815")
    path("events.${HOST}")
    schema(
      "ts"       TIMESTAMP => "$UNIXTIME"
      "host"     STRING    => "$HOST"
      "program"  STRING    => "$PROGRAM"
      "severity" INT64     => "$LEVEL_NUM"
      "msg"      STRING    => "$MSG"
    )
    batch-lines(1000)
    batch-bytes(1048576)
    batch-timeout(5000)
    workers(4)
    worker-partition-key("${HOST}")
  );
};
```


modules/arrow-flight/arrow-flight-worker.cpp
```diff
@@ -139,6 +142,56 @@ DestinationWorker::create_builders()
...
+bool
+DestinationWorker::append_value(arrow::ArrayBuilder *builder, const std::shared_ptr<arrow::DataType> &type,
+                                const char *str, gssize len)
+{
+  arrow::Status s;
+
+  switch (type->id())
+    {
+    case arrow::Type::STRING:
+      s = static_cast<arrow::StringBuilder *>(builder)->Append(str, (int32_t) len);
```
bshifter: This looks good given the current design
Since the schema is the single source of truth and builders are created directly from it (with the caller iterating over schema fields), the builder type should always match type->id().

That said, this relies on a fairly strong invariant. The static_casts here would lead to undefined behavior if this function were ever reused in a different context where that guarantee doesn’t hold.

It might be worth either:
- documenting this assumption explicitly, or
- adding a debug assertion (assert(builder->type()->id() == type->id())) to make the dependency clear and catch violations early.

Just calling it out to make future refactors safer.

mitzkia:
Thanks for the Light updates. I can approve that part.

bshifter (dismissed):
Nice job! I've enjoyed reviewing this. Approving.

modules/arrow-flight/arrow-flight-worker.cpp
alltilla: Nice catch!
I added a comment for now. I don't want to risk adding asserts to hotpaths.

alltilla:
@bshifter I will drop the CI image commit, which will make CI red under this PR, but eventually green on main, when the images get rebuilt. After that if you feel like this PR is OK, kindly approve and merge it. Thanks!

bshifter (dismissed):

modules/arrow-flight/arrow-flight-worker.cpp
```diff
@@ -196,6 +204,69 @@ DestinationWorker::append_value(arrow::ArrayBuilder *builder, const std::shared_
...
+              return LTR_NOT_CONNECTED;
+            }
+        }
+    }
+
+  if (this->get_batch_size() == 1)
+    {
+      GString *path_buf;
+      ScratchBuffersMarker path_marker;
+      const gchar *path = this->format_path(msg, &path_buf, &path_marker);
```
bazsi: do we need to pass the marker around? I think it is enough to protect this via the local mark in this function.

alltilla: You are right. I managed to remove scratch buffers from the worker code, with a `GString *buffer` in the worker.

Parts of the code was squashed to the `format_path()` commit, the rest came as a new commit.

bshifter (approved):

PR #1076, opened 2026-05-19
plugin: keep modules resident during discovery scan

alltilla:
`plugin_discover_loadable_modules()` and `plugin_list_modules()` dlopen every `.so` in the module dir to read its `module_info`, then `dlclose` it. That's safe for syslog-ng's own registration model because no module init function runs during the scan, but it's not safe for C++ static initializers in third-party libraries.

Protobuf's generated `.pb.cc` files contain static initializers that call `InternalAddGeneratedFile()` with a raw pointer into the calling `.so`'s `.rodata`. The `DescriptorPool` is a process-wide singleton in `libprotobuf` and stores that pointer without copying. If the module gets `dlclose`'d and its `rodata` unmaps, the pool's entry dangles. The next module that registers its descriptors walks the pool's tree to find the insert position, the comparator `memcmp`s an existing entry's filename, and crashes on the unmapped page.

This was not a problem for a while, because we moved all the protobuf descriptors into the `libgrpc-protos.so` file, which is not a plugin, so it does not get unloaded during plugin disco.

I am planning to add Arrow Flight, which brings its own descriptors, and the plugin links to them. when the plugin gets unloaded, the descriptors get unloaded too, and the next plugin that registers its own descriptors crashes when it tries to compare against the dangling pointer.

The production load path in `plugin_load_module()` has called `g_module_make_resident()` since the
plugin framework was added in 2010, because plugins register function pointers and struct pointers into long-lived registries, unloading the `.so` would dangle those too. We can apply the same treatment to the discovery paths: every
successfully-opened module stays mapped for the rest of the process's lifetime. `dlopen` on the same module later just bumps the refcount, no static initializers re-run.

Assisted-by: Claude:claude-opus-4-7[1m]

### DIFF
--- a/lib/plugin.c
+++ b/lib/plugin.c
@@ -587,7 +587,10 @@ plugin_discover_loadable_modules(PluginContext *context)
                 plugin_extract_candidate_plugins(context, module_name, module_info);
               g_free(module_name);
               if (mod)
-                g_module_close(mod);
+                {
+                  g_module_make_resident(mod);
+                  g_module_close(mod);
+                }
               else
                 mod = NULL;
             }
@@ -719,7 +722,10 @@ plugin_list_modules(FILE *out, gboolean verbose)
                 }
               g_free(module_name);
               if (mod)
-                g_module_close(mod);
+                {
+                  g_module_make_resident(mod);
+                  g_module_close(mod);
+                }
             }
         }
       g_dir_close(dir);


bazsi:
This was the reason I did the discovery in a separate process originally but that was reverted for reasons beyond my recollection. This will significantly increase the memory use, but not more than my monolithic builds, which is what our production docker images already use, so :+1:

bazsi (approved):

alltilla:
```
log {
  destination { stdout(); };
};
```

58.7969M -> 77.2148M

+20M memory usage, which is not so good on a mostly idle, mostly empty config instance, but negligible for a busy instance.
I am fully open to implement a discovery in a separate process again if we deem this too bad.

PR #1078, opened 2026-05-19
Logscheduler batching improvements

bazsi:
This improved parallelize performance by ~9.3% on c5.2xlarge

This PR significantly reduces window-full scenarios, thereby reducing the times we need to suspend and wakeup the source. This in turn reduces the pressure on the main thread's ivykis loop and related machinery (iv_event_post).

### DIFF
--- a/lib/logscheduler.c
+++ b/lib/logscheduler.c
@@ -359,6 +359,10 @@ _queue_thread(LogScheduler *self, LogSchedulerThreadState *thread_state, LogMess
   log_msg_unref(msg);
 
   stats_counter_inc(self->partitions[partition_index].metrics.assigned_events_total);
+  if (self->options->batch_size && thread_state->partitions[partition_index].len >= self->options->batch_size)
+    {
+      _move_batch_to_partition(self, thread_state, partition_index);
+    }
 }
 
 
--- a/lib/logscheduler.c
+++ b/lib/logscheduler.c
@@ -623,7 +623,7 @@ void
 log_scheduler_options_defaults(LogSchedulerOptions *options)
 {
   options->num_partitions = -1;
-  options->batch_size = 0;
+  options->batch_size = -1;
   options->partition_key = NULL;
   options->log_fetch_limit = 1000;
 }
@@ -644,6 +644,8 @@ log_scheduler_options_init(LogSchedulerOptions *options, GlobalConfig *cfg)
                   evt_tag_int("used_workers", LOGSCHEDULER_MAX_PARTITIONS));
       options->num_partitions = LOGSCHEDULER_MAX_PARTITIONS;
     }
+  if (options->batch_size == -1)
+    options->batch_size = 100;
 
   return TRUE;
 }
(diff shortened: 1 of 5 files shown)


lib/timeutils/misc.c
```diff
@@ -66,39 +66,33 @@ check_nanosleep(void)
 }
 
 void
-timespec_add_msec(struct timespec *ts, glong msec)
+timespec_add_nsec(struct timespec *ts, glong nsec)
```
MrAnno: `timespec_add_usec()` used `gint64` for its parameter.  `nsec` is even more precise.

On 32-bit archs, this would limit us to ~2 second. This affects msec and usec too due to the refactor.

Can we use gint64 everywhere?

lib/logscheduler.c
```diff
@@ -642,6 +642,8 @@ log_scheduler_options_init(LogSchedulerOptions *options, GlobalConfig *cfg)
                   evt_tag_int("used_workers", LOGSCHEDULER_MAX_PARTITIONS));
       options->num_partitions = LOGSCHEDULER_MAX_PARTITIONS;
     }
+  if (options->partition_key && options->batch_size == 0)
+    options->batch_size = 100;
```
MrAnno, line 648: This also means the user can't disable batching when partition-key is used.
They can set it to 1, though.

lib/logsource.c
```diff
@@ -121,29 +121,45 @@ _flow_control_rate_adjust(LogSource *self)
...
                 }
               else
                 {
-                  /* ok, we seem to have a close enough measurement, this means
-                   * we do have a high rate.  Calculate how much we should sleep
-                   * in case the window gets full */
+
+                  /* ok, we seem to be running at more than 2340 eps (e.g.
+                   * 0.42msec/message), this means we do have a high rate.
+                   * Since going back to the main loop is expensive, let's
```
MrAnno, line 137: The nanosleep was originally here to overcome the cost of returning to the main loop and back  (the cost of the ivykis machinery).

This cost itself hasn't changed, but we started using parallelize() and huge amounts of FilterX processing on the pipeline, which triggered this msec change.

So this should be about something like this: the cost of processing a message on the source thread increased by an order of magnitude, so we increased both the max allowed time to sleep and the amount of messages we want to wait for. Is this true?
We could somehow adjust this comment to contain this cost, not just the main loop cost.

MrAnno:
I measured the perf of this PR  together with the atomic refcount elimination, it's huge :)

lib/timeutils/misc.c
bazsi: good catch. I was just focusing on 1) the refactor, 2) that a long should be enough to hold a single second, but it is true that sometimes call-sites may want to add more to the timespec value than a single second.

fixing this.

lib/logscheduler.c
bazsi: fixed this up. I've now set batch_size to -1 initially and only override the value to a 100  if it is still -1 at the time of init.

this way batching can be disabled by batch-size(0) but we will default to 100.

lib/logsource.c
bazsi: added another round or refinements to the comments.

MrAnno (approved):

bazsi:
\o/

PR #1080, opened 2026-05-21
FilterX JIT compiler - Inlining and optimizing parts of libfilterx

MrAnno:


fernya (changes requested):
The core of this PR is nice and clean (generation of the bitcode).
I flag it because it contains some manually tuning of the LLVM code generator which are proven to be bad for actual performance. If you drop their hardwired nature then we are ready to ship this. 💪 

cb95eefa632caebdd3dfc250ba5583310308e157 - inlining decision is best left for LLVM to make
d047a167672eeb2f865cf4b82417496c23bc1c0a - custom codegen attributes were tried and most of the time they make performance worse
e76b81fbe827eecc73e6015d45051b2ee8ebd560 - empty commit?

MrAnno:
3,0,1 was our best codeopt setting, so I kept that commit, but I removed everything else.

fernya (approved):
The PR fixed the observations made earlier. I use it as base for my work for the last weeks and consider it as stable. This cleaned-up state is ripe enough to get merged!

alltilla (approved):
Same as @fernya

PR #1083, opened 2026-05-22
Reduce GSockAddr->refcnt related cache ping pong

bazsi:
This PR attempts to reduce the cache contention on GSockAddr instances in case we share those across parallelize workers.

In those cases each LogMessage will point to the same pair of saddr/daddr instances and we will be banging their refcnt from all of the threads in parallel.

This PR was originally combined with a refactor, which I have now split out to #1086 I am still addressing the review comments here, but in general the two can be reviewed/merged separately.

### DIFF
--- a/lib/gsockaddr.c
+++ b/lib/gsockaddr.c
@@ -159,6 +159,20 @@ g_sockaddr_get_address(GSockAddr *self, guint8 *buffer, gsize buffer_size, gsize
     }
 }
 
+void
+g_sockaddr_unshare(GSockAddr **pa)
+{
+  GSockAddr *a = *pa;
+
+  if (g_atomic_counter_get(&a->super.refcnt) == 1)
+    return;
+  gsize len = g_sockaddr_len(a);
+  GSockAddr *new = g_malloc(len);
+  memcpy(new, a, len);
+  g_atomic_counter_set(&new->super.refcnt, 1);
+  *pa = new;
+}
+
 /* AF_INET socket address */
 /*+
 
--- a/lib/gsockaddr.h
+++ b/lib/gsockaddr.h
@@ -89,6 +89,7 @@ guint8 *g_sockaddr_get_address(GSockAddr *self, guint8 *buffer, gsize buffer_siz
 gsize g_sockaddr_len(GSockAddr *a);
 
 GSockAddr *g_sockaddr_new(struct sockaddr *sa, int salen);
+void g_sockaddr_unshare(GSockAddr **pa);
 
 static inline struct sockaddr *
 g_sockaddr_get_sa(GSockAddr *self)
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -471,6 +471,11 @@ _set_addresses(LogReader *self, LogMessage *msg, LogTransportAuxData *aux)
 set:
   log_msg_set_saddr(msg, source_addr);
   log_msg_set_daddr(msg, aux->local_addr ? : self->local_addr);
+
+  g_sockaddr_unref(aux->peer_addr);
+  aux->peer_addr = NULL;
+  g_sockaddr_unref(aux->local_addr);
+  aux->local_addr = NULL;
 }
 
 static gboolean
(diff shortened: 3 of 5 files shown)


lib/gsockaddr.h
```diff
@@ -40,10 +40,14 @@
 #define GSA_ADDRESS_ONLY  1
 #define GSA_ADDRESS_PORT  2
 
+#define G_SOCKADDR_HEADER \
+  GAtomicCounter refcnt; \
+  guint32 salen
```
MrAnno: I think this could be a real header struct without causing anything bad around alignments.

lib/gsockaddr.c
```diff
@@ -558,7 +558,7 @@ g_sockaddr_unix_new(const gchar *name)
 GSockAddr *
 g_sockaddr_unix_new2(struct sockaddr_un *saun, int sunlen)
 {
-  GSockAddrUnix *addr = g_slice_new0(GSockAddrUnix);
+  GSockAddrUnix *addr = g_new0(GSockAddrUnix, 1);
```
MrAnno, line 578: g_slice_free1 is not replaced here.

lib/logreader.c
```diff
@@ -471,6 +471,11 @@ _set_addresses(LogReader *self, LogMessage *msg, LogTransportAuxData *aux)
 set:
   log_msg_set_saddr(msg, source_addr);
   log_msg_set_daddr(msg, aux->local_addr ? : self->local_addr);
+
+  g_sockaddr_unref(aux->peer_addr);
+  aux->peer_addr = NULL;
+  g_sockaddr_unref(aux->local_addr);
+  aux->local_addr = NULL;
```
MrAnno, line 478: I think this is the responsibility of the `LogTransportAuxData` class, we could add a method that reinits or clears the addresses.

lib/gsockaddr.c
```diff
@@ -152,6 +152,20 @@ g_sockaddr_get_address(GSockAddr *self, guint8 *buffer, gsize buffer_size, gsize
     }
 }
 
+void
+g_sockaddr_unshare(GSockAddr **pa)
+{
+  GSockAddr *a = *pa;
+
+  if (g_atomic_counter_get(&a->refcnt) == 1)
```
MrAnno: This is racy if we have parallel increments/decrements while this function runs.

Since GSockAddr is shared across different messages from the same source, I think this can be a real race as older messages are freed on the destination side.

lib/gsockaddr.c
MrAnno: I found it, it was in the previous commit.

lib/gsockaddr.c
MrAnno: As an alternative, can't we call an unconditional copy for each message? We could even embed the addresses into the message entirely.

Another question that came into my mind: do we want to make this dependent of `parallelize()`? (would it be bad if we did this regardless of parallelize?)

lib/logreader.c
bazsi: There's one function there, but that "destroys" the aux data.  Right now that one is called in logreader once the message is posted. The problem with that is that multiple refs are being held, preventing g_sockaddr_unshare() to work.

maybe I'll split this branch into the refactor, which is mostly trivial stuff and then discuss how we could unshare the sockaddress instances.

lib/gsockaddr.c
bazsi: This is not racy, the caller has to have a ref to call this function, and since the ref count is 1, that means that the only reference is held by the caller. And the caller can't increment the reference as we are executing the current function.

A different thread simply can't own a reference at this point.

I was thinking about using embedding the GSockAddr in the message, unfortunately we can't be certain about the allocation size thus we would have to take the max at that point. Even with these patches that would be 120 bytes per sockaddr, stored two times. The LogMessage struct is 160 bytes at the moment, increasing the size to 300.

lib/gsockaddr.h
bazsi: I did this in my second iteration.

bazsi: moved this refactor to #1086

MrAnno (approved):

PR #1084, opened 2026-05-22
Retire logmsg refcache

bazsi:
This PR gets rid of the LogMessage refcache. 

The refcount for LogMessage instances are mostly changed by the same CPU with a single handover from the source to the destination worker. With that said, in most cases this is  going to be exclusive in the cache. 

It is complex, does not help performance much if at all. This PR improves the CPU cycle count, although only a small improvement. 

Let's get rid of refcache, prime example of premature optimization on my part.

RIP.

### DIFF
--- a/lib/logmsg/logmsg.c
+++ b/lib/logmsg/logmsg.c
@@ -51,42 +51,6 @@
 #include <ctype.h>
 #include <syslog.h>
 
-/*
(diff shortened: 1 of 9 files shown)


MrAnno (approved):

MrAnno:
RIP.

PR #1087, opened 2026-05-23
light: stabilize transport error test across SSL versions

MrAnno:
```
E         - SSL routines::wrong version number
E         + SSL routines::packet length too long
```

Spaces at the beginning of the TCP stream make OpenSSL parse a smaller packet size.

The OpenSSL version where this caused a problem:
OpenSSL 3.5.5 27 Jan 2026 (Library: OpenSSL 3.5.5 27 Jan 2026)


bazsi:
wow, great catch, thank you.

bazsi (approved):

PR #1091, opened 2026-05-25
FilterX: O(1) variable lookup without memmove/realloc

MrAnno:
I need this for the JIT compiler so that I can cheaply map a real stack-allocated area to the FilterX variable storage.

In general (even without the JIT feature), this PR improves variable lookup.

----

Dictionary and variable lookups have to be as cheap as possible.

For example, if we have plans to optimize "common subexpression" lookups in dicts, it means we are putting the subexpressions into variables or something similar, hence the need for trivial variable lookups (or this can be done inside LLVM IR, it's undecided yet).

This PR makes variable lookup cheaper:
- each FilterX block collects its variables during init/compile time
- variables are densely packed within each FilterXScope, similarly to how local variables would look like on a stack frame:
  - the lookup is predictable and cheap (O(1)), the operation itself is just an indexing into an inlined array
  - the storage never grows or shrinks, no memmove or realloc
  - only the first lookup falls back to being expensive as we need to steal variables from the previous block (binary-search)

TLDR: This PR gets rid of a binary-search on each variable access, and some `memmove` and `realloc` operations.
The new cost is a `memset` when entering a given block, which should be worth it in the case of many lookups.

The overall perf improvement may not be huge on this branch, but combining this work with JIT gives us a 10-15% perf improvement (by giving variable visibility to LLVM in a cheap way).

----

Notes/question:

- `loads_vars()` loads variables dynamically, which can't be computed at compile time, so it should either have a slow-path dynamic area or we should consider removing it (I think it's unused, it was created only for symmetry reasons). Let me know what you think.

- The last 2 patches get rid of the scope reuse logic and the write-protection-based pulling (it was 3% in some scenarios); a predictable O(1) lookup seems a bigger win in general, but I can drop these patches, if needed.


alltilla:
Very nice!

My only fear is not fully removing/deprecating `load_vars()`. Maybe we should have at least a code for it that fails config parse/init time, that it got removed.

I don't think we use it anywhere, so I am open to remove its functionality, but at least make it emit an easier-to-debug error for the unlikely users of it.

MrAnno:
I had plans to either implement a dynamic storage for it or remove it normally, depending on your answers.
This is just a WIP state, I won't leave it like that.

MrAnno:
This PR should be a big improvement on blocks with many lookups. The code should be optimized further to only fill the variable layout for a block, when lookups (and writes) worth the memcpy cost (blocks returning early, blocks writing nothing and reading only a few variables).

bazsi:
It now shows about 5% degradation in the callgrind tests at the moment.

MrAnno:
@bazsi 

The callgrind config we use is not really representative in this specific case, but I will work on making that cost disappear soon and also improve "global variable" access.

For now, this is what we have:

**+15% perf improvement with a config that accesses 30 variables ~300 times.**

----

### PR (4.25.0.271.g26673c2.dirty+4.25.0_git20260528):
```
==3460142== I   refs:      15,178,043
==3460142== I1  misses:        53,745
==3460142== LLi misses:        52,909
==3460142== I1  miss rate:       0.35%
==3460142== LLi miss rate:       0.35%
==3460142==
==3460142== D   refs:       6,376,556  (4,158,419 rd + 2,218,137 wr)
==3460142== D1  misses:       285,044  (  253,657 rd +    31,387 wr)
==3460142== LLd misses:       264,165  (  236,892 rd +    27,273 wr)
==3460142== D1  miss rate:        4.5% (      6.1%   +       1.4%  )
==3460142== LLd miss rate:        4.1% (      5.7%   +       1.2%  )
==3460142==
==3460142== LL refs:          338,789  (  307,402 rd +    31,387 wr)
==3460142== LL misses:        317,074  (  289,801 rd +    27,273 wr)
==3460142== LL miss rate:         1.5% (      1.5%   +       1.2%  )
```

avg_rate=208644.77 msg/s, stdev_rate=1.86%, count=6263900, time=30.0218, avg_msg_size=256, bandwidth=52161.19 KiB/s
avg_rate=203388.08 msg/s, stdev_rate=2.29%, count=6104200, time=30.0126, avg_msg_size=256, bandwidth=50847.02 KiB/s


### Common parent `main` (4.25.0.261.g0bd1e4b+4.25.0_git20260528):
```
==3566844==
==3566844== Events    : Ir Dr Dw I1mr D1mr D1mw ILmr DLmr DLmw
==3566844== Collected : 17605662 4360543 2135332 57654 252788 33507 56608 235646 28980
==3566844==
==3566844== I   refs:      17,605,662
==3566844== I1  misses:        57,654
==3566844== LLi misses:        56,608
==3566844== I1  miss rate:       0.33%
==3566844== LLi miss rate:       0.32%
==3566844==
==3566844== D   refs:       6,495,875  (4,360,543 rd + 2,135,332 wr)
==3566844== D1  misses:       286,295  (  252,788 rd +    33,507 wr)
==3566844== LLd misses:       264,626  (  235,646 rd +    28,980 wr)
==3566844== D1  miss rate:        4.4% (      5.8%   +       1.6%  )
==3566844== LLd miss rate:        4.1% (      5.4%   +       1.4%  )
==3566844==
==3566844== LL refs:          343,949  (  310,442 rd +    33,507 wr)
==3566844== LL misses:        321,234  (  292,254 rd +    28,980 wr)
==3566844== LL miss rate:         1.3% (      1.3%   +       1.4%  )
```

avg_rate=181062.68 msg/s, stdev_rate=2.71%, count=5436500, time=30.0255, avg_msg_size=256, bandwidth=45265.67 KiB/s
avg_rate=180563.59 msg/s, stdev_rate=2.05%, count=5422700, time=30.0321, avg_msg_size=256, bandwidth=45140.90 KiB/s

bazsi:
I think it should even improve the callgrind config file case, seems that for some reason we allocate a lot more FilterXRef and dict instances than without the patch:

Patched:
```
 82,729 ( 4.04%) 21,935 ( 4.10%) 11,342 ( 3.53%) 268 ( 1.31%)   108 ( 0.39%) 1,227 (13.10%) 224 ( 1.88%)  96 ( 0.62%) 1,225 (14.19%)  < /source/lib/filterx/filterx-eval.h:_filterx_ref_new (1,380x)
 10,878 ( 0.53%)  2,886 ( 0.54%)  1,776 ( 0.55%)  12 ( 0.06%)     0            444 ( 4.74%)   5 ( 0.04%)   0            444 ( 5.14%)  < /source/lib/filterx/filterx-eval.h:filterx_dict_new_with_table (222x)
```

Unpatched:
```
 52,741 ( 2.76%) 13,367 ( 2.75%)  7,058 ( 2.35%) 274 ( 1.35%)   115 ( 0.43%)   570 ( 6.84%) 230 ( 2.02%)  96 ( 0.63%)   563 ( 7.33%)  < /source/lib/filterx/filterx-eval.h:_filterx_ref_new (666x)
  7,938 ( 0.42%)  2,106 ( 0.43%)  1,296 ( 0.43%)  17 ( 0.08%)     0            329 ( 3.95%)   5 ( 0.04%)   0            329 ( 4.28%)  < /source/lib/filterx/filterx-eval.h:filterx_dict_new_with_table (162x)

```

All the other allocations line up perfectly. Maybe the order of a new FilterXRef being allocated vs an old one destroyed is different which triggers a copy-on-write?

The lookup_variable call is completely eliminated. Nice!

MrAnno:
Yes, this is the last note in the PR description:

> The last 2 patches get rid of the scope reuse logic and the write-protection-based pulling (it was 3% in some scenarios); a predictable O(1) lookup seems a bigger win in general, but I can drop these patches, if needed.

I'm working on optimizing this PR further.

bazsi:
Maybe the ref move was broken, here's a breakdown of callers to filterx_ref_clone():

PR:
```
 74,778 ( 3.65%) 20,736 ( 3.87%) 14,034 ( 4.36%)   2 ( 0.01%)   709 ( 2.57%)   591 ( 6.31%)   0          276 ( 1.78%)   591 ( 6.85%)  < /source/lib/filterx/filterx-object.h:_filterx_dict_clone_container (768x)
 28,225 ( 1.38%)  6,347 ( 1.19%)  3,710 ( 1.15%) 296 ( 1.44%)   172 ( 0.62%)    57 ( 0.61%) 254 ( 2.13%) 168 ( 1.08%)    57 ( 0.66%)  < /source/lib/filterx/filterx-object.h:_do_assign (36x)
 21,090 ( 1.03%)  5,670 ( 1.06%)  3,966 ( 1.23%)  10 ( 0.05%)   133 ( 0.48%)   398 ( 4.25%)   0            0            397 ( 4.60%)  < /source/lib/filterx/filterx-object.h:filterx_scope_lookup_variable (210x)
  3,456 ( 0.17%)    972 ( 0.18%)    648 ( 0.20%)   0             72 ( 0.26%)    60 ( 0.64%)   0           72 ( 0.46%)    60 ( 0.70%)  < /source/lib/filterx/filterx-object.h:_setattr_eval (36x)
  2,304 ( 0.11%)    648 ( 0.12%)    432 ( 0.13%)   0             48 ( 0.17%)    15 ( 0.16%)   0           48 ( 0.31%)    15 ( 0.17%)  < /source/lib/filterx/filterx-object.h:_set_with_fallbacks (24x)
  1,728 ( 0.08%)    486 ( 0.09%)    324 ( 0.10%)   1 ( 0.00%)    36 ( 0.13%)    32 ( 0.34%)   0           36 ( 0.23%)    32 ( 0.37%)  < /source/lib/filterx/filterx-object.h:_literal_dict_eval (18x)
  1,212 ( 0.06%)    324 ( 0.06%)    228 ( 0.07%)   0              6 ( 0.02%)     9 ( 0.10%)   0            0              9 ( 0.10%)  < /source/lib/filterx/filterx-object.h:_literal_dict_store_elem (12x)
  7,728 ( 0.38%)  3,312 ( 0.62%)  1,104 ( 0.34%)  14 ( 0.07%)     0              0           12 ( 0.10%)   .              .           *  /source/lib/filterx/filterx-ref.c:_filterx_ref_clone [/install/sbin/syslog-ng]
116,401 ( 5.68%) 30,767 ( 5.75%) 21,830 ( 6.79%) 293 ( 1.43%)   642 ( 2.32%) 1,162 (12.41%) 242 ( 2.03%) 348 ( 2.24%) 1,161 (13.45%)  >   /source/lib/filterx/filterx-ref.c:_filterx_ref_new (1,104x) [/install/sbin/syslog-ng]

```

Base:
```
 29,437 ( 1.54%)  6,671 ( 1.37%)  3,938 ( 1.31%) 295 ( 1.45%)   171 ( 0.65%)    69 ( 0.83%) 260 ( 2.28%) 168 ( 1.11%)    69 ( 0.90%)  < /source/lib/filterx/filterx-object.h:_do_assign (48x)
 27,828 ( 1.46%)  7,776 ( 1.60%)  5,220 ( 1.74%)  16 ( 0.08%)   511 ( 1.93%)   231 ( 2.77%)   0          276 ( 1.82%)   225 ( 2.93%)  < /source/lib/filterx/filterx-object.h:_filterx_dict_clone_container (288x)
  7,758 ( 0.41%)  2,106 ( 0.43%)  1,458 ( 0.49%)  28 ( 0.14%)   118 ( 0.45%)    60 ( 0.72%)   0            0             60 ( 0.78%)  < /source/lib/filterx/filterx-object.h:filterx_scope_lookup_variable (78x)
  3,456 ( 0.18%)    972 ( 0.20%)    648 ( 0.22%)   7 ( 0.03%)    72 ( 0.27%)    53 ( 0.64%)   0           72 ( 0.48%)    53 ( 0.69%)  < /source/lib/filterx/filterx-object.h:_setattr_eval (36x)
  2,304 ( 0.12%)    648 ( 0.13%)    432 ( 0.14%)   0             48 ( 0.18%)    27 ( 0.32%)   0           48 ( 0.32%)    27 ( 0.35%)  < /source/lib/filterx/filterx-object.h:_set_with_fallbacks (24x)
  1,728 ( 0.09%)    486 ( 0.10%)    324 ( 0.11%)   3 ( 0.01%)    39 ( 0.15%)    27 ( 0.32%)   0           36 ( 0.24%)    27 ( 0.35%)  < /source/lib/filterx/filterx-object.h:_literal_dict_eval (18x)
  1,212 ( 0.06%)    324 ( 0.07%)    228 ( 0.08%)   0             15 ( 0.06%)     9 ( 0.11%)   0            0              9 ( 0.12%)  < /source/lib/filterx/filterx-object.h:_literal_dict_store_elem (12x)
  3,528 ( 0.18%)  1,512 ( 0.31%)    504 ( 0.17%)  51 ( 0.25%)     0              0           12 ( 0.11%)   .              .           *  /source/lib/filterx/filterx-ref.c:_filterx_ref_clone [/install/sbin/syslog-ng]
 66,601 ( 3.49%) 16,967 ( 3.49%) 11,630 ( 3.87%) 298 ( 1.47%)   547 ( 2.07%)   476 ( 5.71%) 248 ( 2.18%) 348 ( 2.30%)   470 ( 6.12%)  >   /source/lib/filterx/filterx-ref.c:_filterx_ref_new (504x) [/install/sbin/syslog-ng]

```

bazsi:
This might be the culprit: https://github.com/axoflow/axosyslog/pull/1091/changes#diff-2c9bea8b99440eb4f9b2a74597d50f89da3c748e4c1750fd966704ada0a0f495L197-L204

Previously, we could borrow variables from the previous scope if the current scope is still read-only. In those cases we
can get away with not copying values, while in this branch we don't do that anymore.

MrAnno:
Yes, that's exactly what I was referring to in the PR description.

lib/filterx/filterx-scope.h
```diff
@@ -148,18 +140,7 @@ filterx_scope_is_write_protected(FilterXScope *self)
 static inline void
 filterx_scope_make_writable(FilterXScope *self)
 {
-  g_assert(self->variables.len == 0);
```
MrAnno, line 151: !

lib/filterx/filterx-scope.c
```diff
@@ -382,70 +309,51 @@ filterx_scope_sync(FilterXScope *self, LogMessage **pmsg, const LogPathOptions *
...
-  self->variables.size = coupled_variables_size;
-  self->coupled_variables = TRUE;
+  self->num_variables = layout ? layout->num_variables : 0;
+  filterx_scope_variable_layout_fill(layout, self->variables);
+
   if (parent_scope)
     {
       self->parent_scope = parent_scope;
       self->generation = parent_scope->generation + 1;
-      if (parent_scope->variables.len > 0)
```
MrAnno, line 405: !

MrAnno:
Now it's okay with `callgrind-syslog-ng.conf` too:

```
PR: I   refs:      1,860,809
Base: I   refs:      1,904,141
```

(The callgrind job doesn't detect crashes, so I asked @mitzkia to pull this config into Light and run a smoke test on it)

Update:
```
I   refs:      1,856,165
```

lib/filterx/filterx-eval.c
```diff
@@ -422,11 +422,12 @@ filterx_eval_begin_context(FilterXEvalContext *context,
...
-      if (previous_context->scope != scope)
+      if (filterx_scope_is_fork_point(previous_context->scope))
         {
-          /* Unless we have our own scope object, we need to reuse the
-           * allocator as well. This means that in that case we won't need
-           * to remember where we need to go back.
+          /* Unless we have our own scope object without variables stolen from
+           * the previous scope, we need to reuse the allocator as well.
+           * This means that in that case we won't need to remember where we
+           * need to go back.
```
MrAnno: This comment became confusing, I need to rephrase it.

lib/filterx/filterx-scope-var-layout.h
```diff
@@ -0,0 +1,48 @@
...
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#ifndef FILTERX_SCOPE_VAR_LAYOUT_H
+#define FILTERX_SCOPE_VAR_LAYOUT_H
+
+#include "syslog-ng.h"
```
bazsi: [nit] we don't  include syslog-ng.h, unless this is a root header.

lib/filterx/filterx-scope.h
```diff
@@ -41,40 +42,33 @@
...
-    gint size;
-    union
-    {
-      FilterXVariable *separate;
-      FilterXVariable coupled[0];
-    };
-  } variables;
+
+  guint32 variables_len;
+  guint32 variables_size;
```
bazsi, line 52: I was a bit confused by the names of these two members. I now get that `variables_size` is the length of the tailing array, but `variables_len` translates sounds confusingly similar to me.

If I get correctly, `variables_len` just counts the number of slots we would actually use, it is increased at variable registration and when we pull a value from the parent.

maybe we should rename this to `variables_used`. Also we only seem to compare it against zero, so maybe a boolean would suffice?

lib/filterx/filterx-scope.c
```diff
@@ -80,14 +80,23 @@ _validate_variable(FilterXScope *self, FilterXVariable *variable)
...
-    output->value = filterx_object_copy(output->value);
+    {
+      if (copy)
+        output->value = filterx_object_copy(output->value);
+      else
+        {
+          /* steal variable without creating refs/copies from a scope that won't be ever used */
+          parent_variable->value = NULL;
+        }
+    }
```
bazsi, line 102: Instead of "steal" as a verb, I would try to use "move" somehow.

we either copy or move the reference from the parent scope, depending on whether it was 
a fork point. Maybe encode this term into an enum which we can pass instead of this boolean.

```
enum {
  FX_PULL_BY_COPYING,
  FX_PULL_BY_MOVING,
};
```

lib/filterx/filterx-scope.c
```diff
@@ -97,21 +106,24 @@ _search_and_pull_from_parent_scopes(FilterXScope *self, FilterXVariableHandle ha
...
         {
           /* NOTE: we validate against @self */
           if (_validate_variable(self, v))
             {
-              if (self->write_protected)
-                return v;
-              else
-                return _pull_variable_from_parent_scope(self, v, output);
+              /* we must do a real copy when crossing fork points */
+              gboolean copy = crossed_fork_point;
```
bazsi: an enum would make this nicer too.

lib/filterx/filterx-scope.c
```diff
@@ -97,21 +106,24 @@ _search_and_pull_from_parent_scopes(FilterXScope *self, FilterXVariableHandle ha
...
     {
+      if (filterx_scope_is_fork_point(scope))
+        crossed_fork_point = TRUE;
+
       if (_search_variable(scope, handle, &v, &v_index))
         {
           /* NOTE: we validate against @self */
           if (_validate_variable(self, v))
             {
-              if (self->write_protected)
```
bazsi, line 199: While I would really like to get rid off  write protection, I don't think the new code is equivalent here. The fork point based optimization helps with cases where we have subsequent filterx blocks following each other, so one can end a filterx block and start a new one, without routing level elements in between them.

The write protection actually optimizes cases where parent _is_ a fork point and the current scope has a conditional around a value propagated from a parent. Something like this:

```
          +--- branch1 (meta.product == "foo")
         /
fork----+
         \
          +--- branch2 (meta.product == "bar")
```

In this case both branch1 and branch2 would copy `meta` just to check the the product field, and if the condition fails, it will throw away the copy. A pattern like this is actually generating N FilterXRef instances:
1) the one pointing to the variable `meta`
2) floating xrefs for any subsequent getattrs on `meta`

With a lot of failing branches this overhead can add up, especially if we have multiple similar conditions at the top of the branch.

Now,  I welcome the removal of complexity, I just wanted to point out that the new code does not implement the same optimization. Also NOTE: the existing solution only works as long as we don't actually store the borrowed value into the new scope, only use it directly, this means that if the initial conditions pass in the branch and the scope eventually becomes writable we will do a 2nd pull, which will then correctly copy the value.

The current performance improvement offsets the loss of the previous optimization, but eventually we will have to optimize the many-falling-routing-branches-before-finding-the-right-one as well. For me a non-linear solution would be preferred over the current iterative approach.

MrAnno: I understand.

One thing to add is that write-protection conflicted with the O(1) lookup, repeating binary search at each access until the first write, so I think this version should pay off after a certain number of lookups.

I think we can reimplement write-protection without losing O(1) lookups with the help of new flags, but I would do it in a separate PR so that I can spend some time concentrating on the JIT side of this PR.

lib/filterx/filterx-scope.c
MrAnno: move is an existing functionality which is the same idea, but a little different, so I tried to avoid the confusion.

I can try to come up with something else.

lib/filterx/filterx-scope-var-layout.h
MrAnno: I don't remember why I needed it, but I remember I added it (not LLM), probably for glib types.

I like to list all direct dependencies of a given unit, but unfortunately I can't do it, because glib shouldn't be included because of the compat layer, but I can't include the compat layer directly either, so the closest I am allowed to do is syslog-ng.h.

lib/filterx/filterx-scope.c
bazsi: I gave this some more thought, let's get rid off this write_protected stuff as you suggest, we gained a lot more than we lose with this and this 3-4% that this gives us at ~50 branches is not really solving anything.

I originally also envisioned  that understanding when we _change_ a variable will use useful for CSE, but that is probably a completely different implementation anyway. So just let's merge this as is.

lib/filterx/filterx-scope-var-layout.h
bazsi: the conventions I tried to maintain were:

1) syslog-ng.h is only included if nothing else is included. This takes care of glib and the configure flags
2) we include all dependencies as you suggested (but not syslog-ng.h)
3) in implementation files we start with their own header
4) include ordering from more specific to more generic (e.g. stdio.h would come last, while we start with syslog-ng specific headers), try to keep headers at roughly the same abstraction level in the same block

This ensures that all headers indeed include their dependencies, as a reverse ordering would allow a header to borrow includes from its users.

But of course this is not validated, so not perfect in the codebase, you will probably find violations.

lib/filterx/filterx-scope.c
bazsi: I think it is the same idea.  filterx_object_move_key() will unset a value and return it in a movable form (e.g. floating if it is a ref, or merely a ref otherwise). but I don't have a strong opinion here.

bazsi (dismissed):
None of my comments are critical, so I am approving this. Great work. If you decide to change my nits, I am happy  to reapprove.

MrAnno:
I will fix them soon, will be home tomorrow.

alltilla (approved):

PR #1092, opened 2026-05-26
Add large timeout for syslog-ng spawn

fernya:
Increase startup timeout during light tests. It's needed due to some large configs, as initial JIT compilation takes time on startup.

```
# JIT enabled
Tests passed in 120.49s (0:02:00)

# JIT disabled
Tests passed in 81.47s (0:01:21)
```

### DIFF
--- a/tests/light/src/axosyslog_light/syslog_ng/syslog_ng.py
+++ b/tests/light/src/axosyslog_light/syslog_ng/syslog_ng.py
@@ -27,7 +27,7 @@ from pathlib import Path
 from subprocess import Popen
 
 from axosyslog_light.common.blocking import wait_until_false
-from axosyslog_light.common.blocking import wait_until_true
+from axosyslog_light.common.blocking import wait_until_true_custom
 from axosyslog_light.syslog_ng.console_log_reader import ConsoleLogReader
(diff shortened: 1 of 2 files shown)


tests/light/src/axosyslog_light/syslog_ng/syslog_ng.py
```diff
@@ -198,7 +198,7 @@ def is_alive(s: SyslogNg) -> bool:
                 self.__validate_crash_returncode(self._process.returncode)
                 self.__error_handling("syslog-ng is not running")
             return s._syslog_ng_ctl.is_control_socket_alive()
-        return wait_until_true(is_alive, self)
+        return wait_until_true_custom(is_alive, (self,), timeout=120)
```
MrAnno, line 201: Why do we need such a big timeout?

syslog-ng should start within a few seconds even when JIT is enabled.

mitzkia: I also disagree with this change. If we keep this change, we can not detect when syslog-ng really starts slower than expected.

fernya:
I close this as we should provide this effect in a cleaner way.

PR #1093, opened 2026-05-26
Filterx switch range case

sodomelle:
Example usage:
```
switch (${values.int}) {
    case 1..4:
        result = "below";
        break;
    case 5:
        result = "exact";
        break;
    default:
        result = "above";
        break;
};
```

### DIFF
--- a/lib/filterx/expr-switch.c
+++ b/lib/filterx/expr-switch.c
@@ -24,6 +24,7 @@
 #include "expr-compound.h"
 #include "expr-comparison.h"
 #include "expr-literal.h"
+#include "expr-function.h"
 #include "object-string.h"
 #include "object-primitive.h"
 #include "filterx-eval.h"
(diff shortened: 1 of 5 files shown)


lib/filterx/expr-switch.c
```diff
@@ -74,7 +102,10 @@ filterx_switch_case_range_new(FilterXExpr *lower, FilterXExpr *upper)
 {
   FilterXSwitchCase *self = g_new0(FilterXSwitchCase, 1);
   filterx_expr_init_instance(&self->super.super, FILTERX_EXPR_TYPE_NAME(switch_case), FXE_READ);
+  self->super.super.free_fn = _switch_case_free;
+  self->super.super.walk_children = _switch_case_walk;
   self->super.operand = lower;
+  self->upper = upper;
```
MrAnno, line 198: This probably belongs to the previous commit.

lib/filterx/expr-switch.c
```diff
@@ -61,11 +62,38 @@ _eval_switch_case(FilterXSwitchCase *self)
...
+    }
+  return TRUE;
+}
+
 FilterXExpr *
 filterx_switch_case_new(FilterXExpr *value)
 {
   FilterXSwitchCase *self = g_new0(FilterXSwitchCase, 1);
   filterx_unary_op_init_instance(&self->super, FILTERX_EXPR_TYPE_NAME(switch_case), FXE_READ, value);
+  self->upper = NULL;
```
MrAnno: We usually don't do this after `g_new0`.

lib/filterx/expr-switch.c
```diff
@@ -225,17 +228,42 @@ _find_matching_case(FilterXSwitch *self, FilterXObject *selector)
     {
       FilterXSwitchCase *switch_case = (FilterXSwitchCase *) g_ptr_array_index(self->cases, i);
 
-      FilterXObject *value = _eval_switch_case(switch_case);
-      if (!value)
-        continue;
+      if (switch_case->upper)
+        {
```
MrAnno: Can we extract this block into an `gboolean _is_switch_range_case_matching()` or anything similar?

(The same can be done with the non-ranged case, but it's small enough to be left inside the else block.)

MrAnno (dismissed):

MrAnno:
We'll need a news entry for this.

bazsi:
$.02: I like this, but depending on the use-case we might want to cache the ranged values as well, especially if those are "small", enumerable ranges (like the ones in the example).

lib/filterx/expr-switch.c
MrAnno: I have been told the PR will be refactored to have an inherited "case" class, which handles ranges more elegantly; that invalidate this note.

mitzkia (dismissed):
Thanks for the Light testcases, I can approve them.

lib/filterx/expr-switch.c
sodomelle: Restructured the commits.

lib/filterx/expr-switch.c
sodomelle: Field doesn't exist in the new structure.

lib/filterx/expr-switch.c
```diff
@@ -55,17 +69,104 @@ filterx_switch_case_get_target(FilterXSwitchCase *self)
...
+      return FALSE;
+    }
+  gboolean result;
+  if (filterx_compare_objects(lower, upper, FCMPX_NUM_BASED | FCMPX_EQ))
+    result = filterx_compare_objects(selector, lower, FCMPX_NUM_BASED | FCMPX_EQ);
+  else if (filterx_compare_objects(lower, upper, FCMPX_NUM_BASED | FCMPX_LT))
+    result = filterx_compare_objects(lower, selector, FCMPX_NUM_BASED | FCMPX_LT | FCMPX_EQ) &&
+             filterx_compare_objects(selector, upper, FCMPX_NUM_BASED | FCMPX_LT);
+  else
+    result = filterx_compare_objects(upper, selector, FCMPX_NUM_BASED | FCMPX_LT | FCMPX_EQ) &&
```
MrAnno: Won't this reverse also the exclusive/inclusive bounds of the range? The real upper bound should be exclusive, or am I wrong?

lib/filterx/expr-switch.c
```diff
@@ -55,17 +69,104 @@ filterx_switch_case_get_target(FilterXSwitchCase *self)
...
+
+static gboolean
+_switch_case_range_match(FilterXSwitchCase *s, FilterXObject *selector)
+{
+  FilterXSwitchCaseRange *self = (FilterXSwitchCaseRange *) s;
+  FilterXObject *lower = filterx_expr_eval_typed(self->lower);
+  if (!lower)
+    return FALSE;
+  FilterXObject *upper = filterx_expr_eval_typed(self->upper);
+  if (!upper)
```
MrAnno, line 129: Optional/later: don't we want to optimize literal ranges just how we do it with single values?

That way, the _optimize() code could became cleaner by not checking back to the type of the case.

lib/filterx/expr-switch.c
MrAnno: We hard-coded numeric checks here, which means we should do some type checking somewhere, otherwise we will fail silently when someone uses non-numeric selectors or ranges.

sodomelle: `test_switch_range_case_reverse_lower_match` tests this, and in that you can see, it matches the lower boundary.

lib/filterx/expr-switch.c
MrAnno: We discussed this IRL; we only optimize for strings at the moment, so let's not to this.

lib/filterx/expr-switch.c
sodomelle: Added checks.

MrAnno (approved):
Nice, thanks :)

PR #1096, opened 2026-05-28
logsource: delay source wakeup

bazsi:
This is another batching improvement to increase the source-side batches. Instead of waking up the source immediately after a single message becomes available, wait up to 100 messages to be drained from the window and then wakup the source.

This removes the pressure on iv_event_post() in the ivykis mainloop, as well as generating larger batches in LogScheduler, reducing log contention and latency.

### DIFF
--- a/lib/logsource.c
+++ b/lib/logsource.c
@@ -45,6 +45,12 @@
 
 gboolean accurate_nanosleep = FALSE;
 
+/* Batch source wakeups: only re-arm the source once this many slots have
+ * been freed up in the flow-control window since it ran out.  Amortizes
+ * wakeup overhead under high message rates.  Clamped to the source's
+ * initial_window_size at use, so small windows still get woken. */
+#define LOG_SOURCE_WAKEUP_BATCH_SIZE 100
+
 void
 log_source_wakeup(LogSource *self)
 {
@@ -77,6 +83,7 @@ _flow_control_window_size_adjust(LogSource *self, guint32 window_size_increment,
     window_size_increment = log_source_gather_dynamic_window_reclamation(self, window_size_increment);
 
   gsize old_window_size = _window_size_add(self, window_size_increment, &suspended);
+  gsize new_window_size = old_window_size + window_size_increment;
 
   msg_diagnostics("Window size adjustment",
                   evt_tag_int("old_window_size", old_window_size),
@@ -87,7 +94,22 @@ _flow_control_window_size_adjust(LogSource *self, guint32 window_size_increment,
   gboolean need_to_resume_counter = !last_ack_type_is_suspended && suspended;
   if (need_to_resume_counter)
     window_size_counter_resume(&self->window_size);
-  if ((window_size_increment != 0 && old_window_size == 0) || need_to_resume_counter)
+
+  /* Wake the source on one of two transitions:
+   *  - the explicit destination-side suspend bit is being lifted (back-pressure
+   *    released — old_window_size > 0 distinguishes this from the natural drain
+   *    case where _is_suspended() was true only because the counter hit zero);
+   *
+   *  - enough slots have accumulated past the batching threshold since the
+   *    window last ran out.  Threshold is clamped to a quarter of
+   *    initial_window_size so sources with small windows still wake.
+   */
+
+  gboolean threshold_crossed = window_size_increment != 0 &&
+                               old_window_size <= self->wakeup_threshold &&
+                               new_window_size > self->wakeup_threshold;
+  gboolean explicit_suspend_lifted = need_to_resume_counter && old_window_size > 0;
+  if (threshold_crossed || explicit_suspend_lifted)
     log_source_wakeup(self);
 }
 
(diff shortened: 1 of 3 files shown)


lib/logsource.c
```diff
@@ -87,7 +94,25 @@ _flow_control_window_size_adjust(LogSource *self, guint32 window_size_increment,
...
+   *  - the explicit destination-side suspend bit is being lifted (back-pressure
+   *    released — old_window_size > 0 distinguishes this from the natural drain
+   *    case where _is_suspended() was true only because the counter hit zero);
+   *
+   *  - enough slots have accumulated past the batching threshold since the
+   *    window last ran out.  Threshold is clamped to a quarter of
+   *    initial_window_size so sources with small windows still wake.
+   */
+
+  gsize wakeup_threshold = MIN((gsize) LOG_SOURCE_WAKEUP_BATCH_SIZE, self->initial_window_size / 4);
```
MrAnno: Should we compute and store this during init time?

lib/logsource.c
```diff
@@ -87,7 +94,25 @@ _flow_control_window_size_adjust(LogSource *self, guint32 window_size_increment,
...
+  if (wakeup_threshold == 0)
+    wakeup_threshold = 1;
+  gboolean threshold_crossed = window_size_increment != 0 &&
+                               old_window_size < wakeup_threshold &&
+                               new_window_size >= wakeup_threshold;
+  gboolean explicit_suspend_lifted = need_to_resume_counter && old_window_size > 0;
+  if (threshold_crossed || explicit_suspend_lifted)
```
MrAnno, line 112: I wonder what this does to destination-side features, like HTTP batching, the partition autoscaling (which measures rate locally), and disk-buffer (which has some new queueing tricks).

@mitzkia It may be worth running some regression and perf tests on this PR while all of these features  are configured.

bazsi: Batching:

* batching consumes messages from the queue but does not ack them until the batch is delivered
* if the fifo size matches the batch size _exactly_, then we might be a few messages short to fill the batches completely. In this case, we will continue only after `batch_timeout`
* right now we do have cases where log_fifo_size() is carefully tuned for `batch_size * worker_count`

So I think this is a problem. 

1) One solution would be to actually increase the fifo size by wakeup_threshold at initialization. In the worst case, this could increase maximum memory use by 25% (init_window_size / 4 extra elements in the window).

2) alternative is to warn the user about this case until they bump the log-iw-size() parameter to add the extra items.

3) detect somehow that we are slowing down due to this condition and stop the delay altogether until the log-iw-size() is increased to cover it.

lib/logsource.c
bazsi: fixed.

lib/logsource.c
MrAnno: Thanks.

I'm also not sure what this delayed wakeup would do with the autoscaling rate measurement code, but hopefully it just continues to work correctly.

bazsi: hmm... I was trying to reproduce the problem and I have now second thoughts about whether this is actually a problem. 

The messages we delay with this change are always the _first_ few messages of the new batch and not the last one messages of a batch being created. This is important, the old batch can fully form and can be sent without delay.

Once the batch is sent,  ACKs are generated, which should immediately wake up the source once we cross the threshold. Since the window opens up with batch-lines elements, that will ensure that we will also cross the threshold.

The only problem is the batch-size is  smaller than the threshold. let my try

bazsi: I had a nice chat with Claude about this whole setup and it really helped me navigate the source->window empties->wakeup and the destination->batch->flush timelines.

1) if `batch_size` > `log-iw-size()` then we have a stall problem, but that's no change from the previous behaviour.

2) if `batch_size` < `wakeup_threshold` (log-iw-size/4), then every time a batch is flushed, we get a round of ACKs. If the `batch_size` is smaller than the `wakeup_threshold`, then one delivered batch may not cause the source to wakeup as the number of ACKs will not cross the `wakeup_threshold`. But even in this case, we a)  still have at least 75% space in our window or b) already sitting in the queue. In both cases, more batches are guaranteed to form, which will eventually reach the threshold as they get flushed.

3) if `batch_size` == `wakeup_threshold`, every batch should trigger a wakeup.

4) if `batch_size` > `wakeup_threshold`, every batch should trigger a wakeup

5) if `batch_size` == `log-iw-size()` we need the entire window to build a batch, which cause the source to sleep until the message is delivered, causes a processing ping-pong (e.g. either the source or the destination works, their processing is not overlapping). Actually it is good practice to have a slightly bigger log-iw-size() than batch size to avoid this. It also works if there are multiple LogSource instances feeding the same destination. In either case, the sources can continue to operate while the batch is being delivered.

bazsi: The other case wrt message rate measurement for partition autoscaling, I think that's fine too. The above is all about whether we will have a slowdown due to batch_size and wakeup interaction, but I am pretty convinced we don't have that.

autoscaling measures per-partition message rate at log_pipe_queue() time and it tries to measure the relative size of a partition compared to the whole mix. The mix does not really change due to how we wake up sources, and the timing of messages queued should not change significantly either.

MrAnno: Thanks.

I think this could go in before the release, but I would feel safer if @mitzkia could take a quick look at the above with an e2e test.

MrAnno (approved):

lib/logsource.c
MrAnno: Let's do it on main, we'll also need some internal tests anyway.

PR #1098, opened 2026-05-30
filterx/expr-comparison: use string() cast to do string based comparison

bazsi:
Previously we were using the format generated by filterx_object_marshal(), however that does not make much sense and is probably an artifact from the past where string() did not exist. This is a breaking change for datetime and null objects.

Let's discuss if this makes sense or not.

### DIFF
--- a/lib/filterx/expr-comparison.c
+++ b/lib/filterx/expr-comparison.c
@@ -87,8 +87,7 @@ _convert_filterx_object_to_string(FilterXObject *obj, gsize *len)
     }
 
   GString *buffer = scratch_buffers_alloc();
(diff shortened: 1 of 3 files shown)


alltilla:
I'll test this with a bigger config.

OverOrion:
I have successfully tested this with a bigger configuration.

I've cherry-picked the branch and used the following image:
`ghcr.io/overorion/axosyslog:dev-filterx-string-comparison-should-use-string-cast`

bazsi:
@OverOrion based on your testing, can you approve this branch? I added a news entry as well.

OverOrion (approved):

PR #1100, opened 2026-06-01
Light filterx pipeline smoke

mitzkia:
Added new test_filterx_pipeline testcase to use with the previously committed callgrind-syslog-ng.conf and callgrind-samples.log files.

### DIFF
--- a/dbld/callgrind-syslog-ng.conf
+++ b/dbld/callgrind-syslog-ng.conf
@@ -724,7 +724,7 @@ channel {
     };
   };
 };
-destination "devnull-dd4560bf" {
+destination "filedest-dd4560bf" {
   channel {
     filterx {
       classified_metric_labels += {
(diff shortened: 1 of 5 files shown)


MrAnno:
The added file destination changes the callgrind instruction counters:
`1,855,582 -> 2,275,190`

It's +20%, I think it's okay, but let's wait for Bazsi's review.

alltilla (dismissed):

bazsi:
I am fine with that, having a destination driver just makes the result closer to reality. I would maybe check if the map-to-text conversion makes sense in addition.

We will just need to recalibrate the callgrind results in our minds and check against this, btw we started at ~2.5M cycles originally and worked our way down to 1.85M now, so it's time to raise the bar :)

MrAnno (approved):

PR #1103, opened 2026-06-02
Http openobserve error handling

bazsi:
This PR adds a htto-adapter that processes HTTP responses from OpenObserve. Normally OpenObserve returns HTTP 200 even for batches which failed, and unless we check the returned JSON in the response, we won't know if the request worked out or not.

The adapter checks if the "failed" number of events is zero or not and returns an error condition to the HTTP destination on those cases so it can be retried.

### DIFF
--- a/modules/http-adapters/CMakeLists.txt
+++ b/modules/http-adapters/CMakeLists.txt
@@ -13,6 +13,8 @@ set(HTTP_ADAPTERS_SOURCES
     http-adapters-plugin.c
     splunk-adapter.c
     splunk-adapter.h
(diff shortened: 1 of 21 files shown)


alltilla:
Please consider improving the Splunk and OpenObserve related SCLs with the adapters.

bazsi:
> Please consider improving the Splunk and OpenObserve related SCLs with the adapters.

I added these.

bazsi:
I have reviewed the rest of the patches and they look ok to me.
Thanks @alltilla

alltilla (approved):

PR #1105, opened 2026-06-03
autotools: Fix versioning running from git worktree

fernya:
In git worktrees the .git directory is replaced with a file pointing to the real one. The versioning script now recognizes this and acts as it's running in a git directory.

### DIFF
--- a/scripts/version.sh
+++ b/scripts/version.sh
@@ -29,7 +29,7 @@ cd $BASEDIR/../
 
 if [ "$VERSION" != "" ]; then
   echo $VERSION | tr -d '\n'
-elif [ -d .git ] && GIT_VERSION=$(git describe --match 'axosyslog-[0-9]*' --tags --dirty --abbrev=7); then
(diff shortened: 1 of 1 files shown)


scripts/version.sh
```diff
@@ -29,7 +29,7 @@ cd $BASEDIR/../
 
 if [ "$VERSION" != "" ]; then
   echo $VERSION | tr -d '\n'
-elif [ -d .git ] && GIT_VERSION=$(git describe --match 'axosyslog-[0-9]*' --tags --dirty --abbrev=7); then
+elif { [ -d .git ] || { [ -f .git ] && grep -q '^gitdir:' .git; }; } && GIT_VERSION=$(git describe --match 'axosyslog-[0-9]*' --tags --dirty --abbrev=7); then
```
MrAnno: Nit: I think `-e` would be simpler and enough, we don't expect anyone to place a file or link named `.git` there to do weird stuff.

MrAnno (commented):
Thanks for fixing this on worktrees.

MrAnno (approved):

PR #1108, opened 2026-06-03
java/hdfs: fix unreleased lock in send() when file open fails

alltilla:
Backports syslog-ng/syslog-ng#5707.

Original PR description follows.

## Summary

Fixes CodeQL alert [java/unreleased-lock #1051](https://github.com/syslog-ng/syslog-ng/security/code-scanning/1051).

In `HdfsDestination.send()`, the lock was acquired unconditionally but the `finally { lock.unlock() }` block was only reached when `getHdfsFile()` returned a non-null value. If `getHdfsFile()` returned `null` (file open failure), the early `return ERROR` path bypassed the `finally` block, leaving the lock permanently acquired and deadlocking all subsequent `send()`, `flush()`, and `close()` calls.

## Fix

Moved `getHdfsFile()` and all subsequent logic inside the `try` block so the `finally { lock.unlock() }` covers all exit paths.


MrAnno (approved):

MrAnno:
The news entry is not for 5707, but the file is numbered correctly, so I'm merging this. Thanks.

PR #1109, opened 2026-06-03
slog: new implementation of the pseudo-random function and enhancements

alltilla:
Backports syslog-ng/syslog-ng#5614.

Original PR description follows.

The previous implementation allowed an attacker to distinguish between the pseudo-random function (PRF) and a real random function by supplying specially crafted inputs to it. This leads to a predictable way of how the PRF is generating output which should not by allowed by a good PRF. The new implementation provides a variable input length and constant output length PRF based on AES CMAC for key derivation using the current key Ki, i.e. a key expansion of Ki using multiple iterations is performed. Subsequently, with Ki as the key used in the PRF, the input is defined as

(i || label || 0x00 || context || L)

with i as the number of the current iteration, L as the total length of the requested output, label and context for parametrization of the PRF output based on the intended purpose. See NIST reports SP-800-56cr2 and SP.800-108r1 for more information on key derivation schemes with a PRF.

In the previous implementation, the input length was fixed at 16 bytes regardless of the actual length of the input data. This is also fixed in the new implementation.

Minor changes:
    - MAC verification fix
    - error logging macros
    - clean-up
    - file handling (DRY)
    - return values
    - source code formatting
    - use of GLib types

The slog sources are cherry-picked from upstream; the build files keep AxoSyslog's `libsecure-logging-core` naming and a final astyle pass adapts formatting to the fork style.

### DIFF
--- a/modules/secure-logging/secure-logging.c
+++ b/modules/secure-logging/secure-logging.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 Airbus Commercial Aircraft
+ * Copyright (c) 2019-2025 Airbus Commercial Aircraft
  *
  * This program is free software: you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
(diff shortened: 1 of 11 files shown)


alltilla:
NOTE:

I haven't reviewed this PR yet. Just did the cherry-pick and reformatting with the help of claude.

MrAnno:
I'll review this soon.

MrAnno (dismissed):

MrAnno (approved):

PR #1110, opened 2026-06-03
correlation: fix radix parser end-of-input handling

alltilla:
Backports syslog-ng/syslog-ng#5690.

Fixes two related radix matcher edge cases at end of input.

- Parser scans now stop before '\0' to avoid reading past end-of-input and to keep captured lengths correct.
- Parser-node traversal now continues with empty remaining input, so OPTIONALSET children can still match.

Also fixes a regression caused by 5b77cceef signaled by @smortex

Upstream issues: syslog-ng/syslog-ng#5678, syslog-ng/syslog-ng#5710

---

Note: the pdb_rule_set_load double-free fix is defensive-only in AxoSyslog — the regression cause (5b77cceef) is not present here, since pdb_rule_set_load does not call _pdb_loader_cleanup() and so never double-frees state.examples. The accompanying tests remain useful regardless.

### DIFF
--- a/modules/correlation/pdb-load.c
+++ b/modules/correlation/pdb-load.c
@@ -1231,7 +1231,11 @@ pdb_rule_set_load(PDBRuleSet *self, GlobalConfig *cfg, const gchar *config, GLis
     }
 
   if (state.load_examples)
-    *examples = state.examples;
+    {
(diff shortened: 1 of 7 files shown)


MrAnno (approved):
Nice.

PR #1113, opened 2026-06-03
journald-source: add read_old_on_error() option to control where to continue after a position restore attempt failure

MrAnno:
Cherry-picked from syslog-ng/syslog-ng#5647


alltilla (approved):
Would be worth a news entry.

bazsi (approved):

PR #1114, opened 2026-06-03
Fix sql crash on dberror

MrAnno:
Cherry-picked from syslog-ng/syslog-ng#5696


alltilla (dismissed):

MrAnno:
I need a NEWS update here

alltilla (approved):

PR #1118, opened 2026-06-04
Fix scope sync

MrAnno:
A feature called "scope reuse" has recently been removed in favor of variable moving and O(1) lookup.

filterx_scope_sync() dependended on scope reuse in a hidden way: it was only called on fork points and when a non-filterx block was inserted after a FilterX block; but sync() itself was not recursive, it only synced the latest scope, not all the consecutive scopes.

This commit fixes it by adding the recursion: going back until the last fork point, syncing everything. `scope->dirty = FALSE` makes sure scopes are synced only once, for example, when a rewrite rule is inserted between 2 FilterX blocks.

Note: This is a temporary fix, a cleaner idea will be implemented later: allocating one single scope (and a merged precalculated variable layout) for all consecutive blocks. This basically reintroduces the "scope reuse" feature without losing O(1) lookups, so the end result will be even faster.

### DIFF
--- a/lib/filterx/filterx-scope.c
+++ b/lib/filterx/filterx-scope.c
@@ -101,6 +101,9 @@ _pull_variable_from_parent_scope(FilterXScope *self, FilterXVariablePullMode pul
         }
     }
 
+  if (pull_mode == FX_VAR_PULL_MOVE)
+    parent_variable->variable_type = FX_VAR_NONE;
+
   msg_trace("Filterx scope, pulling scope variable",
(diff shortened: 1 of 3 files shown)


MrAnno:
`git diff --ignore-space-change` is recommended when reviewing this PR.

MrAnno:
### Callgrind

O(1) PR: 1,904,141 -> 1,856,165; diff: `-47976`
This PR: 2,271,338 -> 2,286,563; diff: `15225`

So we are still fine.

alltilla (approved):

PR #1119, opened 2026-06-05
github-actions: add callgrind instruction count comment job

alltilla:
Posts a single comment (and edits it if necessary) if we modified filterx related, non-test code.

PRs for reference:
* https://github.com/alltilla/axosyslog/pull/1
* https://github.com/alltilla/axosyslog/pull/2
* https://github.com/alltilla/axosyslog/pull/3
* https://github.com/alltilla/axosyslog/pull/4

The only problem is that the measurement is not fix (see the comments in PR 2, 3, 4). There are small differences in the base measurement, too. We might want to fix that.

### DIFF
--- a/.github/workflows/comment-callgrind-changes.yml
+++ b/.github/workflows/comment-callgrind-changes.yml
@@ -1,0 +1,126 @@
+name: Comment callgrind changes (PR)
+
+on:
+  workflow_run:
+    workflows: [CI @ devshell]
+    types:
+      - completed
+
+concurrency:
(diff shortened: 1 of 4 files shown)


alltilla:
Aikido is false alarm, we use `workflow_run` the safe way.

MrAnno:
I think we broke the stability by adding a file destination. We'll fix it with Micek.

MrAnno:
#1120

bazsi:
I love this!

MrAnno (approved):

PR #1121, opened 2026-06-06
Filterx drop generic eval errors

bazsi:
This should reduce our error stack size, as the 32 elements in FilterXEvalContext takes over 1kB of space for all filterx blocks in sequence. Also, this contains cleanup steps and a more readable error stack in general, plus some refactors in expr-plus and expr-arithmetic-operators.


lib/filterx/filterx-eval.c
```diff
@@ -100,102 +76,70 @@ _push_new_error(FilterXEvalContext *context)
...
+filterx_eval_update_error_location_from_expr(FilterXExpr *expr)
 {
-  if (!error)
-    return;
+  FilterXEvalContext *context = filterx_eval_get_context();
 
-  gboolean is_error_local = !context;
-  if (is_error_local)
-    filterx_error_clear(error);
+  FilterXError *error = &context->errors[context->error_count - 1];
```
alltilla: If we forget to push an error in an expr, but we still return `NULL`, this function receives `context->error_count == 0`, and indexes `-1` here.

We should either have an assert (might be too harsh) or a safety guard.

lib/filterx/filterx-eval.c
```diff
@@ -321,7 +265,8 @@ _fill_failure_info(FilterXEvalContext *context, FilterXExpr *block, FilterXObjec
 
   if (!error->message && context->failure_info_collect_falsy)
     {
-      filterx_falsy_error_set_values(&failure_info->errors[0], "Falsy expression", block, block_res);
+      filterx_falsy_error_set_values(&failure_info->errors[0], "Falsy expression", block_res);
+      filterx_eval_update_error_location_from_expr(block);
```
alltilla: Do we want to modify the `context->errors` here or the `failure_info->errors`. We do the former, but I believe we want the latter.

lib/filterx/filterx-eval.c
bazsi: good catch

alltilla (dismissed):

alltilla:
Thanks for the fixes.

lib/filterx/filterx-eval.c
bazsi: I tried to come up with a case where this is triggered, but could not find any. With that said, I fixed this as well. thanks for pointing it out.

alltilla (approved):

alltilla:
https://github.com/axoflow/axosyslog/pull/1166 will workaround the flaky light smoke test.

PR #1123, opened 2026-06-09
FilterX: fix borrowed/indirect objects getting overwritten by scope syncs

alltilla:
These bugs have been in FilterX for a while, not recent regressions.
I did a rename of "borrowed" -> "indirect", which follows the naming of `NVTable`.

### DIFF
--- a/lib/filterx/expr-template.c
+++ b/lib/filterx/expr-template.c
@@ -53,7 +53,7 @@ _eval_template(FilterXExpr *s)
    * expressions, thus the FilterXObject is shorter lived than the scratch
    * buffer.  */
 
-  return filterx_message_value_new_borrowed(value->str, value->len, t);
+  return filterx_message_value_new_indirect(value->str, value->len, t);
 }
 
 static void
--- a/lib/filterx/filterx-object.h
+++ b/lib/filterx/filterx-object.h
@@ -204,7 +204,7 @@ struct _FilterXObject
    *
    */
   guint weak_referenced:1, is_dirty:1, allocator_used:1, floating_ref:1, early_allocation:1, early_allocation_checked:1,
-        flags:5;
+        is_nvtable_backed:1, flags:5;
   volatile guint32 hash;
   FilterXType *type;
 };
@@ -754,6 +754,12 @@ filterx_object_dup(FilterXObject *self)
     return filterx_ref_dup(self);
 
   return self->type->clone(self);
+}
+
+static inline gboolean
+filterx_object_is_nvtable_backed(FilterXObject *self)
+{
+  return self->is_nvtable_backed;
 }
 
 static inline FilterXObject *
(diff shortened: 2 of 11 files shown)


MrAnno:
`make_direct()` is created to fix an issue caused by an optimization made for strings and message values.

We agreed to refactor this to use `clone()` directly + an indirect flag.

alltilla:
Ready for review.

lib/filterx/filterx-scope.h
```diff
@@ -90,17 +90,31 @@ filterx_scope_get_variable(FilterXScope *self, FilterXVariable *v)
...
     {
       G_STATIC_ASSERT(sizeof(v->generation) == sizeof(self->generation));
+      if (filterx_variable_is_declared(v))
+        filterx_scope_make_object_direct(value);
       filterx_variable_set_value(v, value, assignment, self->generation);
     }
   else
     {
       G_STATIC_ASSERT(sizeof(v->generation) == sizeof(self->msg->generation));
+      filterx_scope_make_object_direct(value);
```
MrAnno: As far as I understand, keeping indirect objects is only a problem if a LogMessage-synced variable is involved.

If this is true, I think those "make_direct" calls may not belong to `filterx_scope_set_variable()`.

For example,
```
log {
    source(genmsg); // "values.int" => int(5)
    filterx { 
                declare captured = 0;
                captured = ${values.int};
                ${values.int} = 7;
             };;
    rewrite { set-tag("force-sync"); };
    if {
        filterx { captured == 5; };
        destination(dest_true);
    } else {
        
        destination(dest_false);
    };
    
    
    destination(dest_final);
};
```


The problem there is that we assumed that `${values.int}` is immutable. It's true for FilterX, but it's not true for LogMessage fields. And the problem only happens when we exit a filterx block, because we need to trigger that mutable logic inside LogMessage with a sync.

Wouldn't it be the real fix if we called `dup()` on all declared and $ variables that reference a LogMessage variable? The location should probably be `filterx_scope_sync()`, but in a separate variable loop, because we shouldn't call `log_msg_set_value_with_type()` until we finished duping all relevant variables.

alltilla: Maybe I generalized it too much. Should we tag and carry if a value is pointing to a LogMessage field?

E.g. for this case:
* `LogMessage` value -> `MessageValue` -> `String` (slice) 

The `String` object only knows it is indirect currently, which is I guess okay, but:
* `String` -> `String` (slice)

This one thinks it is indirect, too, and gets copied, but it does not need to probably.

MrAnno: Let's discuss this IRL sometime to make sure I didn't miss anything.

But my current understanding is that we don't need to track FilterX-only indirections, because everything works automatically, we just need to fix this "LogMessage has mutable fields, FilterX doesn't" asymmetry around indirect values.

alltilla: Fixed, I think.

MrAnno: Thanks.

MrAnno (approved):

PR #1124, opened 2026-06-12
Fix/invalid frame metrics crash

OverOrion:
TODO:

- [x] NEWS file

### DIFF
--- a/lib/logproto/logproto-auto-server.c
+++ b/lib/logproto/logproto-auto-server.c
@@ -24,6 +24,7 @@
 #include "logproto-framed-server.h"
 #include "logproto.h"
 #include "transport/logtransport.h"
+#include "stats/stats-cluster-key-builder.h"
 #include "messages.h"
 #include "str-utils.h"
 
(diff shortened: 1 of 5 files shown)


lib/logproto/logproto-auto-server.c
```diff
@@ -137,6 +149,7 @@ log_proto_auto_server_new(LogTransport *transport, const LogProtoServerOptions *
   log_proto_server_init(&self->super, transport, options);
   self->super.handshake = log_proto_auto_handshake;
   self->super.poll_prepare = log_proto_auto_server_poll_prepare;
-  self->kb = kb;
+  self->super.free_fn = log_proto_auto_server_free;
+  self->kb = kb ? stats_cluster_key_builder_clone(kb) : NULL;
```
MrAnno, line 153: Don't we need a copy for the framed proto as well? To be precise: don't we need to copy for all protos (auto, framed, all the future ones)?

`DEFINE_LOG_PROTO_SERVER_WITH_KB()` could do such a thing, but that would suggest that this kb field should rather be a field of the `LogProtoServer` abstract class and the copy and free should happen inside that class.

OverOrion: I don't think we need: `LogProtoFramedServer` does not hold the `kb` as member variable. It only uses it during its constructor. I could not crash it the same way either.

The others used it later and created new labels with the `kb`.

Let me know if I missed something though. :eyes:

MrAnno: You're right, thanks :)

MrAnno (approved):

PR #1125, opened 2026-06-15
messages: fix infinite loop while producing the "looping back" message

MrAnno:
While sending the "internal() messages are looping back" message, _grow_payload() produces a new internal message, which causes an infinite recursion.


OverOrion (approved):

MrAnno:
Let's wait for @mitzkia, because I couldn't reproduce the issue.

mitzkia (approved):
Verified. 
I have tested this locally, and the original issue has been fixed with this..

PR #1127, opened 2026-06-16
bigquery: fix crash (GRPC_CALL_ERROR_TOO_MANY_OPERATIONS) on server-side stream restart

### LINKED ISSUE #957: BigQuery destination crashes with GRPC_CALL_ERROR_TOO_MANY_OPERATIONS on server-side stream restart (segfault, exit code 139)
# AxoSyslog
## Version of AxoSyslog

4.22.0

## Platform

Kubernetes; GCP; linux;

# Issue
## Failure

At times it prints this message and Pod restarts.

GRPC_CALL_ERROR_TOO_MANY_OPERATIONS
Check failed: false
Error closing BigQuery stream; error='Closing the stream because server is restarted'

K8 Event Logs:
    Last State:     Terminated
      Reason:       Error
      Exit Code:    139
      Started:      Thu, 19 Feb 2026 18:33:08 -0700
      Finished:     Thu, 19 Feb 2026 23:41:45 -0700
    Ready:          True

## Configuration

@version: current
@include "scl.conf"

options {
  log_fifo_size(100000);
  log_msg_size(262144);
  stats(freq(10));
  time_reopen(10);
  use_fqdn(no);
  use_dns(no);
  chain_hostnames(no);
  keep_hostname(yes);
  keep_timestamp(yes);
};

source s_internal { internal(); };

destination d_bigquery_rp {
  bigquery(
    project("${GOOGLE_CLOUD_PROJECT}")
    dataset("d_id")
    table("rp")
    workers(2)
    batch-timeout(3000)
    batch-lines(20)
    schema(
      "timestamp" STRING => "${TIMESTAMP}"
      "trace_id" STRING => "${TRACE_ID}"
      "func_name" STRING => "${FUNC_NAME}"
      "level" STRING => "${LOG_LEVEL}"
      "request_payload" STRING => "${REQUEST_PAYLOAD}"
      "jsonResponse" JSON => "${JSON_RESPONSE}"
    )
    on-error("drop-property")
  );
};

log {
  source(s_otlp);
  filterx {
    declare log = otel_logrecord(${.otel_raw.log});
    log.attributes._bigquery_table == "retrieval_prompts";
    ${TIMESTAMP} = isset(log.attributes.timestamp) ? string(log.attributes.timestamp) : "";
    ${TRACE_ID} = isset(log.attributes.trace_id) ? string(log.attributes.trace_id) : "";
    ${FUNC_NAME} = isset(log.attributes.func_name) ? string(log.attributes.func_name) : "";
    ${LOG_LEVEL} = isset(log.attributes.level) ? string(log.attributes.level) : "";
    ${REQUEST_PAYLOAD} = isset(log.attributes.request_payload) ? format_json(log.attributes.request_payload) : "";
    ${JSON_RESPONSE} = isset(log.attributes.jsonResponse) ? format_json(log.attributes.jsonResponse) : null;
  };

  destination(d_bigquery_rp);
};

- **Destination**: BigQuery Storage Write API

sagarswamirao:
## Summary

Fixes #957.

The `bigquery()` destination crashes with a SIGSEGV (exit code 139) when the BigQuery Storage Write API performs a server-side restart of the `AppendRows` stream. On a busy destination this turns into a crash loop.

## Root cause

Each worker opens one bidirectional `AppendRows` streaming call (`batch_writer`) and half-closes it with `WritesDone()` + `Finish()` in `disconnect()`.

When the server restarts the stream (it may do so at any time and signals the client to reconnect, e.g. `"Closing the stream because server is restarted"` / `RST_STREAM`), the in-flight `Write()`/`Read()` in `flush()` fails and the gRPC call reaches a terminal state. `disconnect()` then unconditionally ran the `WritesDone()`/`Finish()` half-close handshake on that already-completed call. Queuing a terminal operation on a call that has already finished makes gRPC core fail a `CHECK`:

```
call_op_set.h: API misuse of type GRPC_CALL_ERROR_TOO_MANY_OPERATIONS observed
call_op_set.h: Check failed: false
```

which `abort()`s the whole process.

## Fix

- Track streaming-call failure with a `batch_writer_failed` flag, set when `Write()`/`Read()` returns `false` in `flush()`.
- In `disconnect()`, skip the `WritesDone()`/`Finish()` half-close handshake when the writer has already failed; the worker tears the call down and reconnects cleanly instead.
- `FinalizeWriteStream()` is an independent unary RPC on its own `ClientContext` and is left untouched, so stream finalization still happens.
- The flag is reset whenever a fresh `batch_writer` is created.

The half-close decision is factored into a small pure predicate (`bigquery_should_half_close_writer()`) so the invariant is covered by a unit test without needing a live gRPC stream.

## Tests

- `modules/grpc/bigquery/tests/test-bigquery-worker.cpp` (Criterion) asserts the half-close handshake runs only when the writer is connected and has not failed, and is skipped once the stream has failed — the exact condition that previously crashed.
- Wired into both the CMake and autotools builds, guarded by `ENABLE_GRPC`, matching the existing `otel` test layout.

## Notes

- News fragment added (`news/bugfix-1127.md`).
- Commit is DCO signed-off.


sagarswamirao:
cc @alltilla @MrAnno @bazsi for a quick review when you have a moment.

@alltilla — this is the same server-restart / reconnect failure mode you fixed for the arrow-flight destination; the BigQuery worker's `disconnect()` just wasn't guarding the half-close handshake against an already-failed stream. Happy to adjust the approach (e.g. fold the flag into the shared gRPC base worker, or change the test style) if you'd prefer a different shape.

alltilla:
Very nice! Thank you for bringing in the issue, the root cause analysis and a fix as well!

This is a good direction, but I think the rebuilding of the stream is not handled with your fix yet.
If you are okay with it, I would put some commits atop of yours to implement and test that.

sagarswamirao:
Thanks. Yes sure, please feel free to

MrAnno (dismissed):

tests/light/src/axosyslog_light/driver_io/bigquery/bigquery_io.py
```diff
@@ -0,0 +1,216 @@
...
+        return self.__servicer.get_batch_count(table) if self.__servicer else 0
+
+    def get_stream_count(self, table: str) -> int:
+        return self.__servicer.get_stream_count(table) if self.__servicer else 0
+
+    def get_create_count(self, table: str) -> int:
+        return self.__servicer.get_create_count(table) if self.__servicer else 0
+
+    def get_finalize_count(self, table: str) -> int:
+        return self.__servicer.get_finalize_count(table) if self.__servicer else 0
```
mitzkia, line 271: I think we have to add  `read_logs()` like API also. And we have to assert with it for the payload in the testcases

alltilla: Fixed. If you find it good, kindly approve and merge. I didn't change any code MrAnno previously approved.

mitzkia (approved):
Thanks for the light changes.

PR #1131, opened 2026-06-19
light: keep the fd-leak baseline in-process

alltilla:
The fd-leak baseline is per-process state but was stored in the shared, file-backed session_data under a single key, so under xdist every worker checked its own fd count against the first worker's snapshot and failed.

Keep it in a module-level variable instead, removing the cross-worker race.


mitzkia (approved):
Thanks for the fix. Approved

PR #1133, opened 2026-06-22
`network()` dest: `flush-lines()`/`writev()` support

alltilla:
This improves performance in certain scenarios, like the one in https://github.com/VirtualMetric/PipeBench

No `framed` support yet, we might want to implement that later.

### DIFF
--- a/lib/logproto/logproto-text-client.c
+++ b/lib/logproto/logproto-text-client.c
@@ -25,6 +25,7 @@
 
 #include <errno.h>
 #include <limits.h>
+#include <string.h>
 
 static LogProtoStatus log_proto_text_client_flush(LogProtoClient *s);
 
@@ -36,7 +37,7 @@ log_proto_text_client_poll_prepare(LogProtoClient *s, GIOCondition *cond, GIOCon
   *cond = G_IO_OUT | G_IO_IN;
   *idle_cond = G_IO_IN;
 
-  return self->partial != NULL;
+  return self->partial != NULL || self->batch.count > 0;
 }
 
 static gboolean
@@ -48,7 +49,7 @@ log_proto_unidirectional_text_client_poll_prepare(LogProtoClient *s, GIOConditio
   *cond = G_IO_OUT;
   *idle_cond = 0;
 
-  return self->partial != NULL;
+  return self->partial != NULL || self->batch.count > 0;
 }
 
 static LogProtoStatus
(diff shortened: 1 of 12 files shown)


lib/transport/transport-tls.c
```diff
@@ -388,6 +390,21 @@ log_transport_tls_write_method(LogTransport *s, const gpointer buf, gsize buflen
...
+
+  if (G_UNLIKELY(!self->writev_buf))
+    self->writev_buf = g_byte_array_new();
+
+  g_byte_array_set_size(self->writev_buf, 0);
+  for (gint i = 0; i < iov_count; i++)
+    g_byte_array_append(self->writev_buf, iov[i].iov_base, iov[i].iov_len);
+
+  return log_transport_tls_write_method(s, self->writev_buf->data, self->writev_buf->len);
+}
```
bazsi, line 455: does this also have a performance advantage? we could just flush the first iov element and then return so the higher layers can continue with the next iov, so we could avoid having to copy the entire payload another time.

lib/logproto/logproto-text-client.c
```diff
@@ -146,12 +146,111 @@ log_proto_text_client_flush(LogProtoClient *s)
...
+      ofs += self->batch.iov[i].iov_len;
+      ++i;
+    }
+
+  self->partial_pos = 0;
+  self->partial_free = (GDestroyNotify) g_free;
+  self->partial_msgs = self->batch.count - first_non_written_chunk_index;
+
+  log_proto_client_msg_ack(&self->super, self->batch.count - self->partial_msgs);
+}
```
bazsi, line 185: I think we could avoid copying the payload by keeping a slightly more complex state. instead of rolling up the payload into "partial", just keep the current index and the position inside that index. Again this would avoid having to copy the payload.

I am not sure how often we get a truncated result, but I'd assume that with 100 messages, 1kB each, 100kB is often not flushed in a single iteration.

lib/transport/transport-tls.c
bazsi: the SSL protocol has 16kB records (https://chatgpt.com/share/6a3ad4fa-0b20-83ed-9a16-5cc5b4b92dcc), and while SSL_write() will split the buffer into multiple records automatically, there's probably not much benefit in buffering huge chunks here

bazsi (commented):
I mostly like what I see here. I'd try to avoid the copying, especially if we need to do that in both the LogProto and the LogTransport layers.

lib/transport/transport-tls.c
alltilla: Yes it improves perf even with `flush-lines(2)`. (4x perf with the default 100)

I have added 2 things though:
1. batch size 1 does not copy.
2. batched buffer is now capped at 16K.

lib/logproto/logproto-text-client.c
alltilla: Agree, implemented this change.

alltilla:
Sorry for the AI generated response, but I could not make this good perf measurement summary myself. These are after fixing your comments, Bazsi.

---

## Benchmark: `flush-lines()` write coalescing on `network()` destinations

### Setup
Machine: My laptop.

`loggen` (4 connections, 256 B messages, max rate) → `network()` TCP source → `network()` TCP/TLS destination → draining sink.
Single dest connection, `flags(flow-control)`, `log-fifo-size(1000000)`.
Steady-state **median msg/s over 3 × 20 s reps**.
Baseline = `main`

### Throughput (median msg/s, 3 reps)

| Config | TCP | vs base | TLS | vs base |
|---|---:|---:|---:|---:|
| `main` (baseline) | 390,690 | 1.00× | 286,680 | 1.00× |
| `@4.26` default (`flush-lines(100)`) | 1,208,050 | **3.09×** | 1,126,150 | **3.93×** |
| `@4.0` fallback | 400,019 | 1.02× | 293,303 | 1.02× |
| `@4.26` `flush-lines(2)` | 590,770 | 1.51× | 469,236 | 1.64× |
| `@4.26` `flush-lines(1)` | 393,794 | 1.01× | 289,638 | 1.01× |

- Default coalescing gives **3.1× (TCP)** and **3.9× (TLS)**.
- Old configs (`@4.0`) and explicit `flush-lines(1)` stay at baseline — no regression on the per-message path.
- Coalescing just two messages already wins: `flush-lines(2)` vs `flush-lines(1)` = **+50% (TCP)**, **+62% (TLS)**.

### Destination-side write syscalls (`strace -f -c`, messages per dominant syscall)

| Config | TCP | TLS |
|---|---|---|
| `main` baseline | `sendto`, 1.0/msg | `sendto`, 1.0/msg |
| `@4.0` fallback | `sendto`, 1.0/msg | `sendto`, 1.0/msg |
| `@4.26` `flush-lines(1)` | `sendto`, 1.0/msg | `sendto`, 1.0/msg |
| `@4.26` default | `writev`, **96.4**/call | `sendto`, **49.5**/call |
| `@4.26` `flush-lines(2)` | `writev`, **2.0**/call (+1 `sendto` tail) | `sendto`, **2.0**/call |

Single-message flushes go out as `sendto` (never `writev(1)`) on both transports;
The `@4.0`/`flush-lines(1)` paths are byte-identical to `main`.
TLS coalesces into ~16 KB records (record cap).

alltilla:
I am chasing some bugs. I will push the changes once they are fixed.

alltilla:
@bazsi Fixed your comments and a lot of other things. The history got rewritten pretty hard, I would ask for a full re-review if you have the time. Thanks!

PS: `framed` is not yet supported, although it would be a big win, I think we should come back to that later.

lib/logproto/logproto-text-client.c
bazsi: great, thanks

lib/transport/transport-tls.c
```diff
@@ -333,9 +339,14 @@ log_transport_tls_write_method(LogTransport *s, const gpointer buf, gsize buflen
   gint ssl_error;
   gint rc;
 
+  /* OpenSSL requires a blocked SSL_write() to be repeated with the identical
+   * buffer; replay the pending one, ignoring whatever the caller passes now */
+  gconstpointer write_buf = self->write_blocked ? self->write_blocked_buf : (gconstpointer) buf;
+  gsize write_len = self->write_blocked ? self->write_blocked_len : buflen;
+
```
bazsi: how can we be sure that the buffer remains valid at the next invocation? is this a writev() specific workaround?

lib/logproto/logproto-text-client.c
```diff
@@ -200,7 +201,12 @@ _flush_batch(LogProtoTextClient *self)
   iov[0].iov_base = (guchar *) first.iov_base + self->batch.partial_pos;
   iov[0].iov_len = first.iov_len - self->batch.partial_pos;
 
-  gssize rc = log_transport_stack_writev(&self->super.transport_stack, iov, iov_count);
+  /* a single chunk doesn't need vectored I/O */
+  gssize rc;
+  if (iov_count == 1)
+    rc = log_transport_stack_write(&self->super.transport_stack, iov[0].iov_base, iov[0].iov_len);
+  else
+    rc = log_transport_stack_writev(&self->super.transport_stack, iov, iov_count);
```
bazsi: I think this should be part of the LogTransport layer. For what it's worth, the transport layer knows much better if using write() over writev() situations makes more sense. It's an implementation detail of LogTransportTLS that it was slower (due to the payload copy), so that's where I'd put this workaround, not here.

lib/transport/transport-tls.c
alltilla: Yes it is `writev` specific. I moved it to the `writev_method`, to make its lifetime cleaner.

lib/logproto/logproto-text-client.c
alltilla: Sure, moved it to TLS.

alltilla:
I would consider NOT merging this before releasing 4.26. We might want to keep it running for a while in nightlies afetr 4.26.

bazsi (approved):
great work, thank you.

PR #1134, opened 2026-06-22
[FilterX-jit] [Devirtualization] Add static type inference

fernya:
This pull request adds a static type inference algorithm to the variables whose types can be decided during configuration parsing time.
The inferred types will be used by the JIT code emission code to build specialized IRs instead of emit opaque function calls. We have proven that when it's used with LLVM-backed variable management, the throughput of the axosyslog improves in a meaningful way.

Key build-time configuration - `#define FILTERX_ATTR_CHAIN_MAX 16`

_One caveat will be resolved in later PR: when an expression does `otel_kvlist` member assigment (LHS) to a `dict` (RHS) we infer `dict` static type. However the memory layout differs between the two types. In order to keep chained get/setattr optimizations in-place, we'll add runtime checks for this case of devirtualization._

# Changes summarized as examples

## 1. Literal scalars — `expr-literal.c`

```filterx
filterx {
    a = "hello";   # resolves to STRING
    b = 42;        # resolves to INTEGER
    c = 2.5;       # resolves to DOUBLE
};
```

`DOUBLE` is a kind of its own rather than a flavour of `INTEGER`. The two are not
interchangeable at runtime — `filterx_object_add()` widens `int + double` to double and
`filterx_integer_unwrap()` rejects a double outright — so a value that may be either has to
meet to `UNKNOWN` rather than to `INTEGER`. Consumers that speculate on an integer
representation depend on that distinction staying visible (see §4).

## 2. Literal containers + element meet — `expr-literal-container.c`

```filterx
filterx {
    # resolves to homogeneous -> element type is committed
    s_dict = {"a": "x", "b": "y"};          # DICT_OF_STRING
    s_list = ["x", "y", "z"];               # LIST_OF_STRING

    # resolves to no depth cap on the element chain
    deep = {"l1": {"l2": {"l3": "leaf"}}};  # DICT->DICT->DICT->STRING (all 3 levels + leaf)

    # can't resolve /  heterogeneous values -> element meets to UNKNOWN (outer kind DICT survives)
    mixed = {"a": "x", "b": [1, 2]};        # DICT, element UNKNOWN

    # can't resolve /  empty container -> element is FRESH, sanitizes to UNKNOWN until first write
    empty = {};                             # DICT, element UNKNOWN
};
```

## 3. Assignment propagation — `expr-assign.c`

```filterx
filterx {
    # resolves to RHS type flows to the variable and to later reads
    x = "v";
    y = x;             # x read here is STRING

    # resolves to last write wins (reassignment with a different type)
    z = {};
    z = "v";
    w = z;             # z is STRING (the dict assignment is overwritten)

    # can't resolve /  RHS of unknown type makes the variable unknown
    u = get_thing();   # function call -> UNKNOWN
    v = u;             # u is UNKNOWN
};
```

## 4. `+` operator — `expr-arithmetic-operators.c`

Addition gets a result type of its own rather than a plain meet of its operands, because
`filterx_object_add()` promotes to double as soon as either side is one.

```filterx
filterx {
    # resolves to string + string -> STRING
    full = "foo" + "bar";

    # resolves to int + int -> INTEGER
    a = 1;
    b = 2;
    n = a + b;                     # INTEGER

    # resolves to numeric promotion: either side DOUBLE -> DOUBLE (a plain meet would lose this)
    d = 2.5;
    mixed  = a + d;                # DOUBLE
    mixed2 = d + a;                # DOUBLE (operand order does not matter)

    # can't resolve /  an UNKNOWN operand may be a double at runtime -> UNKNOWN
    total = a + unknown_field;     # UNKNOWN, *not* INTEGER

    # can't resolve /  mismatched kinds -> meet collapses to UNKNOWN
    bad = some_str + [1, 2];       # STRING vs LIST -> UNKNOWN
};
```

**Both operand kinds must be statically numeric for promotion to fire.** An earlier revision of
this branch instead claimed `INTEGER` whenever the *LHS* was `INTEGER`, on the grounds that the
JIT integer fast path can fall back for a non-integer RHS. The fallback does keep evaluation
correct, but the inferred type is not only consumed to pick that fast path — it is also recorded
for the assigned variable and read back as fact by whatever uses it next:

```filterx
x = int_field + double_field;   # runtime: DOUBLE, but was claimed INTEGER
y = x + 1;                      # LHS believed INTEGER -> integer fast path on a double
```

The second statement is the problem, and it is where the damage surfaces — one statement after
the bad claim. `fx_jit_do_plus_int` guards its RHS, but on the fast branch it unwraps its LHS as
an `int64` having already been promised it is one, so a stale claim aborts on that assertion (or
reads uninitialised storage under `G_DISABLE_ASSERT`). Reading `UNKNOWN` as "probably integer" is
what made this reachable, so inference no longer does it — at the cost of the old rule's
propagation through chains containing unknown fields, which was never sound to begin with.

## 5. `+=` and `??=` — `expr-plus-assign.c` / nullv-assign in `expr-assign.c`

```filterx
filterx {
    # resolves to same-kind += keeps the type (meet of prior and RHS)
    msg = "a";
    msg += "b";          # STRING

    # can't resolve /  mismatched += collapses the variable to UNKNOWN
    msg2 = "a";
    msg2 += [1];         # meet(STRING, LIST) -> UNKNOWN

    # resolves to ??= with same-kind RHS keeps the kind in the env
    name = "x";
    name ??= "default";  # env: STRING (meet of prior STRING and RHS STRING)
};
```

`+=` still uses the plain meet, so `int += double` conservatively yields `UNKNOWN` rather than
`DOUBLE`. That is sound, just less precise than §4; extending promotion to it is a follow-up.

## 6. `getattr` (`.key`) — `expr-getattr.c`

```filterx
filterx {
    d = {"a": {"b": {"c": "leaf"}}};
    # resolves to chain shifts one nesting level per hop
    leaf = d.a.b.c;            # DICT->DICT->DICT->STRING, so d.a.b.c is STRING

    # resolves to per-key tracking survives a heterogeneous sibling
    foo = {};
    foo.a = {};
    foo.b = "s";              # smears foo's *element* to UNKNOWN...
    x = foo.a;                # ...but per-key map keeps foo.a precisely DICT

    # can't resolve /  attr-chain tracking stops past FILTERX_ATTR_CHAIN_MAX (16) hops
    y = root.k0.k1 ... .k19;  # ~20 hops: past the cap -> UNKNOWN
};
```

Reads past the cap are untracked rather than wrong: within-cap prefixes still resolve, and the
deeper hops simply contribute no information (no false propagation, no crash).

## 7. `get_subscript` (`[k]`) — `expr-get-subscript.c`

```filterx
filterx {
    d = {"a": "x", "b": "y"};
    v = d["a"];        # resolves to element of DICT_OF_STRING -> STRING

    l = ["p", "q"];
    e = l[0];          # resolves to element of LIST_OF_STRING -> STRING

    # can't resolve /  subscript of a scalar / unknown -> UNKNOWN
    s = "string";
    bad = s[0];        # operand is STRING (not a container) -> element UNKNOWN
};
```

## 8. Incremental build via `setattr` / `set_subscript` — `expr-setattr.c` / `expr-set-subscript.c`

```filterx
filterx {
    # resolves to lift-on-write: empty container's FRESH element is committed by the first write
    foo = {};
    foo.a = {};
    foo.a.b = {};     # foo is DICT->DICT->DICT
    deep = foo.a.b;   # resolves to DICT (incremental build matches a nested literal)

    # resolves to list lift-on-write
    l = [];
    l[0] = {};
    first = l[0];      # DICT

    # can't resolve /  heterogeneous write pessimizes the *element* type
    d = {"a": "x"};    # DICT_OF_STRING
    d["b"] = 42;       # meet(STRING, INTEGER) -> element UNKNOWN
    all = d;           # DICT, element UNKNOWN

    # can't resolve /  write past the depth cap is a safe no-op (within-cap levels still resolve)
    m = {};
    m.a = {}; m.a.b = {}; m.a.b.c = {};
    ...
    m.a.b.c.d.e.f.g.h.i.j.k.l.m.n = {};
    m.a.b.c.d.e.f.g.h.i.j.k.l.m.n.o = {};    # past the cap: no false propagation, no crash
};
```

## 9. Conditional `if/else` (branch meet) — `expr-condition.c`

```filterx
filterx {
    # resolves to both branches agree -> type kept after the if
    if (cond) { z = "a"; } else { z = "b"; }
    r1 = z;            # STRING

    # can't resolve /  branches disagree -> meet to UNKNOWN
    if (cond) { w = "a"; } else { w = [1, 2]; }
    r2 = w;            # UNKNOWN

    # can't resolve /  one-sided write -> meet with the pre-branch (unknown) state -> UNKNOWN
    if (cond) { u = "a"; }
    r3 = u;            # UNKNOWN (else-path leaves u untouched/unknown)

    # can't resolve /  INTEGER and DOUBLE are distinct kinds, so a numeric disagreement
    #    collapses too -- which is what stops an integer-only fast path claiming the result
    if (cond) { m = 1; } else { m = 2.5; }
    r4 = m;            # UNKNOWN
};
```

## 10. `switch` (pessimistic body meet) — `expr-switch.c`

```filterx
filterx {
    # can't resolve /  switch body writes are only kept if they'd also hold on the no-match path;
    #    v1 meets the body back conservatively, so a write inside a case collapses.
    switch (sel) {
        case "x": t = "str";
        default:  t = "other";
    }
    after = t;         # UNKNOWN (body env met against the pre-switch env)

    # resolves to ...unless the pre-switch state already agrees, which makes the meet a no-op
    t2 = "seed";
    switch (sel) {
        case "x": t2 = "a";
        default:  t2 = "b";
    }
    after2 = t2;       # STRING
};
```

The selector and case-value reads still see the env; only the body's writes are pessimized.
Note that the body is one fallthrough-flattened compound expression, so cases are walked in
program order rather than met branch-by-branch the way `if`/`else` are — a `default` therefore
does not currently count as covering the no-match path.

## 11. Cross-block propagation — persistent `FilterXConfig` env

```filterx
filterx {
    $log.severity_text = "info";     # records $log.severity_text -> STRING
    $log.observed_time = 1700;       # records $log.observed_time -> INTEGER (per-key, no smear)
};

filterx {
    # resolves to knowledge carried over from the previous block via the persistent FilterXConfig env
    s = $log.severity_text;          # STRING
    t = $log.observed_time;          # INTEGER

    # can't resolve /  a field assigned different types in different blocks meets to UNKNOWN here
    x = $log.flaky;                  # UNKNOWN if an earlier block set it to a conflicting kind
};
```

## 12. Macros are never trackable — `expr-variable.c`

```filterx
filterx {
    # can't resolve /  hard macro-backed names always infer UNKNOWN (handle_is_macro)
    f = $FACILITY;     # UNKNOWN
    p = $PRI;          # UNKNOWN
    i = $ISODATE;      # UNKNOWN

    # resolves to contrast: ordinary message fields are trackable -- including the
    #    well-known names that merely *look* like macros
    $PROGRAM = "x";
    q = $PROGRAM;      # STRING
};
```

Only names carrying `LM_VF_MACRO` are excluded, i.e. the ones evaluated straight off the message
and rejected as assignment targets. `$PROGRAM`, `$HOST`, `$MESSAGE` and friends are pre-registered
*assignable* NV pairs rather than hard macros (`log_msg_is_handle_macro()` is false for them), so
they take part in inference like any other field.

### DIFF
--- a/lib/filterx/expr-arithmetic-operators.c
+++ b/lib/filterx/expr-arithmetic-operators.c
@@ -383,6 +383,20 @@ _optimize_plus(FilterXExpr *s)
   return NULL;
 }
 
+#if SYSLOG_NG_ENABLE_JIT
+static void
+_infer_types_plus(FilterXExpr *s, FilterXTypeEnv *env)
+{
+  filterx_expr_infer_types_default(s, env);
+  FilterXArithmeticOperator *self = (FilterXArithmeticOperator *) s;
+  FilterXStaticTypeSpec lhs_spec = self->super.lhs ? self->super.lhs->static_type : INITIAL_FILTERX_STATIC_TYPE_SPEC;
+  FilterXStaticTypeSpec rhs_spec = self->super.rhs ? self->super.rhs->static_type : INITIAL_FILTERX_STATIC_TYPE_SPEC;
+
+  s->static_type = filterx_static_type_spec_meet(lhs_spec, rhs_spec);
+}
+
+#endif
+
 static FilterXObject *
 _do_uminus(FilterXObject *operand_obj, FilterXExpr *expr)
 {
@@ -651,6 +665,7 @@ filterx_operator_plus_new(FilterXExpr *lhs, FilterXExpr *rhs)
   self->super.super.free_fn = _free_arithmetic_common;
 #if SYSLOG_NG_ENABLE_JIT
   self->super.super.compile = _compile_plus;
+  self->super.super.infer_types = _infer_types_plus;
 #endif
 
   return &self->super.super;
--- a/lib/filterx/expr-arithmetic-operators.c
+++ b/lib/filterx/expr-arithmetic-operators.c
@@ -384,6 +384,12 @@ _optimize_plus(FilterXExpr *s)
 }
 
 #if SYSLOG_NG_ENABLE_JIT
+static inline gboolean
+_is_numeric_static_type(FilterXStaticType kind)
+{
+  return kind == FILTERX_STATIC_TYPE_INTEGER || kind == FILTERX_STATIC_TYPE_DOUBLE;
+}
+
 static void
 _infer_types_plus(FilterXExpr *s, FilterXTypeEnv *env)
 {
@@ -391,6 +397,25 @@ _infer_types_plus(FilterXExpr *s, FilterXTypeEnv *env)
   FilterXArithmeticOperator *self = (FilterXArithmeticOperator *) s;
   FilterXStaticTypeSpec lhs_spec = self->super.lhs ? self->super.lhs->static_type : INITIAL_FILTERX_STATIC_TYPE_SPEC;
   FilterXStaticTypeSpec rhs_spec = self->super.rhs ? self->super.rhs->static_type : INITIAL_FILTERX_STATIC_TYPE_SPEC;
+  FilterXStaticType lhs_kind = filterx_static_type_kind(lhs_spec);
+  FilterXStaticType rhs_kind = filterx_static_type_kind(rhs_spec);
+
+  /* Numeric promotion: filterx_object_add() keeps int+int in the integer domain but widens
+   * to double as soon as either side is one, so a mixed pair is statically DOUBLE — a
+   * result a plain meet would throw away as UNKNOWN.
+   *
+   * Both kinds must be known numerics for this to hold. An UNKNOWN operand is not a
+   * "probably integer": it may be a double at runtime, which would widen the result, so the
+   * pair stays UNKNOWN. Claiming INTEGER on a merely-unknown operand would let an
+   * integer-speculating consumer treat a double as an int64. */
+  if (_is_numeric_static_type(lhs_kind) && _is_numeric_static_type(rhs_kind))
+    {
+      FilterXStaticType result = (lhs_kind == FILTERX_STATIC_TYPE_DOUBLE || rhs_kind == FILTERX_STATIC_TYPE_DOUBLE)
+                                 ? FILTERX_STATIC_TYPE_DOUBLE
+                                 : FILTERX_STATIC_TYPE_INTEGER;
+      s->static_type = filterx_static_type_kind_only(result);
+      return;
+    }
 
   s->static_type = filterx_static_type_spec_meet(lhs_spec, rhs_spec);
 }
--- a/lib/filterx/expr-condition.c
+++ b/lib/filterx/expr-condition.c
@@ -258,6 +258,30 @@ _compile_conditional(FilterXExpr *s, FilterXJIT *jit)
   return LLVMBuildLoad2(ir, ffi->ptr_ty, result_slot, "result");
 }
 
+static void
+_conditional_infer_types(FilterXExpr *s, FilterXTypeEnv *env)
+{
+  FilterXConditional *self = (FilterXConditional *) s;
+
+  /* The condition runs unconditionally before the branches. */
+  filterx_expr_infer_types(self->condition, env);
+
+  /* Walk each branch on its own clone of the env, then meet the resulting envs into the
+   * outer env. A missing branch is equivalent to a no-op, which still contributes the
+   * pre-branch env to the meet. */
+  FilterXTypeEnv *false_env = filterx_type_env_clone(env);
+
+  filterx_expr_infer_types(self->true_branch, env);
+  filterx_expr_infer_types(self->false_branch, false_env);
+
+  filterx_type_env_meet_into(env, false_env);
+  filterx_type_env_free(false_env);
+
+  FilterXStaticTypeSpec t_t = self->true_branch ? self->true_branch->static_type : INITIAL_FILTERX_STATIC_TYPE_SPEC;
+  FilterXStaticTypeSpec f_t = self->false_branch ? self->false_branch->static_type : INITIAL_FILTERX_STATIC_TYPE_SPEC;
+  s->static_type = filterx_static_type_spec_meet(t_t, f_t);
+}
+
 #endif
 
 FilterXExpr *
@@ -270,6 +294,7 @@ filterx_conditional_new(FilterXExpr *condition)
   self->super.walk_children = _conditional_walk;
   self->super.free_fn = _free;
 #if SYSLOG_NG_ENABLE_JIT
+  self->super.infer_types = _conditional_infer_types;
   self->super.compile = _compile_conditional;
 #endif
   self->super.suppress_from_trace = TRUE;
--- a/lib/filterx/expr-literal.c
+++ b/lib/filterx/expr-literal.c
@@ -108,6 +108,9 @@ _spec_from_filterx_object(FilterXObject *obj)
   if (filterx_object_is_type_or_ref(obj, &FILTERX_TYPE_NAME(integer)))
     return filterx_static_type_kind_only(FILTERX_STATIC_TYPE_INTEGER);
 
+  if (filterx_object_is_type_or_ref(obj, &FILTERX_TYPE_NAME(double)))
+    return filterx_static_type_kind_only(FILTERX_STATIC_TYPE_DOUBLE);
+
   FilterXStaticType outer_kind = FILTERX_STATIC_TYPE_UNKNOWN;
   if (filterx_object_is_type_or_ref(obj, &FILTERX_TYPE_NAME(dict)))
     outer_kind = FILTERX_STATIC_TYPE_DICT;
--- a/lib/filterx/expr-literal.c
+++ b/lib/filterx/expr-literal.c
@@ -67,35 +67,22 @@ _literal_walk(FilterXExpr *s, FilterXExprWalkFunc f, gpointer user_data)
 #include "filterx/jit/jit.h"
 #include "filterx/jit/ffi.h"
 
-/* Classify a literal FilterXObject's outer kind. Containers are recognised by kind only;
- * one level of element type is computed separately in _spec_from_filterx_object. */
-static FilterXStaticType
-_kind_of_filterx_object(FilterXObject *obj)
-{
-  if (!obj)
-    return FILTERX_STATIC_TYPE_UNKNOWN;
-  if (filterx_object_is_type_or_ref(obj, &FILTERX_TYPE_NAME(string)))
-    return FILTERX_STATIC_TYPE_STRING;
-  if (filterx_object_is_type_or_ref(obj, &FILTERX_TYPE_NAME(integer)))
-    return FILTERX_STATIC_TYPE_INTEGER;
-  if (filterx_object_is_type_or_ref(obj, &FILTERX_TYPE_NAME(dict)))
-    return FILTERX_STATIC_TYPE_DICT;
-  if (filterx_object_is_type_or_ref(obj, &FILTERX_TYPE_NAME(list)))
-    return FILTERX_STATIC_TYPE_LIST;
-  return FILTERX_STATIC_TYPE_UNKNOWN;
-}
-
+/* Recursively determine the full FilterXStaticTypeSpec for a literal FilterXObject.
+ * Walks dict / list elements to compute element types, bottoming out at STRING / leaf
+ * types. The recursion follows the literal's own (finite) nesting — there is no depth cap. */
 typedef struct
 {
   FilterXStaticTypeSpec spec;   /* the meet of all elements seen so far (valid once seen) */
   gboolean seen;                /* whether any element has been observed yet */
 } _LiteralElementMeetCtx;
 
+static FilterXStaticTypeSpec _spec_from_filterx_object(FilterXObject *obj);
+
 static gboolean
 _collect_value_spec(FilterXObject *key, FilterXObject *value, gpointer user_data)
 {
   _LiteralElementMeetCtx *ctx = (_LiteralElementMeetCtx *) user_data;
-  FilterXStaticTypeSpec value_spec = filterx_static_type_kind_only(_kind_of_filterx_object(value));
+  FilterXStaticTypeSpec value_spec = _spec_from_filterx_object(value);
   /* The `seen` flag distinguishes "no elements yet" from "elements meet to UNKNOWN":
    * the first element seeds the accumulator, subsequent ones actually meet. */
   if (!ctx->seen)
@@ -109,20 +96,37 @@ _collect_value_spec(FilterXObject *key, FilterXObject *value, gpointer user_data
   return ctx->spec != FILTERX_STATIC_TYPE_UNKNOWN_SPEC;
 }
 
-/* Determine the FilterXStaticTypeSpec for a literal FilterXObject: the outer kind plus,
- * for dict / list, one level of element type (the meet of the elements' kinds). */
 static FilterXStaticTypeSpec
 _spec_from_filterx_object(FilterXObject *obj)
 {
-  FilterXStaticType outer_kind = _kind_of_filterx_object(obj);
-  if (outer_kind != FILTERX_STATIC_TYPE_DICT && outer_kind != FILTERX_STATIC_TYPE_LIST)
-    return filterx_static_type_kind_only(outer_kind);
+  if (!obj)
+    return FILTERX_STATIC_TYPE_UNKNOWN_SPEC;
 
-  /* Empty container leaves the accumulator at UNKNOWN (the seed); a non-empty one carries the
-   * element meet. Either way ctx.spec is the element type. */
+  if (filterx_object_is_type_or_ref(obj, &FILTERX_TYPE_NAME(string)))
+    return filterx_static_type_kind_only(FILTERX_STATIC_TYPE_STRING);
+
+  if (filterx_object_is_type_or_ref(obj, &FILTERX_TYPE_NAME(integer)))
+    return filterx_static_type_kind_only(FILTERX_STATIC_TYPE_INTEGER);
+
+  FilterXStaticType outer_kind = FILTERX_STATIC_TYPE_UNKNOWN;
+  if (filterx_object_is_type_or_ref(obj, &FILTERX_TYPE_NAME(dict)))
+    outer_kind = FILTERX_STATIC_TYPE_DICT;
+  else if (filterx_object_is_type_or_ref(obj, &FILTERX_TYPE_NAME(list)))
+    outer_kind = FILTERX_STATIC_TYPE_LIST;
+  else
+    return FILTERX_STATIC_TYPE_UNKNOWN_SPEC;
+
+  /* Determine the meet of all element specs. The `seen` flag distinguishes "no elements
+   * observed yet" from "element type is UNKNOWN". An empty container ends the iter unseen
+   * and gets a FRESH element type — "element not yet committed" — so that the first write to
+   * it lifts the element to the written value's type (see
+   * filterx_type_env_update_on_write). FRESH is stripped before it reaches codegen. */
   _LiteralElementMeetCtx ctx = { .spec = FILTERX_STATIC_TYPE_UNKNOWN_SPEC, .seen = FALSE };
   filterx_object_iter(obj, _collect_value_spec, &ctx);
-  return filterx_static_type_make_container(outer_kind, ctx.spec);
+  FilterXStaticTypeSpec element_spec = ctx.seen
+                                       ? ctx.spec
+                                       : filterx_static_type_kind_only(FILTERX_STATIC_TYPE_FRESH);
+  return filterx_static_type_make_container(outer_kind, element_spec);
 }
 
 static void
--- a/lib/filterx/expr-literal.c
+++ b/lib/filterx/expr-literal.c
@@ -21,6 +21,10 @@
  */
 #include "filterx/expr-literal.h"
 #include "filterx/filterx-eval.h"
+#include "filterx/object-dict.h"
+#include "filterx/object-list.h"
+#include "filterx/object-string.h"
+#include "filterx/object-primitive.h"
 
 typedef struct _FilterXLiteral
 {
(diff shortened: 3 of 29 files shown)


sodomelle:
Build fails when `--disable-jit` flag is used during configure:
- `_infer_types_plus` is not used in this case, should be guarded by `SYSLOG_NG_ENABLE_JIT`
- struct contains `infer_types` when `SYSLOG_NG_ENABLE_JIT` is enabled, but referenced outside the `#if`-s

lib/filterx/filterx-type-inference.h
```diff
@@ -0,0 +1,109 @@
...
+
+typedef enum
+{
+  /* "no information" — the kind of the canonical unknown singleton spec
+   * (FILTERX_STATIC_TYPE_UNKNOWN_SPEC). Every enum value is a legal node kind. */
+  FILTERX_STATIC_TYPE_UNKNOWN = 0,
+  FILTERX_STATIC_TYPE_DICT    = 1,
+  FILTERX_STATIC_TYPE_LIST    = 2,
+  FILTERX_STATIC_TYPE_STRING  = 3,
+  FILTERX_STATIC_TYPE_INTEGER = 4,
```
sodomelle, line 36: Shouldn't we have `double` and `boolean` too?

lib/filterx/tests/test_type_inference.c
```diff
@@ -0,0 +1,268 @@
...
+}
+
+Test(filterx_type_inference, if_one_sided_assign_collapses_to_unknown)
+{
+  /* $u was unknown before; one branch sets STRING, other does nothing.
+   * Meet of STRING with UNKNOWN -> UNKNOWN. */
+  FilterXExpr *cond = filterx_floating_variable_expr_new("c");
+  FilterXExpr *then_assign = filterx_assign_new(
+                               filterx_floating_variable_expr_new("u"),
+                               filterx_literal_new(filterx_string_new("a", -1)));
```
sodomelle, line 217: "$u" is STRING here, no?

lib/filterx/tests/test_type_inference.c
```diff
@@ -0,0 +1,268 @@
...
+
+  cr_assert_eq(filterx_static_type_kind(d->static_type), FILTERX_STATIC_TYPE_DICT);
+  cr_assert_eq(filterx_static_type_kind(filterx_static_type_element(d->static_type)),
+               FILTERX_STATIC_TYPE_STRING);
+  filterx_expr_unref(d);
+}
+
+Test(filterx_type_inference, mixed_value_types_in_literal_dict_collapse_element)
+{
+  /* {"a": "x", "b": [1, 2]} — string and list values; element type meets to UNKNOWN. */
```
sodomelle: This seems like a one element list, not a 2 element one, as the comment above implies.

lib/filterx/tests/test_type_inference.c
```diff
@@ -0,0 +1,268 @@
...
+
+  cr_assert_eq(filterx_static_type_kind(d->static_type), FILTERX_STATIC_TYPE_DICT);
+  /* element type collapses because string and list don't meet. */
+  cr_assert_eq(filterx_static_type_kind(filterx_static_type_element(d->static_type)), FILTERX_STATIC_TYPE_UNKNOWN);
+  filterx_expr_unref(d);
+}
+
+Test(filterx_type_inference, get_subscript_shifts_operand_spec)
+{
+  /* d = {"a": "x", "b": "y"} → DICT_OF_STRING. d["a"] → STRING. */
```
sodomelle, line 391: Also missing `"b": "y"`.

lib/filterx/expr-arithmetic-operators.c
```diff
@@ -383,6 +383,23 @@ _optimize_plus(FilterXExpr *s)
...
+  filterx_expr_infer_types_default(s, env);
+  FilterXArithmeticOperator *self = (FilterXArithmeticOperator *) s;
+  FilterXStaticTypeSpec lhs_spec = self->super.lhs ? self->super.lhs->static_type : INITIAL_FILTERX_STATIC_TYPE_SPEC;
+  FilterXStaticTypeSpec rhs_spec = self->super.rhs ? self->super.rhs->static_type : INITIAL_FILTERX_STATIC_TYPE_SPEC;
+
+  /* When LHS is statically INTEGER, the result is INTEGER regardless of RHS: the JIT fast
+   * path (fx_jit_do_plus_int) handles non-integer RHS via fallback. This lets type info
+   * propagate through chains like `int_field + unknown_field + 34` without collapsing. */
+  if (filterx_static_type_kind(lhs_spec) == FILTERX_STATIC_TYPE_INTEGER)
+    s->static_type = filterx_static_type_kind_only(FILTERX_STATIC_TYPE_INTEGER);
```
sodomelle: Does this mean that:
after adding 5 + 3,14, tha value will be 8,14 but the inferred type will be INTEGER?

lib/filterx/expr-arithmetic-operators.c
```diff
@@ -651,6 +668,7 @@ filterx_operator_plus_new(FilterXExpr *lhs, FilterXExpr *rhs)
   self->super.super.free_fn = _free_arithmetic_common;
 #if SYSLOG_NG_ENABLE_JIT
   self->super.super.compile = _compile_plus;
+  self->super.super.infer_types = _infer_types_plus;
```
sodomelle, line 693: Why the other arithmetic operators have no infer_types set? I guess the example config didn't have those, but I feel like we should do it for every arithmetic operator we have.
Can be a follow up PR, no to bloat this one more.

lib/filterx/expr-condition.c
```diff
@@ -260,6 +260,30 @@ _compile_conditional(FilterXExpr *s, FilterXJIT *jit)
...
+
+  filterx_expr_infer_types(self->true_branch, env);
+  filterx_expr_infer_types(self->false_branch, false_env);
+
+  filterx_type_env_meet_into(env, false_env);
+  filterx_type_env_free(false_env);
+
+  FilterXStaticTypeSpec t_t = self->true_branch ? self->true_branch->static_type : INITIAL_FILTERX_STATIC_TYPE_SPEC;
+  FilterXStaticTypeSpec f_t = self->false_branch ? self->false_branch->static_type : INITIAL_FILTERX_STATIC_TYPE_SPEC;
+  s->static_type = filterx_static_type_spec_meet(t_t, f_t);
```
sodomelle, line 282: Why is else branch necessary for type_infer? Can't we just go with the true branch, if the false one is not present?

lib/filterx/expr-condition.c
```diff
@@ -269,6 +293,7 @@ filterx_conditional_new(FilterXExpr *condition)
   self->super.optimize = _optimize;
   self->super.walk_children = _conditional_walk;
   self->super.free_fn = _free;
+  self->super.infer_types = _conditional_infer_types;
```
sodomelle: Compiler error when JIT is disabled, all infer_types should be insed the `#if SYSLOG_NG_ENABLE_JIT`.

lib/filterx/expr-literal.c
```diff
@@ -58,6 +62,71 @@ _literal_walk(FilterXExpr *s, FilterXExprWalkFunc f, gpointer user_data)
...
+ * one level of element type is computed separately in _spec_from_filterx_object. */
+static FilterXStaticType
+_kind_of_filterx_object(FilterXObject *obj)
+{
+  if (!obj)
+    return FILTERX_STATIC_TYPE_UNKNOWN;
+  if (filterx_object_is_type_or_ref(obj, &FILTERX_TYPE_NAME(string)))
+    return FILTERX_STATIC_TYPE_STRING;
+  if (filterx_object_is_type_or_ref(obj, &FILTERX_TYPE_NAME(integer)))
+    return FILTERX_STATIC_TYPE_INTEGER;
```
sodomelle: I think we should have a double type too.

lib/filterx/filterx-type-inference.c
```diff
@@ -23,77 +23,185 @@
...
 filterx_type_env_free(FilterXTypeEnv *self)
 {
+  if (!self)
+    return;
+  g_hash_table_destroy(self->handle_to_spec);
   g_free(self);
 }
 
-/* Placeholder: variable type lookup is implemented by a later change. */
 FilterXStaticTypeSpec
```
sodomelle, line 183: For me the name implies it should return a FilterXTypeEnv (like the `filterx_type_env_clone` above).

lib/filterx/tests/test_type_inference.c
```diff
@@ -250,6 +276,242 @@ Test(filterx_type_inference, get_subscript_shifts_operand_spec)
...
+Test(filterx_type_inference, set_subscript_on_variable_pessimizes_element_type)
+{
+  /* d = {"a": "x"} → DICT_OF_STRING. d["b"] = 42; later read d → DICT (element UNKNOWN). */
+  GList *elems = g_list_append(NULL, filterx_literal_element_new(
+                                 filterx_literal_new(filterx_string_new("a", -1)),
+                                 filterx_literal_new(filterx_string_new("x", -1))));
+  FilterXExpr *assign = filterx_assign_new(
+                          filterx_floating_variable_expr_new("d"),
+                          filterx_literal_dict_new(elems));
+  FilterXExpr *set = filterx_set_subscript_new(
```
sodomelle, line 450: Optional: assert before extending the dict to show DICT_OF_STRING being changed to DICT, UNKNOWN?

lib/filterx/tests/test_type_inference.c
```diff
@@ -250,6 +276,242 @@ Test(filterx_type_inference, get_subscript_shifts_operand_spec)
...
+                         filterx_floating_variable_expr_new("d"),
+                         filterx_string_new("b", -1),
+                         filterx_literal_new(filterx_string_new("s", -1)));
+  FilterXExpr *read_a = filterx_getattr_new(filterx_floating_variable_expr_new("d"),
+                                            filterx_string_new("a", -1));
+
+  FilterXExpr *block = filterx_compound_expr_new_va(TRUE, assign_d, set_a, set_b, read_a, NULL);
+  block = _run(block);
+
+  cr_assert_eq(filterx_static_type_kind(read_a->static_type), FILTERX_STATIC_TYPE_DICT);
```
sodomelle, line 526: Maybe add an assert for the element, to see it being UNKNOWN?

lib/filterx/tests/test_type_inference.c
```diff
@@ -0,0 +1,268 @@
...
+#include "filterx/expr-condition.h"
+#include "filterx/expr-plus-assign.h"
+#include "filterx/object-string.h"
+#include "filterx/object-dict.h"
+#include "filterx/object-list.h"
+#include "filterx/expr-get-subscript.h"
+#include "filterx/expr-getattr.h"
+#include "filterx/expr-set-subscript.h"
+#include "filterx/expr-setattr.h"
+
```
sodomelle, line 43: needs a `#include "filterx/expr-arithmetic-operators.h"`, otherwise it doesn't compile.

lib/filterx/tests/test_type_inference.c
```diff
@@ -250,6 +276,242 @@ Test(filterx_type_inference, get_subscript_shifts_operand_spec)
...
+  FilterXExpr *read_y = filterx_getattr_new(filterx_floating_variable_expr_new("d"),
+                                            filterx_string_new("y", -1));
+
+  FilterXExpr *block = filterx_compound_expr_new_va(TRUE, assign_d, set_x, set_y, read_y, NULL);
+  block = _run(block);
+
+  cr_assert_eq(filterx_static_type_kind(read_y->static_type), FILTERX_STATIC_TYPE_DICT);
+  filterx_expr_unref(block);
+}
+
```
sodomelle, line 556: Claude's comment:
filterx_type_env_set never prunes attr_to_spec entries for the handle. So d = {}; d.a = {} records handle_d:a → DICT; a later d = "s" overwrites only the
  handle spec, and a subsequent d.a hits the per-key lookup first (_getattr_infer_types) and returns the stale DICT.
Unittest to validate:
```
Test(filterx_type_inference, reassigning_root_invalidates_stale_per_key_type)
{
  /* d = {}; d.a = {};   -> per-key records handle_d:a = DICT
    * d = "s";            -> whole variable overwritten; d is now a STRING
    * d.a                 -> must NOT still resolve to DICT. Reading an attr off a
    *                        string is statically UNKNOWN; the earlier per-key entry is stale. */
  FilterXExpr *assign_dict = filterx_assign_new(
                                filterx_floating_variable_expr_new("d"),
                                filterx_literal_dict_new(NULL));
  FilterXExpr *set_a = filterx_setattr_new(
                          filterx_floating_variable_expr_new("d"),
                          filterx_string_new("a", -1),
                          filterx_literal_dict_new(NULL));
  FilterXExpr *reassign_str = filterx_assign_new(
                                filterx_floating_variable_expr_new("d"),
                                filterx_literal_new(filterx_string_new("s", -1)));
  FilterXExpr *read_a = filterx_getattr_new(filterx_floating_variable_expr_new("d"),
                                            filterx_string_new("a", -1));

  FilterXExpr *block = filterx_compound_expr_new_va(TRUE, assign_dict, set_a, reassign_str, read_a, NULL);
  block = _run(block);

  cr_assert_eq(filterx_static_type_kind(read_a->static_type), FILTERX_STATIC_TYPE_UNKNOWN);
  filterx_expr_unref(block);
}
```

lib/filterx/jit/Makefile.am
```diff
@@ -26,7 +26,7 @@ filterx_bitcode_compile = \
 	$(DEFS) $(DEFAULT_INCLUDES) \
 	$(AM_CPPFLAGS) $(CPPFLAGS) \
 	$(filter-out -fsanitize=%,$(AM_CFLAGS) $(CFLAGS)) \
-	-emit-llvm -O3 --no-warnings
+	-emit-llvm -O3 --no-warnings -mno-outline-atomics
```
sodomelle, line 29: Shouldn't this flag be added to the CMake too?

sodomelle (commented):
Respect for the amount of work, time and effort being put into this PR. I noted a few things, giving a short summery here:
Missing floating point type could cause strange things (like 6,7 being an INTEGER).
I feel like if we started doing the arithmetic operators, we shouldn't stop with the addition, but it could be a followup PR, as this is already quite big.
I have mixed feelings about the number of comments present, as it is usually not the way of AxoSyslog.
Found a few build breaking things yesterday, note above.
Claude's unit test suggestions (one already mentioned in the new file):
- Switch — no switch test exists at all. A filterx_switch_new(selector, body) where one case writes a var, then read the var after the switch and assert
  UNKNOWN (body may not run) — covers _switch_infer_types and its meet-with-before.
  - Macro → UNKNOWN — build a macro-backed variable, assert its static_type stays UNKNOWN and that filterx_variable_expr_get_handle returns FALSE so it's never
  tracked. (I'm not certain of the exact libtest constructor for a macro var — worth confirming against how the harness makes $-macro variables.)
  - Cross-block persistent env — the current _run uses filterx_expr_infer_types_root (no persistence). A test needs its own driver calling
  filterx_expr_infer_types_root_persistent(block, shared_env) on block 1 (writing a type) then block 2 (reading it), asserting the type carried across — the only
  coverage of propagate_to_persistent and the clone-seed path used from filterx-pipe.c.

fernya:
> Build fails when `--disable-jit` flag is used during configure:
> 
> * `_infer_types_plus` is not used in this case, should be guarded by `SYSLOG_NG_ENABLE_JIT`
> * struct contains `infer_types` when `SYSLOG_NG_ENABLE_JIT` is enabled, but referenced outside the `#if`-s

Fixed, will be rebased soon.

lib/filterx/filterx-type-inference.h
fernya: As you mentioned, that'll be done in a separate PR.

lib/filterx/tests/test_type_inference.c
fernya: `Meet of STRING with UNKNOWN -> UNKNOWN.`

Yes, this arm of the if is STRING, the other is UNKNOWN, thus after the end of the if we consider it unknown (pessimistic behavior).

lib/filterx/tests/test_type_inference.c
fernya: You're right, I modified the comment as the code shows the appropriate behavior.

lib/filterx/tests/test_type_inference.c
fernya: Here the code is aligned to the comment.

lib/filterx/expr-arithmetic-operators.c
fernya: Yup, you're right. The implemented int fast-path handler will handle non-integer rhs the case runtime.

lib/filterx/expr-condition.c
fernya: Been there, fixed that, thanks!

lib/filterx/expr-literal.c
fernya: Will be done in another PR.

lib/filterx/filterx-type-inference.c
fernya: Bad naming, will rename it!

lib/filterx/expr-arithmetic-operators.c
fernya: I focused only on the test configuration, will be extended in later PRs.

lib/filterx/jit/Makefile.am
fernya: Nice catch!

lib/filterx/tests/test_type_inference.c
fernya: Did you find this with CMake?

sodomelle: Autotools.

lib/filterx/tests/test_type_inference.c
sodomelle: Yes, but the comments says "$u was UNKNOWN before", but it wasn't, it was STRING, no?

fernya: It wasn't defined (thus UNKNOWN) before the if (the first assignment was in `then_assign` which is the part of the conditional block). It was set to STRING in the true branch and left UNKNOWN in the else branch.

lib/filterx/tests/test_type_inference.c
fernya: Added.

lib/filterx/tests/test_type_inference.c
fernya: Nice catch, it is now handled.

lib/filterx/tests/test_type_inference.c
fernya: My build environment let undefined functions defined slip through, so it I added the include for correctness.

lib/filterx/tests/test_type_inference.c
fernya: Added.

lib/filterx/expr-arithmetic-operators.c
fernya: I implemented the proper behavior in 5240c7d and fb2149c

lib/filterx/filterx-type-inference.c
sodomelle: I see the old name here. But i did a lot of reviews recently so I might be hallucinating too.

sodomelle:
Build fails when --disable-jit flag is used during configure:
infer_types is set when SYSLOG_NG_ENABLE_JIT is enabled, using the defined `*infer_types` functions, but if JIT is disabled, those functions are unused resulting in build failure.

- `_infer_types_plus`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
- `_assign_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
- `_nullv_assign_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
- `_compound_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
- `_conditional_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
- `_get_subscript_infer_types`: create a new `#if SYSLOG_NG_ENABLE_JIT` block
- `_getattr_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
- `_literal_dict_infer_types`: create a new `#if SYSLOG_NG_ENABLE_JIT` block
- `_literal_list_infer_types`: put it next to `_literal_dict_infer_types`
- `_meet_element_specs`: put it next to `_literal_dict_infer_types` and `_literal_list_infer_types`
- `_literal_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
- `_plus_assign_infer_types`: create a new `#if SYSLOG_NG_ENABLE_JIT` block
- `_set_subscript_infer_types`: create a new `#if SYSLOG_NG_ENABLE_JIT` block
- `_setattr_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
- `_switch_infer_types`: create a new `#if SYSLOG_NG_ENABLE_JIT` block
- `_variable_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block

fernya:
> Build fails when --disable-jit flag is used during configure: infer_types is set when SYSLOG_NG_ENABLE_JIT is enabled, using the defined `*infer_types` functions, but if JIT is disabled, those functions are unused resulting in build failure.
> 
> * `_infer_types_plus`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
> * `_assign_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
> * `_nullv_assign_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
> * `_compound_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
> * `_conditional_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
> * `_get_subscript_infer_types`: create a new `#if SYSLOG_NG_ENABLE_JIT` block
> * `_getattr_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
> * `_literal_dict_infer_types`: create a new `#if SYSLOG_NG_ENABLE_JIT` block
> * `_literal_list_infer_types`: put it next to `_literal_dict_infer_types`
> * `_meet_element_specs`: put it next to `_literal_dict_infer_types` and `_literal_list_infer_types`
> * `_literal_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
> * `_plus_assign_infer_types`: create a new `#if SYSLOG_NG_ENABLE_JIT` block
> * `_set_subscript_infer_types`: create a new `#if SYSLOG_NG_ENABLE_JIT` block
> * `_setattr_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block
> * `_switch_infer_types`: create a new `#if SYSLOG_NG_ENABLE_JIT` block
> * `_variable_infer_types`: move into already existing `#if SYSLOG_NG_ENABLE_JIT` block

This should be rectified now.

lib/filterx/expr-condition.c
fernya: It is not. If we don't have else branch, then call to `filterx_expr_infer_types` is noop, and `filterx_type_env_meet_into` won't change the inferred type set.

lib/filterx/filterx-type-inference.c
fernya: It is renamed now.

lib/filterx/expr-condition.c
sodomelle: i don't get this tc then: `if_one_sided_assign_collapses_to_unknown`

fernya:
Abandoned for #1192

PR #1135, opened 2026-06-22
light: use unique loggen client ports in increased-send window tests

mitzkia:
The test_static_window_increased_send and test_dynamic_window_increased_send tests open a new loggen connection in each loop iteration and expect one distinct Prometheus time series per connection.

- Without a fixed client port the kernel assigns a random ephemeral source port, which does not guarantee it is unique across the iterations.
- The per-connection window metrics are labeled with connection="<ip:port>" (the peer address). When two connections happen to reuse the same port their series collapse into one, so fewer series appear than the test expects, causing intermittent assertion failures.
- Passing client_port=port_allocator() pins a distinct, free source port for each iteration, guaranteeing unique connection labels and stable metrics.

This PR fixes such rare failures:
```
FAILED functional_tests/source_drivers/network_source/test_static_and_dynamic_window.py::test_dynamic_window_increased_send - AssertionError: Metric syslogng_input_window_available expected: [0, 0, 0, 0, 0, 0, 0, 0, 0], actual: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
```

### DIFF
--- a/tests/light/functional_tests/source_drivers/network_source/test_static_and_dynamic_window.py
+++ b/tests/light/functional_tests/source_drivers/network_source/test_static_and_dynamic_window.py
@@ -56,7 +56,7 @@ def set_config_for_dynamic_window(config, port_allocator, dynamic_window_size, l
     return config, network_source, network_destination
 
 
-def send_messages_with_loggen(loggen, network_source, active_connections, message_counter):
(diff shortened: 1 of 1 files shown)


alltilla (approved):

alltilla:
I'm relieved we started to use this somewhere :D It was dead code for a bit of time.

MrAnno (approved):

PR #1139, opened 2026-06-23
[FilterX-jit] [Devirtualization] Implement devirtualization for some expressions

fernya:
The devirtualized IR is only emitted, when the static inference marked the operands with appropriate type, falls back to FilterXObject vtable lookups otherwise.

## Intro via examples

Legend: **[YES]** devirtualized helper, **[NO]** stays on the generic path.

### 1. Subscript reads -- `dict[key]`, `list[index]`

```
declare d = { "host": "h1", "port": 514 };
declare l = [ 10, 20, 30 ];

x = d["host"];   # [YES] dict[string-literal] -> fx_jit_do_get_subscript_dict_string_key (skips the hashable check)
k = "host";
x = d[k];        # [YES] key is statically string -> same string-key fast path
x = l[0];        # [YES] list[index] -> fx_jit_do_get_subscript_list
```

### 2. Subscript assignment -- `=` and `=??`

```
declare d = {};
declare l = [ 0 ];

d["a"] = 1;       # [YES] fx_jit_do_set_subscript_dict
l[0]   = 9;       # [YES] fx_jit_do_set_subscript_list
d["a"] =?? 2;     # [YES] null-coalescing -> fx_jit_do_nullv_set_subscript_dict
l[0]   =?? 9;     # [YES] null-coalescing -> fx_jit_do_nullv_set_subscript_list
```

### 3. Attribute access -- `dict.attr`

```
declare d = { "name": "" };

y = d.name;       # [YES] getattr on dict -> fx_jit_do_getattr_dict (dict.attr == dict[attr], known-string key)
d.name = "x";     # [YES] setattr on dict -> fx_jit_do_setattr_dict
d.name =?? "x";   # [YES] null-coalescing -> fx_jit_do_nullv_setattr_dict
```

### 4. String concatenation -- `str + str`

```
declare a = "foo";
declare b = "bar";

s = a + b;        # [YES] both operands statically string -> fx_jit_do_plus_string
s = a + "-baz";   # [YES] string + string-literal
```

### 5. Integer addition -- `int + ...`

```
declare n = int($num);

t = n + 1;            # [YES] lhs is integer -> fx_jit_do_plus_int
t = int($a) + $b;     # [YES] lhs integer => result is integer regardless of rhs (rhs type checked at runtime)
t = 1 + n;            # [YES] integer-literal lhs
```

---

### Counterexamples -- these stay on the generic path

```
# --- type not statically known ---
js = json($MSG);
v = js["k"];          # [NO] json() result type is unknown -> generic get_subscript
v = $MSG[0];          # [NO] message variable, no inferred type

# --- list append form (NULL key) ---
declare l = [];
l[] = 5;              # [NO] append has no key expression -> interpreter path (v1 limitation)

# --- string concat needs BOTH sides statically string ---
s = "prefix-" + $MSG; # [NO] rhs is a message value (unknown); meet(STRING, UNKNOWN) -> UNKNOWN -> generic plus

# --- integer plus: operand ORDER matters ---
t = $num + 1;         # [NO] lhs is unknown (message value) -> not the integer fast path
                      #      compare: `1 + $num` [YES] -- an integer lhs forces the result to integer

# --- only `+` is specialized among the arithmetic operators ---
declare n = int($x);
t = n - 1;            # [NO] subtraction is not devirtualized
t = n * 2;            # [NO] multiplication is not devirtualized
```

### DIFF
--- a/lib/filterx/expr-arithmetic-operators.c
+++ b/lib/filterx/expr-arithmetic-operators.c
@@ -20,6 +20,8 @@
  *
  */
 #include "expr-arithmetic-operators.h"
+#include "expr-arithmetic-operators-private.h"
+#include "expr-arithmetic-operators-devirt.h"
 #include "filterx/object-primitive.h"
 #include "filterx/expr-literal.h"
 #include "filterx/object-extractor.h"
@@ -30,14 +32,6 @@
 /* If you want to make the arithmetic operators support other types,
  * follow the pattern of filterx_operator_plus_new() which uses
  * filterx_object_add() for type-generic addition. */
-
-typedef struct FilterXArithmeticOperator_
-{
-  FilterXBinaryOp super;
-  FilterXObject *literal_lhs;
-  FilterXObject *literal_rhs;
-} FilterXArithmeticOperator;
-
 
 /* consumes operand objects */
 static gboolean
@@ -501,7 +495,7 @@ fx_jit_arithmetic_uminus(FilterXObject *operand, FilterXExpr *expr)
   return _do_uminus(operand, expr);
 }
 
-static FilterXIRValue
+FilterXIRValue
 _compile_binary_arithmetic(FilterXExpr *s, FilterXJIT *jit, const gchar *fn_name)
 {
   FilterXArithmeticOperator *self = (FilterXArithmeticOperator *) s;
@@ -557,12 +551,6 @@ _compile_binary_arithmetic(FilterXExpr *s, FilterXJIT *jit, const gchar *fn_name
   filterx_jit_ir_add_sequence_to_block(jit, finish, block);
   filterx_jit_ir_set_insert_point_to_sequence_tail(jit, finish);
   return LLVMBuildLoad2(ir, ffi->ptr_ty, result_slot, "result");
-}
-
-static FilterXIRValue
-_compile_plus(FilterXExpr *s, FilterXJIT *jit)
-{
-  return _compile_binary_arithmetic(s, jit, "fx_jit_arithmetic_plus");
 }
 
 static FilterXIRValue
--- a/lib/filterx/expr-arithmetic-operators.c
+++ b/lib/filterx/expr-arithmetic-operators.c
@@ -500,57 +500,27 @@ _compile_binary_arithmetic(FilterXExpr *s, FilterXJIT *jit, const gchar *fn_name
 {
   FilterXArithmeticOperator *self = (FilterXArithmeticOperator *) s;
   FilterXJITFFI *ffi = filterx_jit_get_ffi(jit);
-  FilterXIRBuilder ir = filterx_jit_get_ir_builder(jit);
-  FilterXIRValue block = filterx_jit_ir_get_current_block(jit);
 
-  FilterXIRValue result_slot = LLVMBuildAlloca(ir, ffi->ptr_ty, "result");
-  LLVMBuildStore(ir, LLVMConstNull(ffi->ptr_ty), result_slot);
-
-  FilterXIRSequence lhs_null = filterx_jit_ir_create_sequence(jit, "arith_lhs_null", block);
-  FilterXIRSequence eval_rhs = filterx_jit_ir_create_sequence(jit, "arith_eval_rhs", block);
-  FilterXIRSequence rhs_null = filterx_jit_ir_create_sequence(jit, "arith_rhs_null", block);
-  FilterXIRSequence do_op = filterx_jit_ir_create_sequence(jit, "arith_do_op", block);
-  FilterXIRSequence finish = filterx_jit_ir_create_sequence(jit, "arith_finish", block);
+  /* mirrors _eval_op(): the rhs is not evaluated when the lhs fails */
+  FilterXIRShortCircuit short_circuit;
+  fx_jit_emit_short_circuit_begin(jit, &short_circuit, "arith");
 
   FilterXIRValue lhs = self->literal_lhs
                        ? fx_jit_emit_object_ref(jit, fx_jit_emit_const_ptr(jit, self->literal_lhs))
                        : filterx_expr_compile_or_eval_typed(self->super.lhs, jit);
+  fx_jit_emit_bail_if_null(jit, &short_circuit, lhs, NULL);
 
-  /* if (!lhs) goto finish; */
-  FilterXIRValue lhs_is_null = LLVMBuildIsNull(ir, lhs, "lhs_is_null");
-  LLVMBuildCondBr(ir, lhs_is_null, lhs_null, eval_rhs);
-
-  filterx_jit_ir_add_sequence_to_block(jit, lhs_null, block);
-  filterx_jit_ir_set_insert_point_to_sequence_tail(jit, lhs_null);
-  LLVMBuildBr(ir, finish);
-
-  /* eval_rhs */
-  filterx_jit_ir_add_sequence_to_block(jit, eval_rhs, block);
-  filterx_jit_ir_set_insert_point_to_sequence_tail(jit, eval_rhs);
   FilterXIRValue rhs = self->literal_rhs
                        ? fx_jit_emit_object_ref(jit, fx_jit_emit_const_ptr(jit, self->literal_rhs))
                        : filterx_expr_compile_or_eval(self->super.rhs, jit);
+  fx_jit_emit_bail_if_null(jit, &short_circuit, rhs, lhs);
 
-  /* if (!rhs) { unref(lhs); goto finish; } */
-  FilterXIRValue rhs_is_null = LLVMBuildIsNull(ir, rhs, "rhs_is_null");
-  LLVMBuildCondBr(ir, rhs_is_null, rhs_null, do_op);
-
-  filterx_jit_ir_add_sequence_to_block(jit, rhs_null, block);
-  filterx_jit_ir_set_insert_point_to_sequence_tail(jit, rhs_null);
-  fx_jit_emit_object_unref(jit, lhs);
-  LLVMBuildBr(ir, finish);
-
-  /* do_op: both operands are non-NULL; the called function consumes them */
-  filterx_jit_ir_add_sequence_to_block(jit, do_op, block);
-  filterx_jit_ir_set_insert_point_to_sequence_tail(jit, do_op);
+  /* both operands are non-NULL, the called function consumes them */
   FilterXIRValue args[] = { lhs, rhs, fx_jit_emit_const_ptr(jit, self) };
   FilterXIRType param_tys[] = { ffi->ptr_ty, ffi->ptr_ty, ffi->ptr_ty };
-  LLVMBuildStore(ir, fx_jit_emit_extern_call(jit, fn_name, ffi->ptr_ty, param_tys, args, 3), result_slot);
-  LLVMBuildBr(ir, finish);
+  FilterXIRValue result = fx_jit_emit_extern_call(jit, fn_name, ffi->ptr_ty, param_tys, args, 3);
 
-  filterx_jit_ir_add_sequence_to_block(jit, finish, block);
-  filterx_jit_ir_set_insert_point_to_sequence_tail(jit, finish);
-  return LLVMBuildLoad2(ir, ffi->ptr_ty, result_slot, "result");
+  return fx_jit_emit_short_circuit_end(jit, &short_circuit, result);
 }
 
 static FilterXIRValue
--- a/lib/filterx/expr-get-subscript.c
+++ b/lib/filterx/expr-get-subscript.c
@@ -21,6 +21,8 @@
  */
 #include "filterx/expr-get-subscript.h"
 #include "filterx/filterx-eval.h"
+#include "filterx/object-dict.h"
+#include "filterx/object-list.h"
 #include "stats/stats-registry.h"
 #include "stats/stats-cluster-single.h"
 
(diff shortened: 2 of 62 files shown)


lib/filterx/expr-get-subscript.c
```diff
@@ -21,6 +21,8 @@
  */
 #include "filterx/expr-get-subscript.h"
 #include "filterx/filterx-eval.h"
+#include "filterx/object-dict.h"
```
sodomelle: Also can be inside `#if SYSLOG_NG_ENABLE_JIT`, as nothing else uses them.

lib/filterx/expr-set-subscript.c
```diff
@@ -26,6 +26,8 @@
 #include "filterx/object-null.h"
 #include "filterx/object-message-value.h"
 #include "filterx/object-extractor.h"
+#include "filterx/object-dict.h"
```
sodomelle: Same as above.

lib/filterx/expr-set-subscript.c
```diff
@@ -235,8 +235,47 @@ fx_jit_do_set_subscript_list(FilterXObject *object, FilterXObject *key, FilterXO
   return _do_set_subscript(object, key, new_value, expr, filterx_list_set_subscript_via_key);
 }
 
+typedef FilterXObject *(*FXSetSubscriptWrapper)(FilterXObject *object, FilterXObject *key,
```
sodomelle: should it be also an if JIT enabled macro?

lib/filterx/expr-getattr.c
```diff
@@ -193,12 +194,40 @@ fx_jit_do_getattr(FilterXObject *variable, FilterXObject *attr, FilterXExpr *exp
...
+    {
+      filterx_eval_push_error_static_info("Failed to get-attribute from object", expr,
+                                          "Failed to evaluate expression");
+      return NULL;
+    }
+
+  FilterXObject *result = filterx_dict_get_subscript_unchecked(variable, attr);
+  if (!result)
+    filterx_eval_push_error_static_info("Failed to get-attribute from object", expr,
+                                        "Failed to evaluate key");
```
sodomelle: This error shouldn't be about a key.

lib/filterx/expr-plus.c
```diff
@@ -125,6 +125,47 @@ fx_jit_do_plus(FilterXObject *lhs, FilterXObject *rhs, FilterXExpr *expr)
...
+      filterx_eval_push_error_static_info("Failed to add values", expr, "Failed to evaluate right hand side");
+      goto exit;
+    }
+
+  gint64 rhs_val;
+  if (G_LIKELY(filterx_integer_unwrap(rhs, &rhs_val)))
+    {
+      gint64 lhs_val;
+      gboolean ok G_GNUC_UNUSED = filterx_integer_unwrap(lhs, &lhs_val);
+      g_assert(ok);
```
sodomelle: I feel like, this can cause an issue, based on the previous PR, that we can have floating point values marked as integer at this point (after some specific additions).

sodomelle (commented):
Does this need new unittests?

lib/filterx/expr-set-subscript.c
fernya: Done.

lib/filterx/expr-get-subscript.c
fernya: Done.

lib/filterx/expr-set-subscript.c
fernya: This is already in a JIT enabled macro.

lib/filterx/expr-plus.c
fernya: This was changed to include all numeric types properly. However in this version on line 149 it already tries to unwrap the integer and it only proceeds into this branch if that was successful.

lib/filterx/object-dict.h
```diff
@@ -37,4 +37,9 @@ FilterXObject *filterx_dict_new_from_args(FilterXExpr *s, FilterXObject *args[],
 FilterXDictAnchor filterx_dict_get_anchor_for_key(FilterXObject *s, FilterXObject *key);
 void filterx_dict_set_subscript_by_anchor(FilterXObject *s, FilterXDictAnchor anchor, FilterXObject **new_value);
 
+/* Devirtualized fast path for dict subscript reads. Unwraps a FilterXRef wrapper and
+ * dispatches directly to the dict's get_subscript implementation, skipping the FilterXObject
+ * vtable indirection. */
```
MrAnno: These kinds of comments are useless, they just mirror the clean and clear implementation (which can change with time, leaving the comment deprecated and confusing).

In general: if we need to add new direct methods for the Object layer only because of devirtualization, we could just place those methods under a `/* for devirtualization */` or similar comment to let the everyone know this is kind of breaking clean OOP architecture for a reason and should not be used anywhere else.

lib/filterx/expr-get-subscript.c
```diff
@@ -177,6 +181,80 @@ filterx_get_subscript_get_operand(FilterXExpr *s)
...
+  switch (filterx_static_type_kind(self->operand->static_type))
+    {
+    case FILTERX_STATIC_TYPE_DICT:
+      fn_name = "fx_jit_do_get_subscript_dict";
+      break;
+    case FILTERX_STATIC_TYPE_LIST:
+      fn_name = "fx_jit_do_get_subscript_list";
+      break;
+    default:
+      return fx_jit_emit_expr_eval(jit, s);
```
MrAnno: I think we should provide a generic compiled version of get_subscript, not secretly falling back to the interpreter.

lib/filterx/expr-get-subscript.c
```diff
@@ -177,6 +181,80 @@ filterx_get_subscript_get_operand(FilterXExpr *s)
...
+#if SYSLOG_NG_ENABLE_JIT
+
+#include "filterx/jit/jit.h"
+#include "filterx/jit/ffi.h"
+
+typedef FilterXObject *(*FXGetSubscriptHelper)(FilterXObject *object, FilterXObject *key);
+
+static inline __attribute__((always_inline)) FilterXObject *
+_do_get_subscript(FilterXObject *variable, FilterXObject *key, FilterXExpr *expr,
+                  FXGetSubscriptHelper helper)
```
MrAnno: Naming something as `Helper` is usually a smell and hint we can't find  a meaningful name -> worth reconsidering the design.

In this specific case, I think we can name these things without refactoring, for example:

- FilterXJITTypedGetSubscript, typed_get_subscript
- fx_jit_typed_get_subscript_dict, fx_jit_typed_get_subscript_list 

We could follow the same for the other expressions as well.

lib/filterx/object-dict.c
```diff
@@ -851,6 +851,22 @@ filterx_dict_get_subscript(FilterXObject *s, FilterXObject *key)
   return _filterx_dict_get_subscript(filterx_ref_unwrap_ro(s), key);
 }
 
+FilterXObject *
+filterx_dict_get_subscript_unchecked(FilterXObject *s, FilterXObject *key)
+{
+  /* The caller is responsible for ensuring @key is hashable (i.e. a string), letting us
+   * skip the normalize_key vtable check that the generic path runs. */
```
MrAnno: Does this worth making the compromise adding such an unchecked method?

filterx_mapping_normalize_key() has no virtual methods/anything indirect, it just checks the is_hashable bitfield (it should fit into the cache, it's directly inside the object), + a null check.

lib/filterx/expr-set-subscript.c
```diff
@@ -170,6 +233,40 @@ _set_subscript_infer_types(FilterXExpr *s, FilterXTypeEnv *env)
...
+  switch (filterx_static_type_kind(self->object->static_type))
+    {
+    case FILTERX_STATIC_TYPE_DICT:
+      fn_name = "fx_jit_do_set_subscript_dict";
+      break;
+    case FILTERX_STATIC_TYPE_LIST:
+      fn_name = "fx_jit_do_set_subscript_list";
+      break;
+    default:
+      return fx_jit_emit_expr_eval(jit, s);
```
MrAnno: Same.

lib/filterx/expr-setattr.c
```diff
@@ -217,13 +272,21 @@ _emit_setattr_call(FilterXSetAttr *self, FilterXJIT *jit, const gchar *fn_name)
 static FilterXIRValue
 _setattr_compile(FilterXExpr *s, FilterXJIT *jit)
 {
```
MrAnno, line 178: Opinion: I prefer this version when doing devirtualization.

A single _compile virt method implementation, and the static-type-based implementation is on the top with early returns.

lib/filterx/expr-string-operators.c
```diff
@@ -238,3 +238,47 @@ filterx_string_slicing_new(FilterXExpr *lhs, FilterXExpr *start, FilterXExpr *en
...
+    filterx_eval_push_error_static_info("Failed to add values", "add() method failed");
+
+exit:
+  filterx_object_unref(lhs);
+  filterx_object_unref(rhs);
+  return result;
+}
+
+FilterXIRValue
+filterx_string_concat_compile(FilterXJIT *jit, FilterXIRValue lhs, FilterXIRValue rhs, FilterXExpr *expr)
```
MrAnno: There is a problem Klaudia probably tried to mask without telling us the problem:

This method belongs to to expr-arithmetic-operators if we follow the pattern we started doing.
The problem is that each expression containing N different devirtualized implementations based on static type doesn't scale well, it adds a lot of boilerplate and noise to the given expression.

For strings, she found a completely separate unit, but I believe this is not the right place. We should instead figure out where to put these implementations. The options I can think of:

- for each type, a separate file which contains all devirtualized expessions
- for each expression, a separate file which contains implementations for all types

I'd vote for the second one because the similar methods should be close to each other because they usually have to be modified together.

For now, I think it's good-enough if we just place these methods inside the expr.c they belong to.

lib/filterx/filterx-ref.h
```diff
@@ -172,4 +172,12 @@ filterx_ref_ground(FilterXObject *s)
 FilterXObject *filterx_ref_dup(FilterXObject *s);
 FilterXObject *_filterx_ref_new(FilterXObject *value);
 
+/* Given a child @s just read out of @c's underlying container, replace a shared child
+ * xref with a floating one parented to @c so that a later mutation triggers copy-on-write
+ * up the chain. This is the COW-critical step of the ref's getattr/get_subscript vtable;
+ * the JIT-devirtualized dict/list read fast paths must call it after a raw (ref-bypassing)
+ * lookup to preserve copy-on-write. @c must be a FilterXRef; @s may be any object (non-ref
+ * children are returned unchanged). */
+FilterXObject *filterx_ref_float_shared_child(FilterXObject *s, FilterXObject *c);
```
MrAnno: I don't understand the comment, but I think it's not right to mention JIT-devirtualization here, we should explain clearly why it is safe to publish and how to use this method.

lib/filterx/expr-get-subscript.c
```diff
@@ -208,6 +208,10 @@ _do_get_subscript(FilterXObject *variable, FilterXObject *key, FilterXExpr *expr
   FilterXObject *result = helper(variable, key);
   if (!result)
     filterx_eval_push_error("Failed to get-subscript from object", key);
+  else if (filterx_object_is_ref(variable))
+    /* The dict/list helpers unwrap @variable read-only, so float the shared child against
+     * it (as the ref's get_subscript vtable does) to preserve copy-on-write on later writes. */
+    result = filterx_ref_float_shared_child(result, variable);
```
MrAnno: This looks like a bugfix commit, but it doesn't state it clearly.
I don't understand unfortunately, a better commit message woul help (and we could probably create a function handling this branch, and the name would explain what it does instead of a comment.

lib/filterx/expr-getattr.c
```diff
@@ -228,6 +233,8 @@ fx_jit_do_getattr_dict(FilterXObject *variable, FilterXObject *attr, FilterXExpr
   if (!result)
     filterx_eval_push_error_static_info("Failed to get-attribute from object",
                                         "Failed to evaluate key");
+  else if (filterx_object_is_ref(variable))
+    result = filterx_ref_float_shared_child(result, variable);
```
MrAnno: Yep, this is definitely a bugfix, and it belongs to the type-optimized version, it doesn't affect the generic getattr/get_subscript implementation.

lib/filterx/expr-arithmetic-operators.c
```diff
@@ -472,6 +512,123 @@ fx_jit_arithmetic_plus(FilterXObject *lhs, FilterXObject *rhs, FilterXExpr *expr
...
+      filterx_eval_push_error_static_info("Failed to add values", "Failed to evaluate right hand side");
+      goto exit;
+    }
+
+  gint64 rhs_val;
+  if (G_LIKELY(filterx_integer_unwrap(rhs, &rhs_val)))
+    {
+      gint64 lhs_val;
+      gboolean ok G_GNUC_UNUSED = filterx_integer_unwrap(lhs, &lhs_val);
+      g_assert(ok);
```
MrAnno: Is this not too strong?

Won't we reach this point with type hints coming from the non-filterx world stating an integer where there may be no integers?

MrAnno:
Klaudia also found a few interesting ones:

1. Fork-ordering bug in JIT set-subscript (real bug). The interpreter copies (cow-forks) the right-hand side before evaluating the left-hand side, but the JIT version forks only after the object was already evaluated. For assignments like d[a][b] = d this aborts debug builds and, in release builds, silently loses the write and leaks a reference cycle.

2. Double COW-float in _do_get_subscript (real bug, perf). The last commit added a float inside the new guarded branch but forgot to remove the older float below it, so shared-container reads now float the child twice and allocate a throwaway ref. Delete the leftover else if; the code currently contradicts its own comment.

Read and write helpers are asymmetric. Write helpers handle all ref/COW logic internally, read helpers make every caller remember to float the child, which is exactly how bug 2 happened. Move the float into the read helpers.

3. String fast path has no runtime type check. fx_jit_string_plus casts the lhs to FilterXString purely on the static type hint — the one thing the last commit taught us not to trust. A wrong hint means reading garbage memory; one is_type check with fallback to filterx_object_add fixes it.

4. The type guard checks inheritance, not layout. filterx_object_is_type_or_ref walks the super-type chain, while the commit message promises "layout equals the hinted". Fine today (dict/list have no subtypes), but a future dict subtype would pass the guard and have its overridden methods bypassed; exact type-pointer comparison matches the intent.

lib/filterx/object-dict.h
fernya: I added a prompt to be terse as possible and comment most important invariants, therefore these abominations will be dropped.

lib/filterx/object-dict.c
fernya: I got rid of it.

lib/filterx/expr-get-subscript.c
fernya: Thanks for the convention!

MrAnno:
The JIT compile methods have been moved into `-devirt.c`. I think we should either call it `-jit.c` and move all compile methods there, or not move them.

Those `-private.h` looks a bit too much in my opinion, I think we can just publish those structs in the original header and treat them as private (we sometimes do this).

LGTM otherwise. The tests are failing unfortunately.

PR #1141, opened 2026-06-24
FilterX: various crash fixes

alltilla:

### DIFF
--- a/lib/filterx/func-unset-empties.c
+++ b/lib/filterx/func-unset-empties.c
@@ -60,8 +60,11 @@ typedef struct FilterXFunctionUnsetEmpties_
   guint64 flags;
 } FilterXFunctionUnsetEmpties;
 
-static gboolean _process_dict(FilterXFunctionUnsetEmpties *self, FilterXObject *obj);
-static gboolean _process_list(FilterXFunctionUnsetEmpties *self, FilterXObject *obj);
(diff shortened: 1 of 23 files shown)


lib/filterx/func-unset-empties.c
```diff
@@ -151,13 +156,15 @@ _add_key_to_unset_list_if_needed(FilterXObject *key, FilterXObject *value, gpoin
...
 static gboolean
-_process_dict(FilterXFunctionUnsetEmpties *self, FilterXObject *obj)
+_process_dict(FilterXFunctionUnsetEmpties *self, FilterXObject *obj, gint depth)
 {
   guint64 len, num = 0;
   filterx_object_len(obj, &len);
-  FilterXObject *keys_to_unset[len];
 
-  gpointer user_data[] = { self, GINT_TO_POINTER(len), &num, &keys_to_unset };
+  FilterXObject *stack_keys[UNSET_EMPTIES_STACK_KEYS];
```
bazsi: maybe use g_alloca() here if the size is not big? This is 8kB per recursion, even if the dict is otherwise small

MrAnno (dismissed):

lib/filterx/func-unset-empties.c
alltilla: Fixed.

MrAnno (approved):

PR #1143, opened 2026-06-24
[FilterX-jit] [Devirtualization] Optimize devirtualization

fernya:
Requires #1134 & #1139.

# Expression optimization

Before: each successfully ran JIT-compiled expression made call to `make_typed_object` which had a lot of call overhead (ie. call count = call places in code - function of filterx code length)

**Optimizations:**
- Drop this call when we know that we know that it's superfluous (see array `_always_typed_helpers`)
- Emit this call only from helpers instead of every expression (see array `_typed_twin_helpers`)
We have much less helpers (~"FilterX language elements") than expressions ("FilterX code") so several call places got eliminated.
- Essentially the call places in code is function of expression kind count minus count of no-op places

# Statement optimization

Before: each statement called `fx_jit_process_expr_result` to set the actual block's state (ie. call count = call places in code - function of filterx code length)

**Optimization:**
- Emit this call only from helpers instead of every statement (see array `_stmt_twin_helpers`)
We have much less helpers (~"FilterX language elements") than expressions ("FilterX code") so several call places got eliminated.

### DIFF
--- a/lib/filterx/expr-getattr.c
+++ b/lib/filterx/expr-getattr.c
@@ -156,6 +156,14 @@ _getattr_infer_types(FilterXExpr *s, FilterXTypeEnv *env)
                                                INITIAL_FILTERX_STATIC_TYPE_SPEC);
 }
 
+FilterXExpr *
+filterx_getattr_get_operand(FilterXExpr *s)
+{
+  if (!filterx_expr_is_getattr(s))
+    return NULL;
+  return ((FilterXGetAttr *) s)->operand;
+}
+
 #if SYSLOG_NG_ENABLE_JIT
 
 #include "filterx/jit/jit.h"
--- a/lib/filterx/expr-getattr.c
+++ b/lib/filterx/expr-getattr.c
@@ -195,9 +195,14 @@ fx_jit_do_getattr(FilterXObject *variable, FilterXObject *attr, FilterXExpr *exp
 }
 
 /* Dict-specialized fast path. dict.attr is semantically dict[attr] with a known-string key,
- * so we call filterx_dict_get_subscript_unchecked directly, bypassing both ref's getattr
- * vtable and mapping's getattr → get_subscript hop. @attr is borrowed (owned by the
- * FilterXGetAttr struct) and must not be unrefed. */
+ * so we call filterx_dict_get_subscript_unchecked directly, bypassing the mapping's
+ * getattr → get_subscript hop. @attr is borrowed (owned by the FilterXGetAttr struct) and
+ * must not be unrefed.
+ *
+ * The unchecked lookup unwraps @variable read-only, so when @variable is a ref we must
+ * still float the shared child against it (exactly as the ref's getattr vtable does) to
+ * preserve copy-on-write; otherwise a subsequent setattr on the returned child would mutate
+ * a dict still shared with the source variable. */
 __attribute__((used))
 FilterXObject *
 fx_jit_do_getattr_dict(FilterXObject *variable, FilterXObject *attr, FilterXExpr *expr)
@@ -213,6 +218,8 @@ fx_jit_do_getattr_dict(FilterXObject *variable, FilterXObject *attr, FilterXExpr
   if (!result)
     filterx_eval_push_error_static_info("Failed to get-attribute from object", expr,
                                         "Failed to evaluate key");
+  else if (filterx_object_is_ref(variable))
+    result = filterx_ref_float_shared_child(result, variable);
 
   filterx_object_unref(variable);
   return result;
--- a/lib/filterx/expr-getattr.c
+++ b/lib/filterx/expr-getattr.c
@@ -234,6 +234,20 @@ fx_jit_do_getattr_dict(FilterXObject *variable, FilterXObject *attr, FilterXExpr
   return result;
 }
 
+__attribute__((used)) __attribute__((noinline))
+FilterXObject *
+fx_jit_do_getattr_typed(FilterXObject *variable, FilterXObject *attr, FilterXExpr *expr)
+{
+  return filterx_expr_make_typed_object(expr, fx_jit_do_getattr(variable, attr, expr));
+}
+
+__attribute__((used)) __attribute__((noinline))
+FilterXObject *
+fx_jit_do_getattr_dict_typed(FilterXObject *variable, FilterXObject *attr, FilterXExpr *expr)
+{
+  return filterx_expr_make_typed_object(expr, fx_jit_do_getattr_dict(variable, attr, expr));
+}
+
 static FilterXIRValue
 _getattr_compile(FilterXExpr *s, FilterXJIT *jit)
 {
--- a/lib/filterx/expr-getattr.c
+++ b/lib/filterx/expr-getattr.c
@@ -146,6 +146,16 @@ _getattr_walk(FilterXExpr *s, FilterXExprWalkFunc f, gpointer user_data)
   return TRUE;
 }
 
+static void
+_getattr_infer_types(FilterXExpr *s, FilterXTypeEnv *env)
+{
+  filterx_expr_infer_types_default(s, env);
+  FilterXGetAttr *self = (FilterXGetAttr *) s;
+  /* Result is one nesting level shallower than the operand. */
+  s->static_type = filterx_static_type_element(self->operand ? self->operand->static_type :
+                                               INITIAL_FILTERX_STATIC_TYPE_SPEC);
+}
+
 #if SYSLOG_NG_ENABLE_JIT
 
 #include "filterx/jit/jit.h"
@@ -190,6 +200,7 @@ filterx_getattr_new(FilterXExpr *operand, FilterXObject *attr_name)
   self->super.is_set = _isset;
   self->super.walk_children = _getattr_walk;
   self->super.free_fn = _free;
+  self->super.infer_types = _getattr_infer_types;
 #if SYSLOG_NG_ENABLE_JIT
   self->super.compile = _getattr_compile;
 #endif
--- a/lib/filterx/expr-getattr.c
+++ b/lib/filterx/expr-getattr.c
@@ -199,10 +199,16 @@ fx_jit_do_getattr(FilterXObject *variable, FilterXObject *attr, FilterXExpr *exp
  * getattr → get_subscript hop. @attr is borrowed (owned by the FilterXGetAttr struct) and
  * must not be unrefed.
  *
+ * The static type DICT is only a hint: coercing containers (e.g. otel_logrecord.body, which
+ * stores an assigned dict as an otel_kvlist) yield a non-dict object at runtime. Since the
+ * unchecked path downcasts to FilterXDictObject, verify the runtime layout first and fall
+ * back to the generic vtable getattr otherwise.
+ *
  * The unchecked lookup unwraps @variable read-only, so when @variable is a ref we must
  * still float the shared child against it (exactly as the ref's getattr vtable does) to
  * preserve copy-on-write; otherwise a subsequent setattr on the returned child would mutate
- * a dict still shared with the source variable. */
+ * a dict still shared with the source variable. (_do_getattr's generic path floats internally
+ * via the ref vtable, so the fallback needs no explicit float.) */
 __attribute__((used))
 FilterXObject *
 fx_jit_do_getattr_dict(FilterXObject *variable, FilterXObject *attr, FilterXExpr *expr)
@@ -213,6 +219,9 @@ fx_jit_do_getattr_dict(FilterXObject *variable, FilterXObject *attr, FilterXExpr
                                           "Failed to evaluate expression");
       return NULL;
     }
+
+  if (!filterx_object_is_type_or_ref(variable, &FILTERX_TYPE_NAME(dict)))
+    return _do_getattr(variable, attr, expr);
 
   FilterXObject *result = filterx_dict_get_subscript_unchecked(variable, attr);
   if (!result)
--- a/lib/filterx/expr-getattr.c
+++ b/lib/filterx/expr-getattr.c
@@ -185,6 +185,7 @@ filterx_getattr_get_operand(FilterXExpr *s)
 
 #include "filterx/jit/jit.h"
 #include "filterx/jit/ffi.h"
+#include "filterx/object-dict.h"
 
 __attribute__((used))
 FilterXObject *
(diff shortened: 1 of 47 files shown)


lib/filterx/jit/ffi.c
```diff
@@ -202,6 +202,79 @@ fx_jit_emit_eval_push_error_info_printf(FilterXJIT *jit, const gchar *msg, Filte
...
+    return FALSE;
+
+  if (_name_in_list(name, _always_typed_helpers, G_N_ELEMENTS(_always_typed_helpers)))
+    return TRUE;
+
+  if (!_name_in_list(name, _typed_twin_helpers, G_N_ELEMENTS(_typed_twin_helpers)))
+    return FALSE;
+
+  gchar *twin_name = g_strconcat(name, "_typed", NULL);
+  LLVMValueRef twin = LLVMGetNamedFunction(jit->mod, twin_name);
```
sodomelle, line 332: The list has to be very precise, adding an entry without having a twin could cause runtine JIT errors.
Let's validate the string lists at filterx_jit_ffi_init.
Claude's suggestion:
```
static void
  _assert_twins_present(FilterXJIT *jit)
  {
    for (gsize i = 0; i < G_N_ELEMENTS(_typed_twin_helpers); i++)
      {
        gchar *t = g_strconcat(_typed_twin_helpers[i], "_typed", NULL);
        LLVMValueRef fn = LLVMGetNamedFunction(jit->mod, t);
        g_assert(fn && !LLVMIsDeclaration(fn));   /* missing/typo'd twin → loud, at init */
        g_free(t);
      }
    /* same loop for _stmt_twin_helpers with "_stmt" */
```

sodomelle (commented):
Looks good, I mentioned something based on Claude's finding.

lib/filterx/expr-switch.c
```diff
@@ -435,6 +435,33 @@ _switch_walk(FilterXExpr *s, FilterXExprWalkFunc f, gpointer user_data)
...
+      FilterXExpr *case_expr = (FilterXExpr *) g_ptr_array_index(self->cases, i);
+      filterx_expr_infer_types(case_expr, env);
+    }
+
+  /* The body may or may not execute (no matching case + no default). Walk it on a clone
+   * and meet back, so any writes the body performs are only kept if they would also hold
+   * under the no-match path. Cross-case fall-through pessimizes the within-body view too,
+   * which is fine for v1. */
+  FilterXTypeEnv *before = filterx_type_env_clone(env);
+  filterx_expr_infer_types(self->body, env);
```
MrAnno, line 459: I don't think we can do this with the body as the body is not executed sequentially, the case branches do jumps and fall-throughs on it.

```
  x = "hoppa";

 switch ($MESSAGE) {
        case "a":
            x = 1;
        case "b":
            total = x + 10;
        };
```

the body is `x = 1; total = x + 10;`, but `x` won't always be an integer.

lib/filterx/expr-getattr.c
```diff
@@ -141,10 +143,44 @@ _getattr_walk(FilterXExpr *s, FilterXExprWalkFunc f, gpointer user_data)
   return TRUE;
 }
 
+static void
+_getattr_infer_types(FilterXExpr *s, FilterXTypeEnv *env)
```
MrAnno, line 167: This crashes:

```
filterx {
    v = {};
    v.a = {};
    v.a.b = 5;
    v.a = {"b": "hopszika"};

    $total = v.a.b + 10;
};
```

lib/filterx/expr-plus-assign.c
```diff
@@ -43,12 +44,41 @@ _eval_plus_assign(FilterXExpr *s)
   return res;
 }
 
+static void
+_plus_assign_infer_types(FilterXExpr *s, FilterXTypeEnv *env)
```
MrAnno, line 49: This crashes:

```
filterx {
    v = {};
    v.a = 1;
    v += {"a": "boom"};
    $total = v.a + 10;
};
```

lib/filterx/expr-switch.c
fernya: We are running through the cases earlier. If their types differ, we store UNKNOWN in our environment. When we infer something else from `self->body` we'd fall back to UNKNOWN.

lib/filterx/filterx-config.c
```diff
@@ -117,6 +118,7 @@ filterx_config_new(GlobalConfig *cfg)
   self->super.deinit = filterx_config_deinit;
   self->super.free_fn = filterx_config_free;
   filterx_env_init(&self->global_env);
+  self->persistent_type_env = filterx_type_env_new();
```
MrAnno, line 121: Types can change between 2 filterx blocks, so we can't blindly trust the inferred types from previous blocks, for example, this crashes:


```
filterx { $foo = 5; };
rewrite { set("hoppa" value("foo")); };
filterx { $bar = $foo + 10; };
```

We need something like we already have in `filterx-scope`: we can detect when filterx blocks don't closely follow each other (the PIF_SYNC_FILTERX_TO_MSG flag and filterx_eval_sync_message).

lib/filterx/expr-setattr.c
```diff
@@ -174,6 +176,120 @@ fx_jit_do_nullv_setattr(FilterXExpr *s, FilterXObject *lhs, FilterXObject *clone
...
+   * across incremental builds (d.a = {}; d.a.b = {}; ...) so deeper accesses devirtualize. */
+  FilterXStaticTypeSpec value_spec = self->new_value ? self->new_value->static_type : INITIAL_FILTERX_STATIC_TYPE_SPEC;
+  filterx_type_env_update_on_write(env, self->object, value_spec);
+
+  const gchar *key = filterx_string_get_value_ref_and_assert_nul(self->attr, NULL);
+
+  /* Per-key update: variable.k0.k1...key = value.  The single-hop case (variable.key)
+   * is just a zero-prefix chain, so this covers it too.  Allows getattr deeper in a
+   * chain to find the precise type. */
+  filterx_type_env_meet_attr_for_chain(env, self->object, key, value_spec);
```
MrAnno, line 196: This doesn't seem to invalidate the the old deeper values.

lib/filterx/expr-switch.c
MrAnno: Klaudia says something is still wrong:

```
1. Switch gets the types wrong inside the body.
- The walk that reads the body top to bottom, as if every case runs: lib/filterx/expr-switch.c:440 (_switch_infer_types), the body walk itself at lib/filterx/expr-switch.c:459.
- Why that's wrong — runtime jumps into the middle of the body instead of running it from the top: lib/filterx/expr-switch.c:303 (filterx_compound_expr_eval_ext(self->body, target)).
- The end-of-switch fix-up that only repairs what comes after the switch: lib/filterx/expr-switch.c:460.
- The test that only covers the harmless case: lib/filterx/tests/test_type_inference.c:673.
```

lib/filterx/filterx-config.c
MrAnno: I have a feeling this doesn't belong to the filterx-config, because there will be problems with fork points too.

This probably belongs to filterx-scope (this is a runtime thing, but the boundaries in filterx-scope are very similar to what we need at compile time), which means we can store type inference info only for consecutive blocks (the same logic as what we do in filterx-scope: fork-points should probably invalidate inferred data, and `PIF_SYNC_FILTERX_TO_MSG` points too).

MrAnno: - two unrelated log paths shouldn't use the same type infer info (let's discuss, maybe I'm wrong)
- fork points should invalidate type info
- `PIF_SYNC_FILTERX_TO_MSG` points should invalidate the type info (non-filterx blocks inserted between filterx blocks)
- $ and declared variables should be considered across filterx blocks, simple local variables don't survive, so those should not be considered here.

lib/filterx/expr-variable.c
```diff
@@ -374,12 +432,16 @@ filterx_floating_variable_expr_new(const gchar *name)
   return filterx_variable_expr_new(name, FX_VAR_FLOATING);
 }
 
-FilterXVariableHandle
-filterx_variable_expr_get_handle(FilterXExpr *s)
+gboolean
+filterx_variable_expr_get_handle(FilterXExpr *s, FilterXVariableHandle *handle_out)
```
MrAnno, line 436: I think we shouldn't change the semantics of this function.

`filterx_expr_is_variable()` is fine for type checking, but `handle_is_macro` could be a separate function call.

lib/filterx/filterx-expr.h
```diff
@@ -294,6 +303,11 @@ filterx_expr_compile(FilterXExpr *self, FilterXJIT *jit)
 {
 #if SYSLOG_NG_ENABLE_JIT
   g_assert(self && self->compile);
+#if SYSLOG_NG_ENABLE_DEBUG
+  /* static_type is only meaningful after inference; compiling an un-inferred node devirtualizes
```
MrAnno, line 307: I don't understand the comment.

My general suggestion: consider removing all AI-generated comments (I usually ask the agent not to even do comments if it can be expressed with clean code instead like better variable/function-names/code organization), and rewrite the ones that are really needed by hand.

lib/filterx/filterx-expr.h
```diff
@@ -294,6 +303,11 @@ filterx_expr_compile(FilterXExpr *self, FilterXJIT *jit)
...
+   * on UNKNOWN, i.e. silently emits the generic path. */
+  g_assert(self->types_inferred);
+#endif
```
MrAnno, line 310: This is compile time, no need to put the assert behind the debug flag.
So overall, this could be a one-liner `g_assert(self->types_inferred);` in each function, which would look less intrusive.

lib/filterx/filterx-expr.h
```diff
@@ -81,12 +82,15 @@ guint32 ignore_falsy_result:1, suppress_from_trace:1, inited:1, optimized:1, sta
...
   FilterXIRValue (*compile)(FilterXExpr *self, FilterXJIT *jit);
   FilterXIRValue (*compile_assign)(FilterXExpr *self, FilterXJIT *jit, FilterXIRValue new_value);
+  void (*infer_types)(FilterXExpr *self, FilterXTypeEnv *env);
 #endif
 
   void (*free_fn)(FilterXExpr *self);
 
   gboolean (*walk_children)(FilterXExpr *self, FilterXExprWalkFunc f, gpointer user_data);
 
+  FilterXStaticTypeSpec static_type;
```
MrAnno, line 92: Shouldn't this belong under the JIT flag?
why don't we use this to mark the `types_inferred` as well?

lib/filterx/filterx-type-inference.h
```diff
@@ -0,0 +1,178 @@
...
+ * pointer copy — exactly the value semantics the inference env and the JIT consumers rely
+ * on. A spec is NEVER NULL: the absent / "no information" type is the canonical unknown
+ * singleton FILTERX_STATIC_TYPE_UNKNOWN_SPEC, which also terminates every element chain.
+ *
+ * Frozen nodes live in a process-global, never-freed pool (the set of distinct chains in
+ * any config is tiny and deduplicated across reloads), so a spec cached on a FilterXExpr
+ * stays valid for as long as that expression does — including at JIT-compile time, after
+ * the inference env has been freed. Inference is not on a hot path, so the freezing cost
+ * is irrelevant. */
+typedef const struct _FilterXStaticTypeNode *FilterXStaticTypeSpec;
```
MrAnno, line 64: Why is this `*`?

lib/filterx/filterx-type-inference.h
```diff
@@ -0,0 +1,178 @@
...
+  FILTERX_STATIC_TYPE_FRESH   = 0xfe,
+} FilterXStaticType;
+
+/* FilterXStaticTypeSpec models the static type of an expression as a linear chain of
+ * nesting levels: the outer kind, then the element type of that container, then the
+ * element type of that, and so on. Each container level has exactly one element type
+ * (heterogeneous elements meet to UNKNOWN), so the structure is a chain, not a tree.
+ *
+ * The chain is represented as frozen / hash-consed immutable nodes: structurally-equal
+ * chains are the SAME pointer, so equality is pointer identity (==) and a copy is just a
```
MrAnno, line 54: These are big blocks of difficult-to-understand AI-generated comments. Please consider removing them, cleaning up the interface and the names to be more self-explanatory and documenting the rest as manually written comments.

Some interesting info on Uncle Bob's opinion about when to comment the code:
https://bpoplauschi.github.io/2021/01/20/Clean-Code-Comments-by-Uncle-Bob-part-2.html
https://blog.cleancoder.com/uncle-bob/2017/02/23/NecessaryComments.html

lib/filterx/filterx-expr.h
MrAnno: Klaudia may have found something here (connected to switch-cases):

```
The new debug assert can misfire on switch.
- Case values moved into the literal cache during optimize: lib/filterx/expr-switch.c:390–391.
- The inference walk never visits them (loop only covers self->cases): lib/filterx/expr-switch.c:440 onward.
- The asserts that would trip if those nodes ever get compiled: lib/filterx/filterx-expr.h:309, lib/filterx/filterx-expr.h:338, lib/filterx/filterx-expr.h:354.
```

lib/filterx/expr-set-subscript.c
```diff
@@ -154,6 +159,211 @@ _free(FilterXExpr *s)
...
+
+__attribute__((used))
+FilterXObject *
+fx_jit_do_set_subscript_list(FilterXObject *object, FilterXObject *key, FilterXObject *new_value, FilterXExpr *expr)
+{
+  return _do_set_subscript(object, key, new_value, expr, filterx_list_set_subscript_via_key, &FILTERX_TYPE_NAME(list));
+}
+
+static void
+_set_subscript_infer_types(FilterXExpr *s, FilterXTypeEnv *env)
```
MrAnno, line 229: setattr and set_subscript are the same thing in case of mappings, so the `attr_to_spec` table is also relevant here.

lib/filterx/expr-set-subscript.c
```diff
@@ -154,6 +159,211 @@ _free(FilterXExpr *s)
...
+{
+  filterx_expr_infer_types_default(s, env);
+  FilterXSetSubscript *self = (FilterXSetSubscript *) s;
+
+  /* Refine the written container's element type with the assigned value (lift if the level
+   * was a freshly-built empty container, meet otherwise). This keeps element-type info alive
+   * across incremental builds (l = []; l[0] = {}; l[0].x = {}; ...) so deeper accesses
+   * devirtualize. */
+  filterx_type_env_update_on_write(env, self->object,
```
MrAnno, line 238: To mirror setattr, a `filterx_type_env_meet_attr_for_chain()` may be missing.

lib/filterx/expr-getattr.c
fernya: confirmed, initial fix done

lib/filterx/expr-plus-assign.c
fernya: confirmed, initial fix done

lib/filterx/expr-switch.c
fernya: I don't get Claudio's reasoning, needs a call

lib/filterx/filterx-type-inference.h
fernya: This is a linked list node.

lib/filterx/filterx-expr.h
fernya: A new "fake" type kind (eg. UNINFERRED) introduces a sudden bump of complexity in inference logic. If the storage of that one plus bit of info is undesired, only then is feasible to change it.

lib/filterx/expr-variable.c
fernya: Reverted, done differently on each new client.

lib/filterx/filterx-type-inference.h
MrAnno: I think it's not worth hiding that `*`.

(We talked about this and naming nested type inference-related stuff IRL.)

PR #1144, opened 2026-06-24
[WIP] Fix diskq data loss on shutdown

MrAnno:
Cherry-pick from https://github.com/syslog-ng/syslog-ng/pull/5722

### DIFF
--- a/lib/logwriter.c
+++ b/lib/logwriter.c
@@ -33,6 +33,7 @@
 #include "seqnum.h"
 #include "str-utils.h"
 #include "find-crlf.h"
+#include "cfg.h"
 #include "mainloop.h"
 #include "mainloop-io-worker.h"
 #include "mainloop-call.h"
@@ -1577,6 +1578,19 @@ _unregister_counters(LogWriter *self)
   _unregister_raw_bytes_stats(self);
 }
 
+static void
+log_writer_deinit_queue(LogWriter *self)
+{
+  GlobalConfig *cfg = log_pipe_get_config(&self->super);
+
+  log_queue_reset_parallel_push(self->queue);
+
+  if (!cfg_is_shutting_down(cfg) || log_queue_flush_on_shutdown(self->queue))
+    log_writer_forced_flush(self);
+  else
+    log_writer_msg_rewind(self);
+}
+
 static gboolean
 log_writer_deinit(LogPipe *s)
 {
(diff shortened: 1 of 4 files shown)


lib/logwriter.c
```diff
@@ -1577,15 +1578,27 @@ _unregister_counters(LogWriter *self)
...
+{
+  GlobalConfig *cfg = log_pipe_get_config(&self->super);
+
+  log_queue_reset_parallel_push(self->queue);
+
+  if (!cfg_is_shutting_down(cfg) || log_queue_flush_on_shutdown(self->queue))
+    log_writer_forced_flush(self);
+  else
+    log_writer_msg_rewind(self);
+}
```
alltilla, line 1592: I think we could just do what threaded dest does (unconditional force flush and rewind the messages left after it):

```
  result = log_threaded_dest_worker_flush(self, mode);
  _process_result(self, result);
  log_queue_rewind_backlog_all(self->queue);
```

@MrAnno wdyt?

MrAnno: I don't really understand the `!cfg_is_shutting_down(cfg) || log_queue_flush_on_shutdown(self->queue)` condition to be honest, your suggestion sounds reasonable. We just have to make sure we won't delay the shutdown/reload too much if the disk-buffer is big.

bazsi: there's no batching in the case of LogWriter (at least not until #1133), which means that flush actually tries to pop  messages from the queue and deliver them to the destination, whereas the LogThreadedDestDriver interface will only flush the partial batch the driver accumulated.

As I understand the original bug report, delivery at shutdown time is not possible (e.g. iptables DROP on the socket).

To match the threaded driver behaviour, we should only invoke log_proto_client_flush() to empty the accumulated batch and then rewind.

But rewind without #1133 will only ever rewind a single message, as there was no batching at the LogProto level. 

What LogWriter does at shutdown is a desperate attempt to empty its queues, without the chance to know which messages get delivered and which are not. If the connection is broken at this point, it is actually worse to send the messages to the socket, as they would happily survive in the disk buffer if one is configured. Also, syslog-ng sleeps 100msec at exit to let the "last messages" to empty from the queue, should the continue to arrive past the 100msec mark, we have no chance but to exit as instructed and drop the messages on the floor.

With all that said, I'd:
* make sure we flush the partial batch in #1133 and that part must be part of the deinit sequence. Right now this is done by the current flush logic on main.
* if the queue is persistent, do not try to empty the queue, we will potentially end up in a worse situation, so just keep the messages in the queue.
* if the queue is not persistent, we can retain the existing desperate attempt to empty it at shutdown, this will ensure nothing, but our hands are clean :)

I think that's what the fix tries to achieve, sans the batching case, which should probably be added to #1133 anyway. The cfg_is_shutting_down() part tries to avoid the flush in case we are only doing a reload and not a shutdown.

alltilla:
Proposed behavior:

1. afsocker == threaded
2. Disk-buffer
  - shutdown == reload: flush any prepared logs (user space buffers, batches), then rewind
3. mem-buffer
  - shutdown: flush the whole queue until the first error, drop the rest
  - reload: flush any prepared logs (user space buffers, batches), then rewind

Make abstractions cleaner.

PR #1145, opened 2026-06-24
logreader: Honour the prepare()-set idle_timeout in the LPPA_SUSPEND branch

MrAnno:
Cherry-pick from https://github.com/syslog-ng/syslog-ng/pull/5729

### DIFF
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -137,11 +137,17 @@ log_reader_stop_watches(LogReader *self)
 }
 
 static void
-log_reader_disable_watches(LogReader *self)
+log_reader_disable_watches_keep_idle_timer(LogReader *self)
 {
   g_assert(self->watches_running);
   if (self->poll_events)
     poll_events_suspend_watches(self->poll_events);
+}
+
+static void
+log_reader_disable_watches(LogReader *self)
+{
+  log_reader_disable_watches_keep_idle_timer(self);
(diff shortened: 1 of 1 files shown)


bazsi (approved):

lib/logreader.c
```diff
@@ -156,6 +162,16 @@ log_reader_suspend_until_awoken(LogReader *self)
   self->suspended = TRUE;
 }
 
+/* Same as log_reader_suspend_until_awoken() but leaves the idle_timer
+ * registered, so a watchdog timeout passed in via log_proto_server_poll_prepare()
+ * can still wake the reader if no wakeup callback fires within that window. */
+static void
+log_reader_suspend_until_awoken_or_timeout(LogReader *self)
+{
+  log_reader_disable_watches_keep_idle_timer(self);
```
MrAnno, line 171: I think this kind of breaks `idle-timeout()`.

From the documentation's perspective, it's not considered a violation, because:
> AxoSyslog closes the client connection if no data is received for the specified amount of time

But if we consider the actual meaning, an idle timeout should kick in when the source really doesn't have anything to receive and we are truly idle.

`LPPA_SUSPEND` means the receiver itself decided to stop receiving messages, and the reason is usually the opposite of being idle: we needed flow-control, because we received too much.

bazsi: you are right. sorry for the noisy approve.

PR #1149, opened 2026-06-24
FilterX compiler: variable visibility in LLVM

MrAnno:


lib/filterx/filterx-expr.h
```diff
@@ -280,6 +283,20 @@ filterx_expr_compile(FilterXExpr *self, FilterXJIT *jit)
   return self->compile(self, jit);
 }
 
+/* TODO partialJIT: remove once all expressions implement compile_assign() */
+static inline gboolean
+filterx_expr_can_compile_assign(FilterXExpr *self)
```
fernya, line 307: I miss the symmetry with `filterx_expr_can_compile` (there are no self NULL check)

lib/filterx/expr-assign.c
```diff
@@ -97,42 +97,58 @@ _nullv_assign_eval(FilterXExpr *s)
...
+    return fx_jit_emit_expr_eval(jit, s);
+
+  return _compile_assign_to_lhs(self, jit);
 }
 
 static FilterXIRValue
 _nullv_assign_compile(FilterXExpr *s, FilterXJIT *jit)
 {
-  return _compile_assignment(s, jit, "fx_jit_do_nullv_assign");
+  return fx_jit_emit_expr_eval(jit, s);
```
fernya: So we're dumping =?? current JIT code, and resort to the interpreted one?

MrAnno: This is bad, I was careless and overlooked some generated code. Thank you.

fernya (approved):
Everything was resolved as expected. LGTM!

PR #1152, opened 2026-06-25
Fix `--check-startup`

### LINKED ISSUE #1150: `--check-startup` aborts with `filterx_expr_unref: !self->inited` assertion on the stock default config
# AxoSyslog

## Version of AxoSyslog

4.25.0 — official container image `ghcr.io/axoflow/axosyslog:4.25.0`.
(Paste full `syslog-ng --version` output here.)

## Platform
Official AxoSyslog container image (Alpine-based), run via Podman.
Host OS:  macOS 26

## Debug bundle
Not attached — the issue reproduces deterministically on the stock image with no custom configuration (see Steps to reproduce). `syslog-ng-debun -r` output can be provided on request.

# Issue

## Failure
`--check-startup` aborts with an assertion failure. Initialization succeeds (the `system()` notice prints and sources come up); the abort happens afterward, during teardown:

```
system(): Neither of the Linux kernel log devices was detected, continuing without polling either /proc/kmsg or /dev/kmsg;
Bail out! ERROR:lib/filterx/filterx-expr.c:336:filterx_expr_unref: assertion failed: (!self->inited)
**
ERROR:lib/filterx/filterx-expr.c:336:filterx_expr_unref: assertion failed: (!self->inited)
```

This reproduces on the stock shipped config, which contains no `filterx {}` block. `system()` (and `default-network-drivers()`) pull a FilterX expression into the generated tree via `app-parser`:

- `system()` generates `app-parser(topic(syslog) ...)` into its expanded block — `modules/system-source/system-source.c`, `system_generate_app_parser()` (around line 357), invoked at line 391. `default-network-drivers()` does the same — `scl/default-network-drivers/plugin.conf`.
- `app-parser` is the `appmodel` generator plugin — `modules/appmodel/appmodel-plugin.c` (around line 60), implemented in `modules/appmodel/app-parser-generator.c`.
- That generator emits `filterx { ... }` blocks (including `filterx { false; }` frames) — `modules/appmodel/app-parser-generator.c` (around lines 138, 236, 251).

So the failure is not specific to user-written `filterx {}` blocks; any config using `system()` or `default-network-drivers()` is affected.

Additional observed behavior:

- `--syntax-only` passes; `--check-startup` aborts.
- A normal start (without `--check-startup`) does not abort.
- A minimal config with an empty `filterx {}` block also reproduces it, so it is not tied to any particular FilterX statement.

## Steps to reproduce
Stock image, no custom config:

```
podman run --rm ghcr.io/axoflow/axosyslog:4.25.0 --check-startup
```

Or with a minimal explicit config (`repro.conf`):

```
@version: 4.25
@include "scl.conf"
log {
  source { internal(); };
  filterx {};
  destination { stdout(); };
};
```

```
syslog-ng -f repro.conf --check-startup
```

Expected: `--check-startup` validates the config and exits 0.
Actual: assertion abort at `lib/filterx/filterx-expr.c:336`.

## Configuration
Stock config shipped in the image (`/etc/syslog-ng/syslog-ng.conf`), unmodified:

```
@version: 4.25
@include "scl.conf"

source s_local {
	system();
	internal();
};

source s_network {
	default-network-drivers(
		#tls(key-file("/path/to/ssl-private-key") cert-file("/path/to/ssl-cert"))
	);
};

destination d_local {
	file("/var/log/messages");
	file("/var/log/messages-kv.log" template("$ISODATE $HOST $(format-welf --scope all-nv-pairs)\n") frac-digits(3));
};

log {
	source(s_local);
	#source(s_network);
	destination(d_local);
};
```

Minimal config that also reproduces it:

```
@version: 4.25
@include "scl.conf"
log {
  source { internal(); };
  filterx {};
  destination { stdout(); };
};
```

## Input and output logs (if possible)
No input is required — this is an init/teardown-time crash, before any message processing. Full console output:

```
system(): Neither of the Linux kernel log devices was detected, continuing without polling either /proc/kmsg or /dev/kmsg;
Bail out! ERROR:lib/filterx/filterx-expr.c:336:filterx_expr_unref: assertion failed: (!self->inited)
**
ERROR:lib/filterx/filterx-expr.c:336:filterx_expr_unref: assertion failed: (!self->inited)
```

MrAnno:
Fixes #1150

Before this PR, `deinit()` was not called at all during `--check-startup`, only `init()`; but once I started calling `deinit()`, more problems arose.

This PR fixes everything functionally, but it looks bad. I couldn't come up with a cleaner fix without reimplemeting how a real shutdown is done.

### DIFF
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -264,6 +264,13 @@ cfg_shutdown(GlobalConfig *cfg)
   main_loop_exit(main_loop);
 }
 
+static gboolean
+cfg_is_dry_run(GlobalConfig *cfg)
+{
+  MainLoopOptions *main_loop_options = main_loop_get_options();
+  return main_loop_is_dry_run(main_loop_options);
+}
+
 gboolean
 cfg_is_shutting_down(GlobalConfig *cfg)
 {
@@ -322,6 +329,9 @@ cfg_init(GlobalConfig *cfg)
 
   _run_module_post_cfg_inits(cfg);
   app_config_post_init();
+
+  if (cfg_is_dry_run(cfg))
+    return TRUE;
 
   /*
    * TLDR: A half-initialized pipeline turned out to be really hard to deinitialize
--- a/lib/mainloop.c
+++ b/lib/mainloop.c
@@ -136,6 +136,7 @@ struct _MainLoop
    * would be gone.
    */
   gboolean _is_terminating;
+  gboolean cfg_inited;
   gboolean last_config_reload_successful;
   time_t last_config_reload_time;
 
@@ -183,6 +184,12 @@ MainLoop *
 main_loop_get_instance(void)
 {
   return &main_loop;
+}
+
+MainLoopOptions *
+main_loop_get_options(void)
+{
+  return main_loop_get_instance()->options;
 }
 
 /* called when syslog-ng first starts up */
@@ -443,6 +450,7 @@ main_loop_exit_finish(gpointer user_data)
    * threads are running.  This will unregister ivykis tasks and timers
    * that could fire while the configuration is being destructed */
   cfg_deinit(self->current_configuration);
+  self->cfg_inited = FALSE;
   iv_quit();
 }
 
@@ -452,6 +460,16 @@ main_loop_exit_timer_elapsed(gpointer user_data)
   MainLoop *self = (MainLoop *) user_data;
 
   main_loop_worker_sync_call(main_loop_exit_finish, self);
+}
+
+static void
+main_loop_deinit_cfg(MainLoop *self)
+{
+  app_pre_shutdown();
+  cfg_deinit(self->current_configuration);
+  self->cfg_inited = FALSE;
+
+  self->_is_terminating = TRUE;
 }
 
 static void
@@ -701,6 +719,10 @@ main_loop_read_and_init_config(MainLoop *self)
     {
       return 2;
     }
+  self->cfg_inited = TRUE;
+
+  if (options->check_startup)
+    return 0;
 
   self->control_server = control_init(resolved_configurable_paths.ctlfilename);
 
@@ -724,6 +746,9 @@ main_loop_free_config(MainLoop *self)
 void
 main_loop_deinit(MainLoop *self)
 {
+  if (main_loop_is_dry_run(self->options) && self->cfg_inited)
+    main_loop_deinit_cfg(self);
+
   main_loop_free_config(self);
 
   if (self->cfg_monitor)
@@ -732,7 +757,8 @@ main_loop_deinit(MainLoop *self)
       file_monitor_free(self->cfg_monitor);
     }
 
-  control_deinit(self->control_server);
+  if (self->control_server)
+    control_deinit(self->control_server);
 
   iv_event_unregister(&self->exit_requested);
   main_loop_call_deinit();
--- a/lib/mainloop.c
+++ b/lib/mainloop.c
@@ -786,12 +786,6 @@ main_loop_thread_resource_deinit(void)
   g_cond_clear(&thread_halt_cond);
 }
 
-gboolean
-main_loop_is_control_server_running(MainLoop *self)
-{
-  return self->control_server != NULL;
-}
-
 GQuark
 main_loop_error_quark(void)
 {
--- a/lib/mainloop.c
+++ b/lib/mainloop.c
@@ -748,6 +748,9 @@ main_loop_deinit(MainLoop *self)
 void
 main_loop_run(MainLoop *self)
 {
+  if (main_loop_is_dry_run(self->options))
+    return;
+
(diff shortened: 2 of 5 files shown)


fernya (dismissed):
This solution creates a remedy for the problem, with its structural shortcomings.

The proper way would be fire up the whole ivykis/worker infrastructure with safeguards to keep the world pristine when requesting a dry-run-ish behavior, and deinitialize it on the same code path as normal runs. However that would require extensive development and validation which is not feasible right now.

lib/mainloop.c
```diff
@@ -748,6 +748,9 @@ main_loop_deinit(MainLoop *self)
 void
 main_loop_run(MainLoop *self)
 {
+  if (main_loop_is_dry_run(self->options))
+    return;
+
```
bazsi, line 779: this should probably go to a separate patch. if this one only extracts the dry_run() function, this is a functional change, that is not strictly needed for this patch

bazsi: I think this is never even exercised. The only place where main_loop_run() is called is the main program and by the time it calls main_loop_run() it bails out if dry run was requested.

So my understanding is that in the case of dry run, we never run main_loop_run(). Is that right? in that case why should we check dry-run here?

lib/cfg.c
```diff
@@ -330,7 +330,9 @@ cfg_init(GlobalConfig *cfg)
    * task/timer/fdwatch ordering in ivykis during this action).
    * See: https://github.com/syslog-ng/syslog-ng/pull/3176#issuecomment-638849597
    */
-  g_assert(cfg_tree_post_config_init(&cfg->tree));
+  MainLoopOptions *main_loop_options = main_loop_get_options();
+  if (!main_loop_is_dry_run(main_loop_options))
+    g_assert(cfg_tree_post_config_init(&cfg->tree));
```
bazsi: I think this what I hate the most, e.g. depending on main loop options within the configuration object. I can see that main_loop_deinit_cfg() attempts to tear down threads even now. Is that code disfunctional?

lib/mainloop.c
MrAnno: This doesn't change functionality. I added it because I wanted to make it clear how these flags change the functionality and deinitialization of mainloop without looking outside of the mainloop unit.

I also wanted to get rid of the conditional calling in main.c, but that didn't happen in the end.

lib/cfg.c
MrAnno: The check-startup functionality was broken, this was the less intrusive way I could fix it without doing a lot more changes.

This patch avoids creating threads in dry-run modes, which sounds reasonable: we don't want to receive messages or start any kind of functionality.

Starting and then immediately stopping threads would be possible, but we need a fully running mainloop to do a full teardown and the moment.

I think the dry-run in this PR is closer to a real dry-run compared to what we have on main (we inited everything, we started threads, we forgot to deinit everything, the OS killed the threads).

MrAnno: The new deinit method only closes the control server threads.

We probably shouldn't even start the control server in dry-run mode.

lib/mainloop.c
bazsi: ok

bazsi:
this is what I came up with to address one of my comments. not fundamentally better, but it extracts the cfg->mainloop dependency into a separate function and places that close to other similar functions.

it's up to you if you want to apply it.

```
diff --git a/lib/cfg.c b/lib/cfg.c
index 3fd1bdbe21..41a41ec562 100644
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -264,6 +264,13 @@ cfg_shutdown(GlobalConfig *cfg)
   main_loop_exit(main_loop);
 }
 
+gboolean
+cfg_is_dry_run(GlobalConfig *cfg)
+{
+  MainLoopOptions *main_loop_options = main_loop_get_options();
+  return main_loop_is_dry_run(main_loop_options);
+}
+
 gboolean
 cfg_is_shutting_down(GlobalConfig *cfg)
 {
@@ -330,9 +337,11 @@ cfg_init(GlobalConfig *cfg)
    * task/timer/fdwatch ordering in ivykis during this action).
    * See: https://github.com/syslog-ng/syslog-ng/pull/3176#issuecomment-638849597
    */
-  MainLoopOptions *main_loop_options = main_loop_get_options();
-  if (!main_loop_is_dry_run(main_loop_options))
-    g_assert(cfg_tree_post_config_init(&cfg->tree));
+  if (cfg_is_dry_run(cfg))
+    return TRUE;
+
+  if (!cfg_tree_post_config_init(&cfg->tree))
+    g_assert_not_reached();
   return TRUE;
 }
 
```

bazsi (dismissed):

MrAnno:
Looks a bit better, thanks.

bazsi (approved):
I compiled and tested the code, --check-startup properly exits at the end, no aborts are seen, with or without filterx.

PR #1153, opened 2026-06-27
RFC: Implement HTTP source

bazsi:
This PR implements a HTTP source that allows you to implement the backend of various HTTP APIs. At the moment it does not support TLS or authentication, but already allows you to post a newline delimited payload that gets delivered as a list of individual log messages.

there's also an example HEC source in the SCL.


MrAnno:
FYI I had an implementation from 8 years ago, where I used the LogProto layer to make the HTTP server multi-threaded, it scaled very well to millions of messages per sec with batching:
https://github.com/MrAnno/syslog-ng/tree/http-sf-fin/lib/http
https://github.com/MrAnno/syslog-ng/blob/http-sf-fin/modules/examples/sources/http/http-source.c

TLS worked automatically as it used the afsocket code as a base.

(It was based on `nodejs/http-parser`, which is the predecessor of `llhttp`.)

bazsi:
> FYI I had an implementation from 8 years ago, where I used the LogProto layer to make the HTTP server multi-threaded, it scaled very well to millions of messages per sec with batching: https://github.com/MrAnno/syslog-ng/tree/http-sf-fin/lib/http https://github.com/MrAnno/syslog-ng/blob/http-sf-fin/modules/examples/sources/http/http-source.c
> 
> TLS worked automatically as it used the afsocket code as a base.
> 
> (It was based on `nodejs/http-parser`, which is the predecessor of `llhttp`.)

I wanted to decouple the http listener from the actual source, so that multiple sources can "bind" to different URLs on the same port. This means that a single HTTP connection can potentially be associated with multiple sources at the same time. I think this design reflects that a bit more.

Another aspect is which thread is responsible for splitting the input body into multiple log messages. the elastic API uses two lines per message, each a separate JSON. The HEC API uses one-line per message. Should we need to parse the JSON as a part of the decomposition of the HTTP request body.

Anyway would be great to discuss the design here, I had some back and forth myself and landed on this design.

tests/light/functional_tests/source_drivers/http_source/test_http_source.py
```diff
@@ -0,0 +1,161 @@
...
+
+    syslog_ng.start(config)
+
+    messages = [f"message {i}" for i in range(2000)]
+    http_source.write_logs(messages)
+
+    assert file_destination.read_logs(2000) == messages
+
+
+def test_http_source_binds_wildcard_by_default(config, syslog_ng, port_allocator):
```
fernya, line 112: This only proves that without explicit `localip` provided loopback works. We may add another IP for the loopback adapter in the test harness and use that for test positive/negative availability of bound service.

modules/http-source/http-server-listener.c
```diff
@@ -0,0 +1,935 @@
...
+  conn->offset = 0;
+  conn->max_request_size = 0;
+  return 0;
+}
+
+static int
+_on_url(llhttp_t *parser, const char *at, size_t length)
+{
+  HTTPServerConnection *conn = parser->data;
+  g_string_append_len(conn->url, at, length);
```
fernya, line 285: IMO a cap for maximum URL length would be great

modules/http-source/http-server-listener.c
```diff
@@ -0,0 +1,935 @@
...
+static int
+_on_body(llhttp_t *parser, const char *at, size_t length)
+{
+  HTTPServerConnection *conn = parser->data;
+
+  /* an error response is already decided; just drain the rest of the body */
+  if (conn->resp_status != 0)
+    return 0;
+
+  if (conn->max_request_size && conn->body->len + length > conn->max_request_size)
```
fernya, line 318: What if I send 0 as `max_request_size`, then I start to upload as many bytes as I can find?
I think we should cap this when that is unset.

PR #1154, opened 2026-06-28
Fix unit tests in monolithic builds

bazsi:
This branch fixes the building of unit tests in case of `--with-linking-mode=monolithic` is used

This now depends on #1164


mitzkia:
I don’t know whether our goal here is to continuously check the build with monolithic linking? If so, we can add it to one of our workflow files, where we can also enable make check and make light-check. For now we have monolithic build only for 'callgrind' job in devshell.yml. 

There was one other topic to run 'make light-check' for our main release (but it runs also with dynamic linking, not with monolithic) - which is missing for sure. Maybe the best place to run 'make light-check' here is in axosyslog-docker.yml  after of image-build job.

libtest/mock-function.h
```diff
@@ -0,0 +1,38 @@
...
+ *
+ */
+#ifndef LIBTEST_MOCK_FUNCTION_H_INCLUDED
+#define LIBTEST_MOCK_FUNCTION_H_INCLUDED
+
+#include "syslog-ng.h"
+
+#if SYSLOG_NG_ENABLE_BUILTIN_MODULES
+
+#define MOCK_FUNCTION(fn)  __wrap_ # fn
```
alltilla: I believe you meant:

```c
#define MOCK_FUNCTION(fn)  __wrap_##fn
```

bazsi: fixed

alltilla (dismissed):

bazsi:
> I don’t know whether our goal here is to continuously check the build with monolithic linking? If so, we can add it to one of our workflow files, where we can also enable make check and make light-check. For now we have monolithic build only for 'callgrind' job in devshell.yml.
> 
> There was one other topic to run 'make light-check' for our main release (but it runs also with dynamic linking, not with monolithic) - which is missing for sure. Maybe the best place to run 'make light-check' here is in axosyslog-docker.yml after of image-build job.

I think adding the unit tests right within the alpine build scripts makes sense, and that's what this branch does at this point. I would not run light-check here, rather I'd try to run light in the actual production image, once it was compiled. For that to work, we need to pip install light and the testcases on top of the image somehow and then run it.

This way we would have light based end-to-end tests run on top of our actual production artifact.

mitzkia:
> > I don’t know whether our goal here is to continuously check the build with monolithic linking? If so, we can add it to one of our workflow files, where we can also enable make check and make light-check. For now we have monolithic build only for 'callgrind' job in devshell.yml.
> > There was one other topic to run 'make light-check' for our main release (but it runs also with dynamic linking, not with monolithic) - which is missing for sure. Maybe the best place to run 'make light-check' here is in axosyslog-docker.yml after of image-build job.
> 
> I think adding the unit tests right within the alpine build scripts makes sense, and that's what this branch does at this point. I would not run light-check here, rather I'd try to run light in the actual production image, once it was compiled. For that to work, we need to pip install light and the testcases on top of the image somehow and then run it.
> 
> This way we would have light based end-to-end tests run on top of our actual production artifact.

Thanks for the answer.

alltilla (approved):
This raises the `CI @ AxoSyslog image` jobs runtime from 14 to 18 minutes, which is not that bad. I think we can live with it.

PR #1155, opened 2026-06-29
JIT: use LLVM 15 compatible APIs

alltilla:
Hopefully fixes the nightly.


packaging/debian/rules
```diff
@@ -124,7 +124,7 @@ override_dh_auto_configure:
 		--enable-tcp-wrapper \
 		--enable-sql \
 		--enable-json \
-		--enable-jit \
+		--enable-jit=auto \
```
bazsi: I think omitting the parameter should make it auto, I would not write it out explicitly.

bazsi: just checked and indeed we default to auto. I'd probably forcibly enable this in the docker image, as we can control dependencies there and we want it as an error when jit is unavailable.

in the case of debian, I depending on auto-detection is fine.

MrAnno:
If only `LLVMPassBuilderOptionsSetInlinerThreshold()` is the problem, we can use the command line arg alternative and keep the JIT support on all platforms.

alltilla:
Yes, it's only that. It would be nice, thanks.
All the plumbing from this PR could be merged even though, it might be useful in the future.
The configuration check is very helpful so the problem shows up a lot sooner.

Feel free to take ownership of this PR/branch/commits. :)

PS: Did one last force push.

MrAnno (approved):
Thank you!

PR #1025, opened 2026-04-02
ci: add support for asan with cmake 

OverOrion:
I don't think it necessitates a news file.

To try it out (assuming `build` build directory and `install_dir` directory for prefix, currently in repository root):
```shell
$ cmake -DSANITIZER=address -DCMAKE_INSTALL_PREFIX=`realpath ./install_dir` -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -B build .
``` 

```shell
$ make -j 8 install
```

```shell
$ nm -na ../install_dir/sbin/syslog-ng | grep -i asan_ -c
365
```

PR #1156, opened 2026-06-29
cmake: fix UTs, enable ASAN

alltilla:
I am not that familiar with using `cmake` for development, but it heavily outperforms autotools.
I recently set up a build guide for `claude`, which uses `cmake` for the sake of performance.
However some features are missing from it.

I am fixing them in this PR and in #1157.

### DIFF
--- a/cmake/add_tests.cmake
+++ b/cmake/add_tests.cmake
@@ -28,6 +28,15 @@ if (BUILD_TESTING)
   check_pie_supported()
 endif()
 
+# Tests with known sanitizer findings, excluded from the leak/ASan error check.
+# The cmake counterpart of the sanitizer_exception_list in scripts/test-grep.sh.
+# TODO fix them.
+set(SANITIZER_EXCEPTION_LIST
+  test_patterndb
+  test_secure_logging
+  test-mongodb-config
+)
+
 function (add_unit_test)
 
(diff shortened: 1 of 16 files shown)


cmake/add_tests.cmake
```diff
@@ -28,6 +28,15 @@ if (BUILD_TESTING)
   check_pie_supported()
 endif()
 
+# Tests with known sanitizer findings, excluded from the leak/ASan error check.
+# The cmake counterpart of the sanitizer_exception_list in scripts/test-grep.sh.
+# TODO fix them.
```
OverOrion, line 33: In my PR #1025 there are fixes for the patterndb and for systemd-journal ODR violation as well.

I am happy to let this in as is, or cherry-pick those commits here.

Let me know how you wish to proceed. 👍

OverOrion (dismissed):
LGTM, thank you!

alltilla:
Oh, I completely missed that. Let me see what I can cherry pick, thanks!

cmake/add_tests.cmake
alltilla: I cherry-picked one commit, thanks!

alltilla: And another one! :D

OverOrion (approved):

PR #1160, opened 2026-06-30
Light: Support to run light tests on local build

mitzkia:
After this PR we can execute `make light-check PYTEST_OPTS="--third-party-runner docker"`
- this will start clickhouse-server and snmptrad service from docker containers -> so we do not have to install them on our local host

### DIFF
--- a/tests/light/Makefile.am
+++ b/tests/light/Makefile.am
@@ -3,6 +3,7 @@ export abs_top_srcdir
 EXTRA_DIST += \
 	tests/light/CMakeLists.txt \
 	tests/light/Makefile.am \
+	tests/light/docker \
 	tests/light/pytest.ini \
 	tests/light/shared_files \
 	tests/light/tox.ini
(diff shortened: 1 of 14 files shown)


tests/light/src/axosyslog_light/fixtures.py
```diff
@@ -99,6 +100,13 @@ def pytest_addoption(parser):
         help="Docker image to use for running syslog-ng. Used when 'runner' is 'docker'. Default: ghcr.io/axoflow/axosyslog:latest",
     )
 
+    parser.addoption(
+        "--third-party-runner",
+        default="local",
+        choices=["local", "docker"],
```
alltilla: Can we have an `auto` option, that tries to reach for the binary, and if it is not available it falls back to docker? If `auto` was the default, it would make it easy to run this in dbld and on the host without any new options to think about and it would always work.

tests/light/Makefile.am
```diff
@@ -27,6 +31,11 @@ LIGHT_PYTEST_CMD=$(LIGHT_PYTHON_CMD) -m pytest --showlocals --verbosity=3
 light-venv:
 	@$(LIGHT_POETRY_CMD) install
 
+light-third-party-deps:
```
alltilla, line 34: It would be nice to make the docker executors give a hint to run `make light-third-party-deps` if they fail.

alltilla (commented):
Please use the `Assisted-by:` way of giving AI contribution, not the `Co-Authored-By:`.

MrAnno (dismissed):
Sorry, I forgot about this, but now I wanted to do some local testing again and it came into my mind.

The copyright check needs to be fixed, LGTM otherwise.

tests/light/Makefile.am
mitzkia: I have fixed that

tests/light/src/axosyslog_light/fixtures.py
mitzkia: I have fixed that

MrAnno (approved):

alltilla (approved):

MrAnno:
distcheck is broken on main, I'll just ignore that.

PR #1164, opened 2026-07-02
docker/axosyslog-builder: add criterion build

bazsi:
To be able to compile our unit tests. This needs to go in before the actual unit tests fixes, so that the axosyslog-builder image is produced.

### DIFF
--- a/docker/apkbuild/axoflow/criterion/APKBUILD
+++ b/docker/apkbuild/axoflow/criterion/APKBUILD
@@ -1,0 +1,65 @@
+# Contributor: Balazs Scheidler <balazs.scheidler@axoflow.com>
+# Maintainer: Balazs Scheidler <balazs.scheidler@axoflow.com>
+pkgname=criterion
+pkgver=2.4.3
+pkgrel=0
+pkgdesc="A cross-platform C and C++ unit testing framework for the 21st century"
+url="https://github.com/Snaipe/Criterion"
+arch="all"
+license="MIT"
+makedepends="
+	cmake
+	git
+	meson
+	ninja
+	linux-headers
+	libffi-dev
+	libgit2-dev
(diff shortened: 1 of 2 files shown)


bazsi:
This is where the build itself can be checked, it will only run in the main repo once it gets merged:

https://github.com/bazsi/axosyslog/actions/runs/28573210195

alltilla (approved):

docker/apkbuild/axoflow/criterion/APKBUILD
```diff
@@ -0,0 +1,65 @@
...
+		subprojects/libgit2-cmake.wrap \
+		subprojects/nanomsg-cmake.wrap
+
+	# Download of nanopb produces an error as it does not contain a
+	# meson.build file. A meson.build file is not necessary, so ignore
+	# the error.
+	meson subprojects download || true
+
+	# Fix FTBS
+	sed -i '/#ifdef __cplusplus/a # include <cstdint>' \
```
MrAnno, line 40: Is this something that needs to be fixed upstream? What was the problem here?

What does FTBS mean?

docker/apkbuild/axoflow/criterion/APKBUILD
```diff
@@ -0,0 +1,65 @@
...
+builddir="$srcdir/Criterion-$pkgver"
+
+prepare() {
+	default_prepare
+
+	# Use system packages for these instead.
+	rm -v \
+		subprojects/libffi.wrap \
+		subprojects/libgit2-cmake.wrap \
+		subprojects/nanomsg-cmake.wrap
```
MrAnno, line 32: When using the default wrap mode, system packages will be used when they are available, otherwise, they fall back to subprojects.

PR #1167, opened 2026-07-10
arrow-flight: deprecate broken timeout(), introduce keep-alive() options

MrAnno:
AxoSyslog uses long-running DoPut streams.

The timeout() option was not a per-WriteRecordBatch() timeout, it was
for the whole stream, which ideally closes only on reload or shutdown,
which implies that the timeout() option should not exist as it always
causes errors if we wait the configured number of seconds.

The option can be reintroduced if we decide to use shorter DoPut sessions
(based on the API docs, this would be the recommended path).

### DIFF
--- a/modules/arrow-flight/arrow-flight-dest.cpp
+++ b/modules/arrow-flight/arrow-flight-dest.cpp
@@ -158,6 +158,27 @@ arrow_flight_dd_set_batch_bytes(LogDriver *d, glong b)
 }
 
 void
+arrow_flight_dd_set_keepalive_time(LogDriver *d, gint t)
+{
+  ArrowFlightDestDriver *self = (ArrowFlightDestDriver *) d;
+  self->cpp->set_keepalive_time(t);
+}
+
+void
+arrow_flight_dd_set_keepalive_timeout(LogDriver *d, gint t)
+{
+  ArrowFlightDestDriver *self = (ArrowFlightDestDriver *) d;
+  self->cpp->set_keepalive_timeout(t);
+}
+
+void
+arrow_flight_dd_set_keepalive_max_pings(LogDriver *d, gint p)
+{
+  ArrowFlightDestDriver *self = (ArrowFlightDestDriver *) d;
+  self->cpp->set_keepalive_max_pings(p);
+}
+
+void
 arrow_flight_dd_add_string_schema_field(LogDriver *d, const gchar *name, LogTemplate *value)
 {
   ArrowFlightDestDriver *self = (ArrowFlightDestDriver *) d;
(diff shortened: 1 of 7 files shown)


alltilla:
Talked about this IRL.

We feel the need to keep `timeout()` with a `Cancel()` call raised from a timer (based on https://arrow.apache.org/docs/cpp/flight.html#closing-unresponsive-connections)

Or at least investigate the idea.

MrAnno:
Unfortunately, only `DoGet()` can be cancelled, `DoPut()` has a writer and reader, they can be closed, but those methods don't look thread-safe and won't really work properly if the problem is the server side getting stuck.

I tried the stop-token trick mentioned in the docs, that doesn't seem to work with `DoPut()` either.

I have no idea how to implement a proper timeout for the long-running `DoPut()` suboperations with the currently available API.

----

Adding a Cancel method to the writer is possible if we contribute it back, as the private stream field inside ClientStreamWriter has a TryCancel method, but the docs say TryCancel is just a hint.

https://github.com/apache/arrow/blob/7932e197eaa00577ff3e83ddf956022df3ef174c/cpp/src/arrow/flight/client.cc#L405

MrAnno:
DoExchange can be cancelled, but we can't just start using that if the server expects a "put".

The complete lack of WriteRecordBatch/ReadMetadata timeouts is weird, they might really want us to occasionally close that stream...

OverOrion:
Here are my observations:

1. `FlightClientOptions` supports supplying `ClientMiddleware`s, and the docs say the following about them:
> Client-side middleware for a call, instantiated per RPC.
> Middleware should be fast and must be infallible: there is no way to reject the call or report errors from the middleware instance.

The middlewares are a way to do inspection/logging/other RO operation during calls, but they are documented to be infallible, meaning they can't be cancellation points.

---

2. As Anno observed `DoPut()` does not support the `stop_token` way, here is why:

In their implementation `DoExchange()`/`DoGet()` use `ClientStreamReader` as _reader_, but `DoPut()` uses `ClientMetadataReader`. If we take a look at those then it seems that the `DoPut` was intentionally implemented without support for the `stop_token`:
```cpp
/// \brief A reader for any ClientDataStream.
class ClientStreamReader : public FlightStreamReader {
....
/// \brief A RecordBatchReader exposing Flight metadata and cancel
/// operations.
class ARROW_FLIGHT_EXPORT FlightStreamReader : public MetadataRecordBatchReader {
``` 

```cpp
class ClientMetadataReader : public FlightMetadataReader {
....
/// \brief A reader for application-specific metadata sent back to the
/// client during an upload.
class ARROW_FLIGHT_EXPORT FlightMetadataReader {
 public:
  virtual ~FlightMetadataReader();
  /// \brief Read a message from the server.
  virtual Status ReadMetadata(std::shared_ptr<Buffer>* out) = 0;
};
```

alltilla:
:(

Let's just remove the failing test cases now, we can't handle those with this API. Thank you for checking these out.

alltilla (approved):

alltilla:
Thank you.

MrAnno:
Halo?

PR #1168, opened 2026-07-10
filterx: uuid7()

MrAnno:
RFC 9562 UUIDv7, similar to the Rust's Uuid::now_v7().

The implementation uses cached time "now" (unix_time_set_now()),
in order not to complicate the counter implementation further by 
supporting timestamps in the past or future.

### DIFF
--- a/lib/filterx/filterx-globals.c
+++ b/lib/filterx/filterx-globals.c
@@ -123,6 +123,7 @@ _simple_init(void)
   g_assert(filterx_builtin_simple_function_register("str_rstrip", filterx_simple_function_str_rstrip));
   g_assert(filterx_builtin_simple_function_register("str_replace", filterx_simple_function_str_replace));
   g_assert(filterx_builtin_simple_function_register("uuid", filterx_simple_function_uuid));
+  g_assert(filterx_builtin_simple_function_register("uuid7", filterx_simple_function_uuid7));
   g_assert(filterx_builtin_simple_function_register("md5", filterx_simple_function_md5));
   g_assert(filterx_builtin_simple_function_register("sha1", filterx_simple_function_sha1));
   g_assert(filterx_builtin_simple_function_register("sha256", filterx_simple_function_sha256));
--- a/lib/filterx/filterx-globals.c
+++ b/lib/filterx/filterx-globals.c
@@ -122,7 +122,8 @@ _simple_init(void)
   g_assert(filterx_builtin_simple_function_register("str_lstrip", filterx_simple_function_str_lstrip));
   g_assert(filterx_builtin_simple_function_register("str_rstrip", filterx_simple_function_str_rstrip));
   g_assert(filterx_builtin_simple_function_register("str_replace", filterx_simple_function_str_replace));
-  g_assert(filterx_builtin_simple_function_register("uuid", filterx_simple_function_uuid));
+  g_assert(filterx_builtin_simple_function_register("uuid", filterx_simple_function_uuid4));
+  g_assert(filterx_builtin_simple_function_register("uuid4", filterx_simple_function_uuid4));
   g_assert(filterx_builtin_simple_function_register("uuid7", filterx_simple_function_uuid7));
   g_assert(filterx_builtin_simple_function_register("md5", filterx_simple_function_md5));
   g_assert(filterx_builtin_simple_function_register("sha1", filterx_simple_function_sha1));
--- a/lib/uuid.c
+++ b/lib/uuid.c
@@ -22,6 +22,7 @@
  */
 
 #include "uuid.h"
+#include "timeutils/unixtime.h"
 
 #include <openssl/rand.h>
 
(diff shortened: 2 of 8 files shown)


lib/uuid.c
```diff
@@ -52,3 +53,58 @@ uuid_gen_random(gchar *buf, gsize buflen)
...
+}
+
+void
+uuid_gen_v7(gchar *buf, gsize buflen)
+{
+  UnixTime now;
+  unix_time_set_now(&now);
+  guint64 unix_ms = unix_time_to_unix_epoch_usec(now) / 1000;
+
+  g_mutex_lock(&uuid_v7_lock);
```
alltilla, line 78: Do we need global, cross-thread monotonicity? Aren't we good with a thread-local one?
Calling `uuid7()` from FilterX creates a synchronization point between **source/processing** threads, which might affect the performance.

The only place I can see the global monotonicity matter is at a destination, where we must send the logs in increasing UUID order, similar to what `$SEQNUM` does. But then here is the question: What happens if we use multiple destination workers? Sure, in a single batch, the UUIDs will automatically be sorted, but not between different batches coming from different workers. So this implementation only guarantee per destination worker monotonicity, but not a per driver one. If that was a requirement from the destination, we cannot grant it even with the current implementation.

`$SEQNUM` gives the same guarantee, but better as it is destination worker thread local, and it doesn't cause processing thread syncs. Unfortunately, FilterX cannot run in destination threads, so if we want the `$SEQNUM`-like ordering we either must do it globally in the processing threads.

Another alternative would be to make this function thread-local monotonic and expose the a `$(uuid7)` template function, and somehow inject it in the destination's template with the FilterX formatted parts, but not all our destinations support that.

lib/filterx/filterx-globals.c
```diff
@@ -122,7 +122,8 @@ _simple_init(void)
   g_assert(filterx_builtin_simple_function_register("str_lstrip", filterx_simple_function_str_lstrip));
   g_assert(filterx_builtin_simple_function_register("str_rstrip", filterx_simple_function_str_rstrip));
   g_assert(filterx_builtin_simple_function_register("str_replace", filterx_simple_function_str_replace));
-  g_assert(filterx_builtin_simple_function_register("uuid", filterx_simple_function_uuid));
+  g_assert(filterx_builtin_simple_function_register("uuid", filterx_simple_function_uuid4));
+  g_assert(filterx_builtin_simple_function_register("uuid4", filterx_simple_function_uuid4));
```
alltilla, line 126: This might be worth a news entry shoutout.

lib/uuid.c
MrAnno: We'd like to use this inside a database for ordering, so a thread-local one may not be enough, I don't know.

We have similar locks, for example `filterx_stash_retrieve()` got accepted, which blocks all source threads when accessing a JSON dict. Those accesses are more frequent, `uuid7()` would be used once per destination in our use-case.

MrAnno: `workers()` will mix things up unless we have a good partition key, so you are probably right and a thread-local one would be enough.

(As a side-note: I had some arguments about `filterx_stash_retrieve()`: I managed to convince @bazsi to use an GRWLock instead of a Mutex, but instead of using a separate lock for each json-cache-file, a GLOBAL LOCK has been merged. I think that was a big deal, but the situation was already uncomfortable, so I didn't want to push it further. `uuid7()` is a very similar example of such a global lock)

lib/filterx/filterx-globals.c
MrAnno: Will do, thanks.

lib/uuid.c
alltilla: IRL we decided to merge this as is. Clearing up the global locks in filterx will be a separate round.

MrAnno: Thanks.

lib/uuid.c
```diff
@@ -52,3 +53,58 @@ uuid_gen_random(gchar *buf, gsize buflen)
...
+  RAND_bytes((guchar *) &seed, sizeof(seed));
+  return seed >> 23;
+}
+
+void
+uuid_gen_v7(gchar *buf, gsize buflen)
+{
+  UnixTime now;
+  unix_time_set_now(&now);
+  guint64 unix_ms = unix_time_to_unix_epoch_usec(now) / 1000;
```
bazsi, line 76: the cached realtime value is per-thread unfortunately, so depending on which thread we are running this from, this may actually go backwards relative to other threads.

Looking at the implementation, I don't think that's an actual problem, if the time is not increasing, we will just reuse the previous time in the uuid value. This is I think correct, as in this case, we indeed processed the message in the previous thread before the current message.

One alternative could be to use the received timestamp which is guraranteed to be monotonic and should reflect the actual ingestion order, rather than the processing order. But it all depends on the use-case for the uuid7() value.

MrAnno: For our specific use-case, I think the received timestamp wouldn't be the best.

I'll fix the shortcoming later (in a separate PR possibly), and even consider adding a variant that accepts a timestamp as an input (it will add some complexity as well).

alltilla (approved):

PR #1172, opened 2026-07-15
github-actions: test `cmake` + `clang` + `ninja` + `lld`

alltilla:
Depends on #1171


.github/workflows/devshell.yml
```diff
@@ -71,8 +72,35 @@ jobs:
...
+            LIGHT_SELF_CHECK_CMD="cmake --build . --target light-self-check"
+            LIGHT_CHECK_CMD="cmake --build . --target light-check"
+            FUNC_TEST_CMD="cmake --build . --target func-test"
+          else
+            BUILD_CMD="make V=1 -j $(nproc)"
+            CHECK_CMD="make V=1 check -j $(nproc)"
+            INSTALL_CMD="make install -j $(nproc)"
+            LIGHT_SELF_CHECK_CMD="make light-self-check"
+            LIGHT_CHECK_CMD="make light-check"
+            FUNC_TEST_CMD="make VERBOSE=1 func-test"
```
MrAnno: I like when the main CI job of a project lists each build step using short commands, without complicated machinery and variables; that can be used as an always up-to-date documentation of  how the project is compiled and installed.

These variables made it a little more convoluted, but it's not that bad :)

MrAnno (dismissed):
LGTM.

(I'll re-approve after removing the TEMP commit.)

alltilla:
Thanks! Pushed without the temp commit. :)

.github/workflows/devshell.yml
alltilla: Yes it's a bit more complicated :(

MrAnno: If we changed almost all commands in the matrix, can't we just break the matrix into 2 separate ones that use these commands directly?

alltilla: Pushed a version like this.

bazsi: Confirming that the command variables are gone.

bazsi:
without the temp commit this does not run, as we need a newer version of the dbld devshell image, which is only built once this one is merged. to resolve the catch-22 let's merge this.

bazsi (approved):

PR #1179, opened 2026-07-28
.github/workflows: fix flaky macos build

bazsi:


OverOrion (approved):
Thank you!

I think we are hitting this bug here:
https://github.com/Homebrew/brew/issues/23328

Brew recently added the auto number of jobs as default, this restores the prior behavior.

PR #1180, opened 2026-07-29
filterx/json-repr: do not bypass the integer cache

bazsi:
As JSON parsing turned GenericNumbers to FilterXPrimitive instances, it bypassed the integer cache. Let's not do that.

no news needed, probably only matters when we are loading JSONs with lots of integers at runtime.

### DIFF
--- a/lib/filterx/object-primitive.c
+++ b/lib/filterx/object-primitive.c
@@ -316,29 +316,6 @@ filterx_primitive_get_value(FilterXObject *s)
 }
 
 FilterXObject *
-filterx_primitive_new_from_gn(const GenericNumber *gn)
-{
-  FilterXType *type;
-
-  switch (gn->type)
-    {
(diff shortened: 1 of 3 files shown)


lib/filterx/object-primitive.c
```diff
@@ -315,29 +315,6 @@ filterx_primitive_get_value(FilterXObject *s)
   return self->value;
 }
 
-FilterXObject *
-filterx_primitive_new_from_gn(const GenericNumber *gn)
```
alltilla, line 319: Why don't we keep the function and modify its implementation instead?

bazsi: this is the only caller, and I am not sure we need it otherwise. Also, this way it's an extra call, whereas the integer cache is normally resolved completely inline.

alltilla (approved):

PR #1184, opened 2026-08-05
HTTP server framework, abstract HTTP source

MrAnno:
This is the resurrection of my 8 years old thesis in the hope it will bring us good abstractions to build HTTP-based sources easily and quickly. The performance should be outstanding as well.

HTTPServer

LogProtoHTTPServer

----

- https is supported
- http upgrade is not supported
- http pipelining is supported
- http keep-alive is supported
- http 0.9, 1.0 and 1.1 is supported
- http 2 is not supported
- the framework won't react on special headers, only upgrade, connection-length. The user of the framework is responsible for everything else.
- automatic error response when the request can not be parsed
- automatic internal error response if no response handler or response is specified

----

TODO:
- [x] get rid of logreader's structured alternative interface
- [x] no example module, place it under modules/http, call it ehttp() until it's stable
  - but curl is not a dependency here...???
- [x] starving fix
- [x] max-request-size
- [x] max-connections default
- [x] HEC could be an ehttp config, JSONL
- [ ] light tests
- router abstraction for path handling - later
- flow-control active response on timeout - later
- optionally store header, path parts, query string in logmsg - later


lib/logreader.c
```diff
@@ -30,6 +30,7 @@
 
 static void log_reader_io_handle_in(gpointer s);
 static gboolean log_reader_fetch_log(LogReader *self);
+static gboolean log_reader_fetch_structured_log(LogReader *self);
```
MrAnno: This is the part that needs to be reconsidered.

modules/splunk-hec/splunk-hec-source.c
```diff
@@ -0,0 +1,362 @@
...
+    *query = '\0';
+
+  gsize len = strlen(path);
+  while (len > 1 && path[len - 1] == '/')
+    path[--len] = '\0';
+
+  HECEndpoint endpoint = HEC_ENDPOINT_UNKNOWN;
+  if (strcmp(path, "/services/collector") == 0
+      || strcmp(path, "/services/collector/event") == 0
+      || strcmp(path, "/services/collector/event/1.0") == 0)
```
MrAnno: We'll need a proper router..

lib/logreader.c
```diff
@@ -578,16 +585,17 @@ log_reader_fetch_log(LogReader *self)
...
+        break;
+
       msg_count++;
 
-      LogMessage *msg = log_reader_construct_message(self, line, line_len);
-      if (!msg)
-        continue;
+      if (!is_structured)
+        {
+          if (!(msg = log_reader_construct_message(self, line, line_len)))
```
bazsi, line 596: our metrics will not be updated in case of a structured logproto. these are currently done in log_reader_construct_message()

lib/logreader.c
```diff
@@ -492,29 +492,35 @@ log_reader_handle_line(LogReader *self, const guchar *line, gsize length, LogTra
...
+
+static gboolean
+log_reader_post_message(LogReader *self, LogMessage *msg, LogTransportAuxData *aux)
+{
   if (aux)
     {
-      _set_addresses(self, m, aux);
-      m->proto = aux->proto;
+      _set_addresses(self, msg, aux);
+      msg->proto = aux->proto;
```
bazsi, line 511: The intention behind LogTransportAuxData was to carry transport level metadata even if we
are only returning a line of text.

with `is_structured`, the proto returns a LogMessage directly, but without feeding the metadata to the message. I don't see a better solution, but I find it a bit odd, that we have both "unstructured" and "structured" interfaces and while the separation of aux makes sense for the unstructured interface, it is not required at all for the structured one.

lib/logreader.c
bazsi: this can be resolved in the next iteration.

lib/logreader.c
bazsi: same

lib/http/source/http-source.c
```diff
@@ -189,4 +195,6 @@ http_sd_init_instance(HTTPSourceDriver *self, SocketOptions *socket_options,
   self->super.super.super.super.free_fn = http_sd_free_method;
   self->super.setup_addresses = http_sd_setup_addresses;
   self->super.construct_proto = http_sd_construct_proto;
+
+  atomic_gssize_set(&self->super.max_connections, 100);
```
fernya, line 201: I see that `afsocket_sd_init_instance` also does atomic set in its ctor. Why is it required in these places?

lib/http/source/http-source.c
```diff
@@ -157,6 +157,16 @@ http_sd_setup_addresses(AFSocketSourceDriver *s)
...
 http_sd_init_method(LogPipe *s)
 {
+  HTTPSourceDriver *self = (HTTPSourceDriver *) s;
+  GlobalConfig *cfg = log_pipe_get_config(s);
+
+  LogProtoServerOptions *proto_options = &self->super.reader_options.proto_options.super;
+  if (proto_options->init_buffer_size == -1)
+    {
+      gint max_msg_size = proto_options->max_msg_size != -1 ? proto_options->max_msg_size : cfg->log_msg_size;
+      proto_options->init_buffer_size = max_msg_size * 1000;
```
fernya: `1000` could be `INIT_BUFFER_MESSAGE_COUNT` or something like that.

fernya (commented):
As I witness the first source module addition to the source tree I mostly focused on matching the patterns to additions of implemented modules. I found few unclear points, some of the added to the respective lines.

My general question is that should we add TCP_NODELAY flag to the http socket? That would be only interesting when our responses could become large enough that each one is sent in multiple `send` calls. In nginx it's set by default.

lib/http/source/http-source.c
MrAnno: `max_connections` is an atomic variable, so I just want to use its interface consistently.

It wouldn't be necessary here, but I remember more than 10 examples where we investigated nasty bugs for long days just because we assumed we don't need atomic semantics, and things have changed a few months later.

lib/logreader.c
MrAnno: that's bad, I'll fix it

MrAnno: I don't completely know what message size and raw message size should mean in complex protocols.
For HTTP sources, calculating this should be the responsibility of the given HTTP source implementation (after extracting the message from a batch, for example).

lib/http/http-parser.c
```diff
@@ -0,0 +1,307 @@
...
+  if (self->parser.upgrade)
+    return g_error_new(HTTP_PARSER_ERROR, HPE_UPGRADE_NOT_SUPPORTED, "HTTP upgrade is not supported");
+
+  return NULL;
+}
+
+static gboolean
+http_parser_previous_header_exists(HTTPParser *self)
+{
+  return self->current_header_field_value->len != 0 && self->current_header_field_name->len != 0;
```
alltilla: An empty header value is legal.

1. RFC 9112 Section 5 defines `field-line = field-name ":" OWS field-value OWS`.
2. RFC 9110 Section 5.5 defines `field-value = *field-content`.
3. RFC 5234 Section 3.6 makes `*` mean zero or more, so a zero-length value is well-formed.
4. llhttp emits a zero-length `on_header_value` for such a header.

If I understand right we use this function to determine whether we the previous header is complete, checking if the value is empty or not might not be the best way to do so.

lib/logreader.c
MrAnno: HTTP implemented as a logreader proto is not ideal, but it was the easiest to do back then.

We can make logreader more generic, or we can merge the http proto+logreader code into a separate non-blocking http server in the future.

MrAnno:
I moved parts of `afsocket` to the lib, I named it `socket`. This is a big change, even if it's not visible clearly by the diff.
Is this okay?

The ehttp() source is inside the `http` module, but that's already a little weird, because that module is detected based on libcurl, which has nothing to do with the source implementation.
(My idea was to allow creating HTTP-based sources, place them in completely new modules, and depend on the core only.)

lib/http/logproto-http-server.c
```diff
@@ -0,0 +1,480 @@
...
+  LogProtoHTTPServer *self = (LogProtoHTTPServer *) s;
+
+  if (self->state == STATE_PROCESS_LOG_MESSAGES)
+    return LPPA_FORCE_SCHEDULE_FETCH;
+
+  GIOCondition proto_io_direction;
+  gboolean unprocessed_data_in_buffer;
+  if (self->state == STATE_SEND_HTTP_RESPONSE || self->state == STATE_HTTP_ERROR)
+    {
+      unprocessed_data_in_buffer = buffer_size(&self->out_buffer) != 0;
```
alltilla: If a client stops reading this will busy loop us.

lib/http/source/http-source.c
```diff
@@ -0,0 +1,200 @@
...
+http_sd_init_method(LogPipe *s)
+{
+  HTTPSourceDriver *self = (HTTPSourceDriver *) s;
+  GlobalConfig *cfg = log_pipe_get_config(s);
+
+  LogProtoServerOptions *proto_options = &self->super.reader_options.proto_options.super;
+  if (proto_options->init_buffer_size == -1)
+    {
+      gint max_msg_size = proto_options->max_msg_size != -1 ? proto_options->max_msg_size : cfg->log_msg_size;
+      proto_options->init_buffer_size = max_msg_size * 1000;
```
alltilla: A `log-msg-size()` larger than 2.1 MB overflows this. 2.1 MB is not the usual case but not unreasonable setting for some strange sources.

lib/http/logproto-http-server.c
```diff
@@ -0,0 +1,480 @@
...
+
+static void
+log_proto_http_server_free(LogProtoServer *s)
+{
+  LogProtoHTTPServer *self = (LogProtoHTTPServer *) s;
+
+  buffer_deallocate(&self->in_buffer);
+  buffer_deallocate(&self->out_buffer);
+  http_parser_free(self->http_parser);
+
```
alltilla, line 446: We should cleanup `pending_log_messages` here, I think.

lib/http/http-message.c
```diff
@@ -0,0 +1,454 @@
...
+
+  {
+    gsize http_version_length = 9;
+    gchar http_version[http_version_length + 1];
+    g_snprintf(http_version, http_version_length + 1, "HTTP/%u.%u ", self->super.http_major, self->super.http_minor);
+    g_byte_array_append(raw_response, (const guint8 *) http_version, http_version_length);
+  }
+
+  const gchar *status_line = http_response_status_code_to_status_line(self->status_code);
+  gsize status_line_length = strlen(status_line);
```
alltilla, line 436: `http_response_status_code_to_status_line()` can return `NULL`, `strlen(NULL)` segfaults, so we should handle it.

lib/logreader.c
```diff
@@ -556,7 +562,12 @@ log_reader_fetch_log(LogReader *self)
 
       log_transport_aux_data_reinit(aux);
       bookmark = ack_tracker_request_bookmark(self->super.ack_tracker);
-      status = log_proto_server_fetch(self->proto, &msg, &msg_len, &may_read, aux, bookmark);
+
```
alltilla, line 565: [optional]

Right now AUX data is only filled for the first message of the HTTP batch.

lib/http/logproto-http-server.c
```diff
@@ -0,0 +1,480 @@
...
+    default:
+      g_assert_not_reached();
+    }
+}
+
+static GIOStatus
+log_proto_http_server_fetch_data(LogProtoHTTPServer *self, LogTransportAuxData *aux)
+{
+  if (G_UNLIKELY(!buffer_allocated(&self->in_buffer)))
+    buffer_allocate(&self->in_buffer, self->super.options->init_buffer_size);
```
alltilla, line 126: [optional]

Preallocating 64 MB per connection could quickly bump the RAM usage to 6 GB with `max-connections(100)` even with mostly idle connections. Starting small and growing it could be a solution, but I don't think we need this right now.

lib/http/logproto-http-server.c
```diff
@@ -0,0 +1,480 @@
...
+
+  // pass LogTransportAuxData
+  ExtractLogMessagesCallback extract_log_messages;
+  CreateResponseCallback create_response;
+};
+
+static LogProtoPrepareAction
+log_proto_http_server_poll_prepare(LogProtoServer *s, GIOCondition *cond, gint *timeout)
+{
+  /* TODO timeout */
```
alltilla, line 73: [optional]

Isn't this just `*timeout = 60`?

lib/http/logproto-http-server.c
```diff
@@ -0,0 +1,480 @@
...
+
+  if (!buffer_is_empty(&self->out_buffer))
+    return status;
+
+  buffer_deallocate(&self->out_buffer);
+
+  if (self->state == STATE_HTTP_ERROR)
+    return LPS_ERROR;
+
+  self->state = STATE_RECEIVE_HTTP_REQUEST;
```
alltilla, line 314: [optional]

At some point we should honor `Connection: close` and clients without `keep-alive` to free up `max-connections`.

lib/http/buffer.h
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
```
alltilla, line 23: Include guard missing.

alltilla:
> The ehttp() source is inside the http module, but that's already a little weird, because that module is detected based on libcurl, which has nothing to do with the source implementation.

I think functionally it is a good start to put it there, we can create a new module if someone wants http sources but not curl.

MrAnno:
@fernya TCP_NODELAY is a good idea, but it can also hurt performance, depending on the shape of the payload.

nginx has an option for this, which is enabled by default, but they only apply the TCP_NODELAY flag "later", when they detect a kept-alive connection (probably after the the first request-response), or when TLS is used (probably because of the handshake ping-pong).

https://nginx.org/en/docs/http/ngx_http_core_module.html#tcp_nodelay

I'd do the same as they did.

lib/http/http-parser.c
MrAnno: It is used to detect whether we want to store that header or not.

But yes, if a header with an empty value has any meaning, then we should store it. I can't come up with a good example, but RFC allows it, so let's store it.

lib/http/logproto-http-server.c
MrAnno: Nice catch.

lib/http/source/http-source.c
MrAnno: I don't get it. `proto_options->max_msg_size` and `cfg->log_msg_size` are both `log-msg-size()` coming from different places. How can anything overflow here?

lib/http/logproto-http-server.c
MrAnno: I believe we honor both, at least that was the intention.
TCP connection closure will be detected and everything will be freed.

> A client that sends a "close" connection option MUST NOT send further requests on that connection (after the one containing the "close") and MUST close the connection after reading the final response message corresponding to this request.

So we are fine. It is only a recommendation to send back a `Connection: close` header when we receive one, but the client MUST close the connection regardless.

lib/http/source/http-source.c
alltilla: In `LogProtoServerOptions`:
```c
gint init_buffer_size;
```
This is an `int`, with `G_MAXINT = 2,147,483,647` limit.

No matter if `max_msg_size` comes from `proto_options` or `cfg`, if it is larger than 2.1 MB, the `* 1000` multiplication will result in `> G_MAXINT`, overflowing `proto_options->init_buffer_size`.

lib/http/logproto-http-server.c
alltilla: From https://www.rfc-editor.org/rfc/rfc9112.html#section-9.6-6

> A server that receives a "close" connection option MUST initiate closure of the connection (see below) after it sends the final response to the request that contained the "close" connection option. 
> ...
> The server MUST NOT process any further requests received on that connection.

I think currently we are not doing these.

lib/http/source/http-source.c
MrAnno: I misunderstood the comment.

I can choose a bigger type for the buffer size, that won't hurt. But as you said in a later comment, this shows we'll need some dynamically growing buffer so that we can support huge requests without allocating too much by default.

lib/http/logproto-http-server.c
MrAnno: TLDR: this is about a TCP reset problem explained in the RFC. We already free up max-connections() in both mentioned cases.

> A client that sends a "close" connection option MUST NOT send further requests on that connection (after the one containing the "close") and MUST close the connection after reading the final response message corresponding to this request.

This implies we won't have any stuck connections.

> A server that receives a "close" connection option MUST initiate closure of the connection (see below) after it sends the final response to the request that contained the "close" connection option. 

We don't do it explicitly, which can result in the TCP reset problem the RFC mentions, but the connection will be closed and everything will be freed due to the previous paragraph. If a client violates the RFC, then we will have an idle connection while the client literally is keeping it open.

> The server SHOULD send a "close" connection option in its final response on that connection.

This is a recommendation, we can implement it together with the previous one.

> The server MUST NOT process any further requests received on that connection.

We currently violate this if the client misbehaves.

alltilla: I can live with this, just thought I mention it with [optional]. Feel free to resolve the conversation.

lib/logreader.c
MrAnno: Nice catch.

lib/http/logproto-http-server.c
MrAnno: Yeah, it's bad at such scale. I'll create dynamically growing buffer in a separate PR.

lib/http/source/http-source.c
MrAnno: Good idea, I added a function-like macro instead to make it clear this is the buffer size.

lib/http/logproto-http-server.c
MrAnno: I just clarified that max-connections is not affected. We can implement it later.

lib/http/http-parser.c
MrAnno: llhttp added a new notification called `on_header_value_complete`, I could got rid of this completely.

lib/http/logproto-http-server.c
MrAnno: It seems Bazsi implemented it 2 years ago inside `log_proto_server_poll_prepare()`.

alltilla (dismissed):
Great work! Kudos

alltilla (approved):

fernya (approved):
(P)awsome work!

PR #1188, opened 2026-08-09
`disk-buffer`: fix fd leak

### LINKED ISSUE #5747: diskq: _release_dirlock() leaks one file descriptor per disk-buffer file allocation
## Summary

`_release_dirlock()` in `modules/diskq/qdisk.c` unlocks the dirlock but never closes the descriptor that `_grab_dirlock()` opened. Every call to `qdisk_get_next_filename()` therefore leaks one file descriptor for the lifetime of the process, on both the success and the failure path.

## Version and platform

- Encountered on syslog-ng **4.8.1** (Debian trixie package `4.8.1-5+deb13u1`)
- Debian 13 (trixie), x86_64, running in a container

**Affected range: 4.2.0 through 4.12.0 inclusive.** `_release_dirlock()` is missing the `close()` in every release tag I checked: 4.2.0, 4.3.0, 4.4.0, 4.5.0, 4.6.0, 4.7.0, 4.8.0, 4.8.1 and 4.12.0. The 4.0.1 and 4.1.0 tags have no dirlock at all.

## The code

`_grab_dirlock()` opens a descriptor on every call, and closes it only on the `flock()` failure path:

```c
static gboolean
_grab_dirlock(const gchar *dir, gint *fd)
{
  ...
  g_mutex_lock(&filename_lock);

  *fd = open(dirlock_file_path, O_RDWR | O_CREAT, 0600);
  if (*fd < 0)
    {
      ...
      g_mutex_unlock(&filename_lock);
      g_free(dirlock_file_path);
      return FALSE;
    }

  if (flock(*fd, LOCK_EX) < 0)
    {
      ...
      close(*fd);                    /* the only close() in the function */
      g_mutex_unlock(&filename_lock);
      g_free(dirlock_file_path);
      return FALSE;
    }

  g_free(dirlock_file_path);
  return TRUE;
}
```

`_release_dirlock()` does not close it:

```c
static void
_release_dirlock(gint fd)
{
  flock(fd, LOCK_UN);
  g_mutex_unlock(&filename_lock);
}
```

`_grab_dirlock()` has exactly one caller (`qdisk_get_next_filename()`), `_release_dirlock()` has exactly three (that function's three exit paths), and `dirlock_fd` is referenced nowhere else. The descriptor has no other owner and is never closed.

## Observed impact

Measured on a central-log receiver using two templated `file()` destinations with `disk-buffer(reliable(yes))`. The template expands to `<host>/<program>/<date>`, so a new disk-buffer file is allocated per destination per day.

**Normal operation:** 2,858 open descriptors on `<spool>/syslog-ng-disk-buffer.dirlock` against 3,056 disk-buffer files allocated. The residual is buffers created before the last restart, so the correspondence with allocations is 1:1. The leak is slow (one fd per buffer created) but unbounded across a process lifetime.

**Under namespace exhaustion it becomes fatal.** `qdisk_get_next_filename()` searches a fixed 10,000-entry namespace:

```c
gchar *
qdisk_get_next_filename(const gchar *dir, gboolean reliable)
{
  gint dirlock_fd = -1;
  if (!_grab_dirlock(dir, &dirlock_fd))    /* succeeds; fd opened */
    return NULL;

  gchar *filename = NULL;
  for (gint i = 0; i < 10000; i++)
    { ... }                                /* every slot taken, nothing found */

  if (!filename)
    {
      msg_error("Error generating unique queue filename, not using disk queue");
      _release_dirlock(dirlock_fd);        /* leaks */
      return NULL;
    }
  ...
}
```

The grab succeeds even when allocation fails, so **each retry leaks a descriptor as well**. Once the namespace filled, we measured roughly **101 fds/sec** at 100.8% of one core (about 10 ms per attempt, consistent with a linear `stat()` scan over 10,000 candidates), reaching the process's 1,048,576 descriptor limit and aborting on SIGABRT approximately every **2h53m**, then restarting under `Restart=always` and repeating. The receiver filed nothing for the ~26 hours this ran.

I do not have a backtrace: no core file was captured at the time, and the journal has since rotated past it.

## Configuration that triggers it

```
destination d_per_program {
    file(
        "/var/log/central/$(sanitize ${HOST})/$(sanitize ${PROGRAM})/${R_YEAR}-${R_MONTH}-${R_DAY}.log"
        template("${R_ISODATE} ${HOST} ${LEVEL} ${PROGRAM}[${PID}]: ${MESSAGE}\n")
        disk-buffer(
            reliable(yes)
            disk-buf-size(2000000000)
            dir("/var/log/central/spool")
        )
    );
};
```

Any high-cardinality templated `file()` destination with a disk-buffer reproduces it. The leak rate scales with the number of distinct template expansions.

## Proposed fix

```diff
 static void
 _release_dirlock(gint fd)
 {
   flock(fd, LOCK_UN);
+  close(fd);
   g_mutex_unlock(&filename_lock);
 }
```

`close()` releases the `flock` on its own, so the explicit `LOCK_UN` becomes redundant, but leaving it in keeps the diff minimal and the intent readable.

## History

`DIRLOCK_FILENAME`, `_grab_dirlock()` and `_release_dirlock()` were added in fd4236227ad032a23b7c9638eb945448775dd287 ("qdisk: hold lock, while calculating next disk-queue filename", part of #4381) to fix a disk-buffer filename race. It first shipped in 4.2.0.

The omission is visible in that commit's own diff: it adds a `close(*fd)` to `_grab_dirlock()`'s `flock()` failure path, but the matching `_release_dirlock()` it adds alongside has no `close()`.

The function was last modified by #4991, "diskq: open the dirlock file in read-write mode", which changed the `open()` mode in `_grab_dirlock()` and did not add a `close()`.

I could not find an existing issue for the descriptor leak.

Happy to open a PR with the above if that is useful.

MrAnno:
A nice catch by @petek.

Fixes syslog-ng/syslog-ng#5747

### DIFF
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -223,6 +223,7 @@ _release_dirlock(gint fd)
 {
   flock(fd, LOCK_UN);
   g_mutex_unlock(&filename_lock);
+  close(fd);
 }
 
 gchar *
--- a/news/bugfix-1188.md
+++ b/news/bugfix-1188.md
@@ -1,0 +1,1 @@
+`disk-buffer()`: fix leaking a file descriptor per disk-buffer file allocation


petek:
Thanks for the credit!

OverOrion (approved):

PR #1192, opened 2026-08-17
Filterx static type inference tree

fernya:
@MrAnno 

Redid the feature laid out in #1134 with a different data structure (two lists replaced with one tree).

Glossary:

- infer: record knowledge about the statically-known type of the variable `declare a = 2; -> INTEGER`
- open: drop knowledge from the inference tree `d.x=1; -> INTEGER, d.y.b = "s" -> STRING; d.y[$k] = "foo"; -> knowledge about d.y is opened (knowledge is dropped), d.x still inferred. 

Missing (broken) features compared to #1134 
- cross-block inference persistence


lib/filterx/filterx-expr.h
```diff
@@ -81,8 +82,15 @@ guint32 ignore_falsy_result:1, suppress_from_trace:1, inited:1, optimized:1, sta
 #if SYSLOG_NG_ENABLE_JIT
   FilterXIRValue (*compile)(FilterXExpr *self, FilterXJIT *jit);
   FilterXIRValue (*compile_assign)(FilterXExpr *self, FilterXJIT *jit, FilterXIRValue new_value);
+  void (*infer_types)(FilterXExpr *self, FilterXTypeEnv *env);
 #endif
 
+  /* Outside the guard on purpose: the hooks above need FilterXIRValue and
+   * FilterXJIT, this does not.  Guarding it would break every unguarded
+   * assignment to it, and the inference pass is a no-op without the JIT
+   * anyway, so the field simply stays UNKNOWN there. */
```
MrAnno: I think it's enough to explain this in the commit message.

lib/filterx/filterx-expr.c
```diff
@@ -31,6 +31,85 @@
...
 #include "perf/perf.h"
 
+#include <string.h>
+
+/* --- paths ---------------------------------------------------------------------------------- */
+
+/* Never emptied: an interned key may outlive the expression tree it was peeled from.  Locked
+ * because config init is not guaranteed to be single threaded. */
+static GMutex key_pool_lock;
+static GHashTable *key_pool;
```
MrAnno: I believe our `init()`s run in the main thread only.

I think this shouldn't stay here across reloads, the data structure grows infinitely with each completely new configuration, so at least we should free them during reload (FilterXConfig).

Considering that this is needed for type inference only, I think we shouldn't even wait until the reload, the right lifetime is probably `filterx_type_env_new` and `free()`?!

lib/filterx/filterx-expr.h
```diff
@@ -27,10 +27,77 @@
...
+typedef struct _FilterXTypePath
+{
+  FilterXVariableHandle root;
+  guint n_steps;
+  /* steps[] is only a prefix of the location: it nests deeper than FILTERX_TYPE_PATH_MAX_DEPTH, or
+   * a step is a key no literal names (a computed subscript, a list index, the `expr[] = v`
+   * append). */
+  gboolean truncated;
+  const gchar *steps[FILTERX_TYPE_PATH_MAX_DEPTH];
+} FilterXTypePath;
```
MrAnno: I'm not sure if this class and its methods belong to filterx-expr, it's a little unclear for me.

- It's called `FilterX*Type*Path`, but there is no type information stored inside it, so it's basically nothing more than a "fully qualified name" for nested objects (dicts, lists).

- `Path` sounds correct, as it tries to capture paths around nested FilterX data types (dicts and lists, so maybe this belongs those units; or the type inference unit?)

- `steps`: there are the elements of the path I guess

MrAnno: It got moved to `filterx-type-path.h`, but the `type` part in its name is still in question.

lib/filterx/filterx-expr.c
MrAnno: It got moved to `filterx-type-path.c`.

lib/filterx/filterx-expr.h
fernya: Done.

lib/filterx/filterx-expr.h
fernya: Renamed in the series.

lib/filterx/filterx-expr.c
fernya: I think reloads are the proper lifetime as each FilterX block has its own `FilterXTypeEnv` and when cross-block persistence will be added, we expect hits on the key pool.

MrAnno (approved):

PR #1194, opened 2026-08-17
Generic HTTP source, Splunk HEC source SCL

MrAnno:
Depends on #1184 

### `ehttp()`

(will be renamed to http once it's config params are stabilized)

For example,

```
 ehttp(
    port(8080)
    mode("json")
    auth-token("Bearer s3cr3t")
    response-body('{"status": "received"}')
    flags(no-parse)
  );
```

`auth-token()` and `response-body()` are optional; without them every request is accepted and answered with a plain "200 OK" body.  They were created to eventually replace the `webhook()` source.

`mode()`:
- `single`: the whole request body becomes one log message
- `line-separated` (`jsonl`): the body is split on newlines, one message per line
- `json`: the body is JSON: either an array of messages or concatenated JSON objects, one message per element/object
- `auto`: bodies that parse as JSON are handled as json, everything else as line-separated

----

### `splunk-hec()`

For example,

```
 splunk-hec(
    token("a1b2c3d4-e5f6-4a7b-8c9d-0e1f2a3b4c5d")
    transport(tls)
    tls(key-file("/etc/syslog-ng/hec.key") cert-file("/etc/syslog-ng/hec.crt"))
  );
```


fernya (commented):
I did behavioral tests and it works great, I'll do another round to check anything suspicious.

modules/http/source/http-source.c
```diff
@@ -0,0 +1,388 @@
...
+    }
+  return messages;
+}
+
+static void
+_json_object_to_message(struct json_object *message_obj, MsgFormatOptions *parse_options, GQueue *messages)
+{
+  const gchar *message = json_object_get_string(message_obj);
+
+  gsize message_length = strlen(message);
```
alltilla, line 112: `json_object_get_string()` returns `NULL` for the JSON value `null`, crashing `strlen()`.
Example input for the crash: `[null]`

modules/http/source/http-source.c
```diff
@@ -0,0 +1,388 @@
...
+
+static GQueue *
+_extract_log_messages(HTTPRequest *http_request, HTTPSourceConnection *connection)
+{
+  EHTTPSourceDriver *self = (EHTTPSourceDriver *) connection->owner;
+
+  if (!_authenticate(self, http_request))
+    return NULL;
+
+  GQueue *messages = ehttp_extract_modes[self->mode](http_request, connection);
```
alltilla, line 274: [optional]

Probably a message count limit would be useful at some point. A `LogMessage` has a lot of memory overhead for the data itself, so if the data is very small (`a\n`), we can quickly explode the memory.

This is post-auth, so it is not necessarily an attack surface, but can become inconvenient in rare cases.

modules/http/source/http-source.c
```diff
@@ -0,0 +1,388 @@
...
+      data += consumed;
+      remaining -= consumed;
+      json_tokener_reset(tok);
+
+      if (remaining == 0)
+        break;
+
+      obj = json_tokener_parse_ex(tok, data, remaining);
+      if (!obj && json_tokener_get_error(tok) != json_tokener_continue)
+        msg_warning("Error parsing JSON messages");
```
alltilla: [optional]

Can we extract the parsing error here somehow?

modules/http/source/http-source.c
```diff
@@ -0,0 +1,388 @@
...
+  gchar *line = strtok_r(data, "\n", &state);
+
+  while (line)
+    {
+      gsize line_length = strlen(line);
+      LogMessage *msg = msg_format_construct_message(parse_options, (guchar *) line, line_length);
+      msg_format_parse_into(parse_options, msg, (guchar *) line, &line_length);
+      g_queue_push_tail(messages, msg);
+
+      line = strtok_r(NULL, "\n", &state);
```
alltilla: [optional]

Embedded `NUL`s will terminate the line separation early.
We shouldn't really get any `NUL`s here in most of the cases, though.

If we want to be a bit more defensive, we could make use of `body->len` to make sure we processed the whole payload.

modules/http/source/http-source.c
MrAnno: Fixed.

modules/http/source/http-source.c
MrAnno: Done.

modules/http/source/http-source.c
MrAnno: We have good primitives to define upper limits: `max-request-size()` for the proto, `log-iw-size()` for the source, and `log-fifo-size()` + the disk-buffer options on the destination side.

Ideally, this GQueue would not exist, we would have an alternative non-buffering way to push messages onto the pipeline. I think `max-request-size()` can protect us from the real attacks, and we should get rid of this queue later.

modules/http/source/http-source.c
MrAnno: Fixed.

modules/http/source/http-source.c
alltilla: Agree, thanks!

alltilla (approved):

PR #1195, opened 2026-08-19
`light/console_log_reader`: scan console logs line by line 

alltilla:
A test run with trace level messages can leave gigabytes of console log behind.

Scanning a 3.1 GB stderr log peaked a light worker at 8.9 GB RSS, it now peaks at 197 MB.


mitzkia (dismissed):
Nice catch, thanks for the fix.

mitzkia (approved):

PR #1196, opened 2026-08-19
gRPC: save TLS meta info into LogMsg variables

alltilla:

### DIFF
--- a/lib/transport/tls-session.c
+++ b/lib/transport/tls-session.c
@@ -30,6 +30,7 @@
 #include <openssl/rand.h>
 #include <openssl/pkcs12.h>
 #include <openssl/ocsp.h>
+#include <openssl/pem.h>
 #include <ctype.h>
 
 /* TLSSession */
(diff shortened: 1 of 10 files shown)


alltilla:
@orymate FYI

lib/transport/tls-session.c
```diff
@@ -601,6 +602,36 @@ tls_session_set_verifier(TLSSession *self, TLSVerifier *verifier)
   self->verifier = verifier ? tls_verifier_ref(verifier) : NULL;
 }
 
+void
+tls_peer_info_set_subject(TLSPeerInfo *self, X509 *cert)
```
bazsi: nit: rename this to tls_peer_info_set_subject_from_x509()

from the prototype alone, I'd assume that X509 is stored/dupicated somehow.

bazsi: or even

tls_peer_info_extract_from_x509()

and similarly tls_peer_info_extract_from_pem()

bazsi (dismissed):

lib/transport/tls-session.c
alltilla: I agree, doing it now.

alltilla: Fixed.

bazsi (approved):

PR #1204, opened 2026-08-19
affile, afsql: fix a lost wildcard file and an sql() memory leak

alltilla:
Preparation step for porting all the `func-test`s to `light`.


mitzkia (dismissed):
LGTM

bazsi (approved):
I validated both fixes and they both look good to me.

PR #1205, opened 2026-08-19
light: port the input driver `func-test`s

alltilla:
Part of porting all the `func-test`s to `light`.

### DIFF
--- a/tests/light/functional_tests/source_drivers/test_input_transports.py
+++ b/tests/light/functional_tests/source_drivers/test_input_transports.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python
+#############################################################################
+# Copyright (c) 2026 Axoflow
+# Copyright (c) 2026 Attila Szakacs-Bertok <attila.szakacs@axoflow.com>
+#
+# This program is free software: you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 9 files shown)


mitzkia (commented):
Thanks for the parametrized testcases (I liked them)

tests/light/functional_tests/source_drivers/test_input_transports.py
```diff
@@ -0,0 +1,116 @@
...
+        pytest.param(_unix_dgram(""), id="unix-dgram-unterminated"),
+        pytest.param(_unix_stream(), id="unix-stream"),
+        pytest.param(_pipe(), id="pipe"),
+        pytest.param(_pipe(pad_size=2048), id="pipe-padded"),
+        pytest.param(_file(), id="file"),
+        pytest.param(_network("tcp"), id="network-tcp"),
+        pytest.param(_network("udp"), id="network-udp"),
+        pytest.param(_network("tls", tls=True), id="network-tls"),
+        pytest.param(_syslog("tcp"), id="syslog-tcp"),
+        pytest.param(_syslog("udp"), id="syslog-udp"),
```
mitzkia, line 112: Optional: maybe a _syslog("tls", tls=True) can also fit here - just for the symmetry 
If it is not a part of the original func_test testcase, in that case fine - we can add it later.

mitzkia (dismissed):
I like both new testcases, they are simple and clean

tests/light/functional_tests/source_drivers/test_input_transports.py
alltilla: Good idea, added :)

bazsi (approved):

mitzkia (approved):
Thanks for the fix

PR #1211, opened 2026-08-24
`s3`: various fixes

### LINKED ISSUE #236: S3: `RuntimeError: dictionary changed size during iteration`
# AxoSyslog
## Version of AxoSyslog
```
4.7.1.831.ge087d9b+4.7.1_git20240709
```

# Issue
```
Exception in thread Thread-3460:
Traceback (most recent call last):
  File "/usr/lib/python3.12/threading.py", line 1073, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.12/threading.py", line 1431, in run
    self.function(*self.args, **self.kwargs)
  File "/usr/lib/syslog-ng/python/syslogng/modules/s3/s3_destination.py", line 352, in __flush_timed_out_s3_objects
    for s3_object in self.s3_objects.values():
RuntimeError: dictionary changed size during iteration
```

alltilla:
Fixes #236


modules/python-modules/syslogng/modules/s3/s3_object.py
```diff
@@ -52,6 +52,19 @@ class AlreadyFinishedError(Exception):
     pass
 
 
+# HTTP status codes for which a request will never succeed as-is.  Mirrors the drop set of the
+# http() destination's default status code handler.
+PERMANENT_DROP_HTTP_STATUS_CODES = {410, 416, 422, 424, 425, 451, 508}
```
MrAnno: There are some interesting error codes here for multipart uploads:

https://docs.aws.amazon.com/AmazonS3/latest/developerguide/ErrorResponses.html

400, 409, and 412 are interesting, but they look more like programming errors in our case.

fernya (commented):
f7002cfb..e0cc07cf looks great (but left unsigned).
The later changes need review from someone else.

If the signing is sorted out, I'll give my approval.

modules/python-modules/syslogng/modules/s3/s3_object.py
alltilla: Nice, added those as well.

alltilla:
@fernya One small change was needed but I think we are good to go now.

Claude also found some more issues, but they can be handled in a separate PR.

fernya:
I had to rebase it to a more recent main.

fernya (approved):
LGTM 🚀

PR #1221, opened 2026-09-02
`http()` source: decompression

alltilla:


MrAnno (approved):

MrAnno:
Content-Encoding is a a comma-separated list based on the RFC, but I don't understand how that would be practical.

PR #1223, opened 2026-09-03
http: add ElasticSearch bulk API server

alltilla:


MrAnno:
Question before the review: don't we want to place this into a separate module? If not, can we put it into a separate directory in the http module so that we make it clear the ehttp implementation is different and there is no dependency between them?

Currently, the source files of the elastic source are added to `modules_http_source_libehttp_source_la_SOURCES`.

alltilla:
Ah, yeah you are right.

I think it still has its place under the `mod-http` module, there are no new dependencies pulled or anything.
However, it should not hook into the ehttp build targets, separate dir and makefiles are due.

I will fix this ASAP, apply some other smaller changes and force push this.
I will notice you once it is review-ready again.

alltilla:
Ready for review.

modules/http/source/elasticsearch-bulk/elasticsearch-bulk-source.c
```diff
@@ -0,0 +1,393 @@
...
+      if ((gsize) (key_end - key) == op_length && strncmp(key, ES_BULK_OPS[i], op_length) == 0)
+        return ES_BULK_OPS[i];
+    }
+  return NULL;
+}
+
+/* bulk clients ack per event: they walk items[] positionally, so every
+ * action needs exactly one item */
+static void
+_append_item(ESBulkSourceConnection *self, const gchar *op, gint status)
```
MrAnno, line 61: _append_item() is about creating the response, I'd put that info into the name of the function or other refactor.

modules/http/source/elasticsearch-bulk/elasticsearch-bulk-source.c
```diff
@@ -0,0 +1,393 @@
...
+          continue;
+        }
+
+      gsize source_length = 0;
+      const gchar *source = _next_line(&data, &remaining, &source_length);
+
+      if (op && strcmp(op, "update") == 0)
+        {
+          /* the source line of an update is a partial document or a script, not a new event */
+          _append_item(connection, op, (source && source_length) ? 200 : 400);
```
MrAnno, line 131: I don't fully understand these `400` cases (when exactly they happen and what they mean), but these are cases when we do nothing silently.

Regardless, shouldn't we log these cases or just produce metrics?

modules/http/source/elasticsearch-bulk/elasticsearch-bulk-source.c
```diff
@@ -0,0 +1,393 @@
...
+      else
+        {
+          _append_item(connection, op ? op : "index", 400);
+        }
+    }
+  return messages;
+}
+
+static gboolean
+_authenticate(ESBulkSourceDriver *self, HTTPRequest *http_request)
```
MrAnno, line 152: Note: There is some copy-paste from `ehttp()`, but we can figure these things out later.

modules/http/source/elasticsearch-bulk/Makefile.am
```diff
@@ -0,0 +1,34 @@
...
+modules_http_source_elasticsearch_bulk_libelasticsearch_bulk_source_la_SOURCES = \
+  modules/http/source/elasticsearch-bulk/elasticsearch-bulk-grammar.y \
+  modules/http/source/elasticsearch-bulk/elasticsearch-bulk-parser.c \
+  modules/http/source/elasticsearch-bulk/elasticsearch-bulk-parser.h \
+  modules/http/source/elasticsearch-bulk/elasticsearch-bulk-source.c \
+  modules/http/source/elasticsearch-bulk/elasticsearch-bulk-source.h
+
+modules_http_source_elasticsearch_bulk_libelasticsearch_bulk_source_la_CPPFLAGS  = \
+  $(AM_CPPFLAGS) \
+  $(JSON_CFLAGS) \
```
MrAnno, line 14: We don't seem to use json-c here.

modules/http/source/elasticsearch-bulk/elasticsearch-bulk-source.c
```diff
@@ -0,0 +1,393 @@
...
+}
+
+static HTTPResponse *
+_create_response(HTTPRequest *http_request, HTTPSourceConnection *connection)
+{
+  ESBulkSourceDriver *self = (ESBulkSourceDriver *) connection->owner;
+  ESBulkSourceConnection *es_connection = (ESBulkSourceConnection *) connection;
+
+  if (!_authenticate(self, http_request))
+    return _new_response(HTTP_FORBIDDEN);
```
MrAnno, line 297: Elastic does 401 here. I think we should change it here, and change it in ehttp() too, it makes more sense.

modules/http/source/elasticsearch-bulk/elasticsearch-bulk-source.c
alltilla: Bulk API has one line that defines and action, then one line (called document) which is the data for the action, like:
```
{"create":{"_index":"filebeat-8.15.0"}}
{"@timestamp":"2026-09-14T12:00:00Z","message":"disk usage at 91%","host":{"name":"web-1"}}
```

Generally there are multiple 2-line "frames"  in batch.

If the "document" part is missing the expected response code is 400:
https://github.com/elastic/elasticsearch/blob/54455dc4fe5fd72769e89a02b4f71b408616406c/server/src/main/java/org/elasticsearch/ExceptionsHelper.java#L85

alltilla: I added a debug log.

modules/http/source/elasticsearch-bulk/Makefile.am
alltilla: Nice catch, removed.

modules/http/source/elasticsearch-bulk/elasticsearch-bulk-source.c
alltilla: Changed in both.

modules/http/source/elasticsearch-bulk/elasticsearch-bulk-source.c
alltilla: Yeah it was confusing. The response has an "items" field, that is what I referenced here. I made it more explicit in the naming.

MrAnno (approved):

PR #1224, opened 2026-09-04
`opentelemetry()` dest: fix crash when sending more than one type of data in one batch

alltilla:
gRPC allows a single RPC per `ClientContext`.

The batch shared one context between the logs, metrics and traces `Export()` calls, so a batch that held two signal types aborted the process on the second call with `Check failed: call_ == nullptr`.

### DIFF
--- a/tests/light/functional_tests/destination_drivers/opentelemetry_destination/test_opentelemetry_destination_batching.py
+++ b/tests/light/functional_tests/destination_drivers/opentelemetry_destination/test_opentelemetry_destination_batching.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python
+#############################################################################
+# Copyright (c) 2026 Axoflow
+# Copyright (c) 2026 Attila Szakacs-Bertok <attila.szakacs@axoflow.com>
+#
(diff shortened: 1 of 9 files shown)


tests/light/functional_tests/destination_drivers/opentelemetry_destination/test_opentelemetry_destination_batching.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
+#############################################################################
```
MrAnno, line 2: The new test case seems to be missing from `tests/light/functional_tests/Makefile.am`.

MrAnno (commented):
LGTM otherwise.

tests/light/functional_tests/destination_drivers/opentelemetry_destination/test_opentelemetry_destination_batching.py
alltilla: Argh, nice catch. Fixed!

MrAnno (approved):

PR #1227, opened 2026-09-05
feat: support gauge like metrics with update_metric

furiel:
This patchset extends the already existing `update_metric` with a new `set=42` parameter. Instead of incrementing a counter, `set` will assign a value to the counter.

```
    update_metric("demo_gauge", set=int($MSG), labels={"host": "demo"});
```

When `set` is omitted, we fall back to increment, exactly the way `update_metric` is used to behave earlier. 

Only nonnegative values are supported. Setting negative values result in error evaluation time:
```
[2026-09-05T11:55:46.251764] FilterX: update_metric() eval error, skipping; err_idx='[1/1]', expr='n/a', error='Failed to evaluate update_metric(): Metric value must be non-negative, got: -100'
```

### DIFF
--- a/modules/metrics-probe/filterx/func-update-metric.c
+++ b/modules/metrics-probe/filterx/func-update-metric.c
@@ -40,34 +40,34 @@ typedef struct FilterXFunctionUpdateMetric_
   struct
   {
     FilterXExpr *expr;
-    gint64 value;
-  } increment;
+    gint64 literal;
+  } value;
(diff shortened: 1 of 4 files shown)


alltilla (approved):
Thanks!

Please follow the commit short message formatting in the git history.
E.g.:
`metrics-probe: support set=<val> for update_metric`

alltilla (approved):

PR #1228, opened 2026-09-05
`s3`: two more small fixes

alltilla:
Two bugs that predate #1211.

---

Only a refused connection was retried on a part upload.

`botocore` raises:
* `ReadTimeoutError`
* `ConnectTimeoutError`
* `ConnectionClosedError`
* `SSLError`

after its own retries give up, and none of them is an `EndpointConnectionError` subclass, so they escaped the retry and abandoned the part until the next start.

Every S3 call now treats `botocore`'s `ConnectionError` and `HTTPClientError` families as transient.

---

`write()` picked the chunk under the object lock and wrote its buffer after releasing it.

A `finish()` from the flush timer in between closed the buffer, and the write failed with an `AssertionError` that `send()` does not catch.

The buffer is now written under the lock.

---

Depends on #1211.

### DIFF
--- a/modules/python-modules/syslogng/modules/s3/s3_object.py
+++ b/modules/python-modules/syslogng/modules/s3/s3_object.py
@@ -25,7 +25,12 @@ from __future__ import annotations
 from .compressable_file_buffer import CompressableFileBuffer
 
 try:
-    from botocore.exceptions import ClientError, EndpointConnectionError
+    from botocore.exceptions import ClientError, ConnectionError as BotocoreConnectionError, HTTPClientError
+
+    # botocore raises these when its own retries did not get past a network failure, like a refused
+    # connection, a timeout or a closed connection.  Errors raised without a network call, like
+    # ParamValidationError, are left out, as the next attempt fails the same way.
+    TRANSIENT_ERRORS = (BotocoreConnectionError, HTTPClientError)
 except ImportError:
     pass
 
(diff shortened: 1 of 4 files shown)


modules/python-modules/syslogng/modules/s3/s3_object.py
```diff
@@ -777,42 +777,34 @@ def write(self, data: bytes) -> None:
...
-                self.__lock.release()
-                self.finish()
+            if new_size <= self.__persist.chunk_size:
                 return
 
-            self.__prev_chunk = self.__current_chunk
-            self.__current_chunk = None
-
-            self.__lock.release()
+            rollover = chunk.part_number < S3Chunk.MAX_PART_NUMBER
```
fernya: The naming looks weird to me. For me the 'rollover' term means that we are out of possible part numbers. If we use the condition on the right hand side, I'd go with the name `next_part_number_available`

modules/python-modules/syslogng/modules/s3/s3_object.py
```diff
@@ -25,7 +25,12 @@
 from .compressable_file_buffer import CompressableFileBuffer
 
 try:
-    from botocore.exceptions import ClientError, EndpointConnectionError
+    from botocore.exceptions import ClientError, ConnectionError as BotocoreConnectionError, HTTPClientError
+
+    # botocore raises these when its own retries did not get past a network failure, like a refused
+    # connection, a timeout or a closed connection.  Errors raised without a network call, like
+    # ParamValidationError, are left out, as the next attempt fails the same way.
+    TRANSIENT_ERRORS = (BotocoreConnectionError, HTTPClientError)
```
fernya, line 33: Excellent idiom!

fernya (changes requested):
Let's meddle about the `rollover` thingy.

Otherwise looks sound, good job!

modules/python-modules/syslogng/modules/s3/s3_object.py
alltilla: Agree, fixed!

fernya (approved):
LGTM! 🚀

PR #1232, opened 2026-09-06
logthreadedsource: fix crash when reloading `keep-alive` workers

alltilla:


news/bugfix-1232.md
```diff
@@ -0,0 +1,5 @@
+`opentelemetry()` source: Fixed a crash on configuration reload.
+
+The source keeps its workers across reloads. A worker whose flow-control window was exhausted at the time
+of the reload, because a destination was not acknowledging messages, crashed on the first message it
```
bazsi: I am not sure I understand the NEWS portion. IIRC we restart threads at reload and the fix seems to change the thread start function.

So what I am understanding from the patch is that if a threaded source has fully exhausted its window, then at reload we start the new threads with zero window.

but then the blocking_post() method basically assumes that at least a single slot is available and that causes the driver to crash.

Do I understand this right? If yes,  then I think it is only the NEWS file that is at fault. If this is not right, t hen I don't think I understand the fix.

please cnfirm.

alltilla: Yeah, I should not have let Claude write the entry. You are correct. I will just remove the long text, it is completely unnecessary.

alltilla: Fixed.

fernya (approved):
Looks sound, the test was successfully ran under heavy load.

LGTM 🚀

PR #1234, opened 2026-09-08
http: relax the worker-partition-key() requirement for message-independent URL templates

MrAnno:
Cherry-picked from syslog-ng/syslog-ng#5740

### DIFF
--- a/lib/template/templates.c
+++ b/lib/template/templates.c
@@ -26,6 +26,7 @@
 #include "template/macros.h"
 #include "template/escaping.h"
 #include "template/repr.h"
+#include "template/simple-function.h"
 #include "timeutils/format.h"
 #include "str-utils.h"
 #include "cfg.h"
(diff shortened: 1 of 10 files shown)


lib/template/templates.c
```diff
@@ -52,6 +53,53 @@ log_template_get_literal_value(const LogTemplate *self, gssize *value_len)
...
+      }
+      return TRUE;
+
+    default:
+      g_assert_not_reached();
+    }
+}
+
+gboolean
+log_template_is_message_independent(const LogTemplate *self)
```
bazsi, line 92: I don't like the name that much. All other template categories are:

* literal
* trivial 

message_independent is not the worst, but I could imagine this to be called one of these:

* static
* fixed

Also, I'd provide an API that allows formatting this value without specifying a message object.

That should make it clear that the message object is not needed at all.

modules/http/http-loadbalancer.c
```diff
@@ -148,6 +148,12 @@ http_lb_target_is_url_templated(HTTPLoadBalancerTarget *self)
   return !log_template_is_literal_string(self->url_template);
 }
 
+gboolean
+http_lb_target_is_url_message_dependent(HTTPLoadBalancerTarget *self)
```
bazsi, line 152: Renaming the template related method would fix this as well, but this flips the negation compared to those. Again by using is_url_fixed should clear this up too.

bazsi (approved):

bazsi (commented):
This is fine as is, but I would prefer this with the naming change.

modules/http/http-loadbalancer.c
alltilla: I will fix this in a new PR.

lib/template/templates.c
alltilla: I will fix this in a new PR.

PR #1245, opened 2026-09-10
wallclocktime: don't depend on the local timezone in strftime %s

alltilla:
FilterX `strftime("%s", dt)` returned a different epoch depending on the daemon's local timezone.

`%s` used `mktime()`, which reads the broken down fields as local time and ignores `wct_gmtoff`.
FilterX formats the fields in the datetime's own offset, so the result was off by the difference between that offset and the local one.

For `strptime("2024-03-18T12:34:13+0900", "%Y-%m-%dT%H:%M:%S%z")` (epoch 1710732853):

| TZ | `%s` |
|---|---|
| UTC | 1710765253 |
| Europe/Budapest | 1710761653 |
| Asia/Tokyo | 1710732853 |

`%s` now goes through `convert_wall_clock_time_to_unix_time()`, which applies `wct_gmtoff`, the same way musl's `strftime()` does.


alltilla:
@orymate FYI

bazsi (approved):
great catch

PR #1246, opened 2026-09-10
filterx/expr-literal-container: set the parent link after the store

alltilla:
A dict literal with a runtime member stored that member with its parent link cleared.

A later write through it found an xref without a parent, treated it as shared and wrote into a shadow that nothing stored back, so the member read back the old value with no error.

Assisted-by: Claude:claude-fable-5-1


bazsi (approved):

bazsi (approved):
Good catch.

PR #1248, opened 2026-09-11
otel/filterx: build the converted containers as xrefs

fernya:
Similar behavior observed as in #1246 but with bit different root cause.

The protobuf to filterx conversion stored its
children into bare dicts and lists.
The parent link is set by the xref set_subscript
after the store, so a child stored into a bare
container never got one.

A later write through such a child found an xref
without a parent, treated it as shared and wrote
into a shadow that nothing stored back. The child
then read back the old value with no error, which
made the resource, scope and log dicts of
mode(filterx-dict) read-only in effect.

The new light test writes into the injected dicts
at each depth and reads the values back.

### DIFF
--- a/modules/grpc/otel/filterx/otel-field-converter.cpp
+++ b/modules/grpc/otel/filterx/otel-field-converter.cpp
@@ -41,6 +41,7 @@
 #include "filterx/object-dict.h"
 #include "filterx/object-list.h"
 #include "filterx/filterx-sequence.h"
+#include "filterx/filterx-ref.h"
 
 #include "compat/cpp-end.h"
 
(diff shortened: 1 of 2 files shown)


modules/grpc/otel/filterx/otel-field-converter.cpp
```diff
@@ -428,6 +429,16 @@ syslogng::grpc::otel::iter_on_otel_protobuf_message_fields(google::protobuf::Mes
...
+/* the stores below must go through an xref, as that is what links the stored
+ * child to its parent container, making a later write through the child land
+ * in the parent */
+static FilterXObject *
+_new_container_ref(FilterXObject *container)
+{
+  filterx_object_cow_prepare(&container);
+  return container;
+}
+
```
bazsi, line 441: I think we should improve this a bit further.

What if:

1) we renamed filterx_dict_new() to filterx_dict_new_bare()
2) and we introduced a filterx_dict_new() that also wraps the root into an xref and returns that
3) and then use the new functions appropriately.

the current constructor is not completely consistent, as any set_subscript() calls will do the wrapping automatically, but the constructor does not.

alltilla: I will create a followup for this.

alltilla (approved):

PR #1250, opened 2026-09-11
dbld: build Criterion from source on non-x86_64 images

alltilla:
Assisted-by: Claude:claude-fable-5-1


bazsi:
when do we plan to use these builds? I don't see them in the CI, or is this change intended to be used in local environments?

also, I'd improve the commit message somewhat, the current is not usable that much, maybe even an empty message is better.

bazsi:
btw, this probably passes CI once rebased.

alltilla:
> when do we plan to use these builds? I don't see them in the CI, or is this change intended to be used in local environments?

I'm trying out development on a mac.

> also, I'd improve the commit message somewhat, the current is not usable that much, maybe even an empty message is better.

Sorry, fixed.

bazsi (approved):

PR #1251, opened 2026-09-12
splunk-s2s: Splunk S2S client and server transport for network()

adamk45:
Forwarding to Splunk today means HEC over HTTP. Splunk's own forwarders and indexers speak S2S on 9997 instead, and there was no way to sit axosyslog in that path: neither to feed an indexer directly, nor to take traffic from a universal forwarder.

This adds a `splunk-s2s` logproto plugin for `network()`, in both directions.

Client (destination):

- header1 + v3 capability handshake, then v4 packets; identifies as `axosyslog`, `fwdType=uf protocolLevel=7`, one channel.
- asks for indexer acknowledgements. A message is acked to the driver only once the indexer confirmed it, not when the write completed. Out-of-order `ACK_ONE`/`ACK_RANGE` are merged and only the contiguous prefix is reported; anything still open on connection loss is rewound.
- index, source, sourcetype and host come from `.splunk.index`, `.splunk.source`, `.splunk.sourcetype`, `.splunk.host`, so they can be set per message. Defaults: `main`, `axosyslog`, `axosyslog`, `$HOST`.

Server (source):

- answers the handshake the way a 10.x indexer does, decodes events into log messages and puts the metadata back into the same `.splunk.*` pairs. Raw chunks are reassembled and split on newlines.
- sends acks back when the forwarder asked for them.
- compression is refused loudly (`compression=0` in the capabilities, zlib switch and the old `--splunk-cooked-mode-v2--:C` greeting both fail the connection). Oversized events are dropped or trimmed according to `log-msg-size()`.

Try it:

```
destination d_splunk {
  network("indexer.example.com" port(9997) transport("splunk-s2s") template("$MSG"));
};

source s_s2s {
  network(port(9997) transport("splunk-s2s") flags(no-parse));
};
```

Tests: unit tests for the codec (with byte vectors captured from a real forwarder/indexer exchange), the client and the server, a client-server roundtrip, and a light test that loops 100 messages through a destination and a source in one process.

`libtest/mock-transport` used `g_strndup()` to copy written buffers, which stopped at the first NUL. Binary payloads never survived that, so it now copies bytes. The unused `empty_write_buffer` helper is gone with it.

Not in this PR: compression, more than one channel, a `splunk-s2s()` option block for the identifier and defaults, docs. News file follows once there is a PR number.


bazsi:
thanks for your contribution. it seems we will need to rebase this to address the conflicts. can you do that?

PR #1252, opened 2026-09-14
grpc: fix abort in isset() on repeated protobuf fields

OverOrion:
ProtobufFieldConverter::is_set() called Reflection::HasField() for every field. HasField() is only valid on singular fields; on a repeated field such as LogRecord.attributes protobuf fails an internal usage check, which newer libprotobuf/abseil builds turn into a FATAL log and abort(). That is not a C++ exception, so the surrounding try/catch never ran and it died with SIGABRT on

```
declare log = otel_logrecord();
log.attributes = {};
isset(log.attributes);
```

Repeated fields have no presence tracking, so treat a non-empty list as set and an empty one as unset via FieldSize(). The converter is shared by the otel logrecord, resource and scope objects, so all of them are covered.

TODO:

- [x] NEWS file


modules/grpc/common/filterx/protobuf-field-converter.hpp
```diff
@@ -200,6 +200,8 @@ class ProtobufFieldConverter
     try
       {
         ProtoReflectors reflectors(*message, field_name);
+        if (reflectors.field_descriptor->is_repeated())
+          return reflectors.reflection->FieldSize(*message, reflectors.field_descriptor) > 0;
```
alltilla: It is strange that:
```
log = {"attributes": {}};
isset(d.attributes); # true
```
but
```
d = otel_logrecord();
isset(d.attributes); # false
```
(or at least this is how I imagine this to work, didn't try the code)

Can't we say repeated fields are always set?
The fact that we could get the field and have an info about whether it is repeated or not tells me that it exists.

OverOrion: You are right, this is how it would behave.

With always-set, `unset(log.attributes)` followed by `isset(log.attributes)` would return true.

I am happy to switch to it always being set, just wanted to discuss this first.

alltilla: In `otel_logrecord` I think we can say certain fields alway exist. These are explicit fields that the protocol defines. We will make `otel_logrecord` read-only anyways hopefully sooner than later.

alltilla (approved):

modules/grpc/common/filterx/protobuf-field-converter.hpp
alltilla: Thanks!

PR #1253, opened 2026-09-14
filterx: more container cow fixes

alltilla:
Followup for https://github.com/axoflow/axosyslog/pull/1248#discussion_r3996634007


fernya (approved):
I did it on another branch with different reasoning, but the result is almost character identical.

We expect that from now on the container types can be used without worrying about copy-on-write bugs.

:fire:

alltilla:
Approve for the tests, thanks @fernya!

PR #1259, opened 2026-09-16
dbld: small build fixes

alltilla:


dbld/rules
```diff
@@ -214,7 +214,7 @@ tag-release:
 	$(GIT) tag $(GIT_RELEASE_TAG)
 
 clean:
-	rm -rf $(BUILD_DIR)/*
+	rm -rf $(BUILD_DIR)/* $(BUILD_DIR)/.[!.]*
```
MrAnno, line 217: How fun it would be to evaluate `$(BUILD_DIR)` as an empty string.

Good thing my `cd` commands take a few seconds to complete every morning.

MrAnno (approved):

dbld/rules
alltilla: HAHA! :D

PR #1261, opened 2026-09-17
Add support for git worktrees in dbld

bazsi:
This allows me to work in various branches in parallel, while keeping them checked out in different worktrees.


alltilla:
Oh I wanted this for a long time! I will check it out tomorrow.

alltilla (approved):

PR #1262, opened 2026-09-20
filterx/jit: track header dependencies of the bitcode units

bazsi:
FYI: I only tested the automake version, and indeed with this a change in the filterx headers cause the bitcode to be recompiled.


lib/filterx/jit/Makefile.am
```diff
@@ -28,14 +28,25 @@ filterx_bitcode_compile = \
...
+# headers exactly (struct layouts, flag bits): track the headers each unit
+# includes, the way automake does for objects, via compiler generated
+# dependency files.  A unit built before this tracking existed has no
+# dependency file yet and is not rebuilt for a header change until its
+# source changes or the tree is cleaned once.
+filterx_bitcode_deps = $(filterx_bitcode_units:.bc=.bc.d)
+
 $(filterx_bitcode_units): %.bc: %.c
-	$(AM_V_GEN)$(filterx_bitcode_compile) -c $< -o $@
+	$(AM_V_GEN)$(filterx_bitcode_compile) -MMD -MP -MT $@ -MF $@.d -c $< -o $@
```
MrAnno, line 40: Does this change belong to this commit? If so, we did we add these flags?

MrAnno (commented):
Thank you for fixing this.

I had one question related to the PR, LGTM otherwise.

(I personally don't mind the Claude-co-authored commit, but I think there was an agreement not to use them.)

